PR #1, opened 2020-02-16
Added capability for keyboard rotation, increased some values.

danieldeankon:


Ralith:
Thanks! In the long run I think we'll actually be increasing the subdivision substantially, but for the time being it can be whatever.

PR #8, opened 2020-03-19
Per-cube-distance based culling based on node searching with a 1 node buffer

cedric-h:


common/src/graph.rs
```diff
@@ -52,41 +53,52 @@ impl<N, C> Graph<N, C> {
     /// correct vertex winding.
     pub fn nearby_cubes(
         &self,
-        start: NodeId,
+        start: Position,
         distance: u32,
```
Ralith: This (and the corresponding config value in the client) should be changed to a `f32`, since it's no longer discrete.

common/src/graph.rs
```diff
@@ -52,41 +53,52 @@ impl<N, C> Graph<N, C> {
...
     ) -> Vec<(NodeId, Vertex, bool, na::Matrix4<f32>)> {
         struct PendingNode {
             id: NodeId,
             parity: bool,
+            parent_in_range: bool,
             distance: u32,
```
Ralith: Is this field dead now?

PR #18, opened 2020-03-28
Working variable-elevation lakes

DimpyRed:
The elevation_boost variable was scrapped.

Temperature is used in place of water_level because of the trouble adding new factors causes.
Stone is being used in the place of water..


Ralith:
Better to add a new water material than overloading stone, I think.

DimpyRed:
@MagmaMcFry Theoretically, could this method lead to flat lakes, or is it doomed to failure?

PR #20, opened 2020-03-29
Rudimentary trees growing on dirt

danieldeankon:


client/src/worldgen.rs
```diff
@@ -155,10 +155,89 @@ pub fn voxels(graph: &mut DualGraph, node: NodeId, cube: Vertex) -> VoxelData {
...
+        // Planting trees on dirt. Trees consist of a block of wood and a block of leaves.
+        // The leaf block is on the opposite face of the wood block as the dirt block.
+        let loc: usize = 2;
+        let voxel_of_interest_index =
+            index(absolute_subchunk_coords(loc, loc, loc, subchunk_offset));
+        let neighbor_data = voxel_neighbors(
+            na::Vector3::new(loc, loc, loc),
+            subchunk_offset,
+            &mut voxels,
+        );
```
Ralith: It looks like this hardcodes the only possible tree location as voxel coordinates 2,2,2. Is that intentional?

client/src/worldgen.rs
```diff
@@ -155,10 +155,89 @@ pub fn voxels(graph: &mut DualGraph, node: NodeId, cube: Vertex) -> VoxelData {
...
+            .filter(|n| n.material == Material::Void)
+            .count();
+
+        // Only plant a tree if there is exactly one adjacent dirt block.
+        if num_void_neighbors == 5 {
+            for i in neighbor_data.iter() {
+                if i.material == Material::Dirt {
+                    voxels.data_mut()[voxel_of_interest_index] = Material::Wood;
+                    let leaf_location = index(i.abs_coords_opposing);
+                    voxels.data_mut()[leaf_location] = Material::Leaves;
```
Ralith: How do we know neither of these coordinates are out of bounds?

client/src/worldgen.rs
```diff
@@ -155,10 +155,89 @@ pub fn voxels(graph: &mut DualGraph, node: NodeId, cube: Vertex) -> VoxelData {
...
+        }
     }
+
     voxels
 }
 
+pub struct NeighborData {
+    #[allow(dead_code)]
+    abs_coords: na::Vector3<f64>,
+    abs_coords_opposing: na::Vector3<f64>,
```
Ralith: Is `f64` justified here? Also, let's avoid adding data that we don't actually need yet; we can always put it in later, but it's harder to go back after the fact and determine what's unnecessary and can be removed.

client/src/worldgen.rs
danieldeankon: Yes, I did intend to hardcode them like that.

client/src/worldgen.rs
danieldeankon: I will remove the unused abs_coords field.
I believe f64 is justified for the abs_coords_opposing vector because I populate it using absolute_subchunk_coords() which returns a Vector3 of f64 values and use it in the index() function which accepts a Vector3 of f64 values. I don't want to change existing functions without discussing it first.

client/src/worldgen.rs
danieldeankon: I have written a test to address this.

client/src/worldgen.rs
cedric-h: Consider `na::Vector3::repeat(loc)` here.

client/src/worldgen.rs
Ralith: Fair enough.

client/src/worldgen.rs
```diff
@@ -627,4 +699,166 @@ mod test {
             epsilon = 1e-8,
         );
     }
+
+    #[test]
+    fn check_tree_planting() {
```
Ralith, line 704: I don't think this (quite large) test is actually useful, since it doesn't exercise any of the code you added/changed, and the hardcoded coordinates are clearly sound anyway. If you strip this out I'll go ahead and merge.

client/src/worldgen.rs
```diff
@@ -155,10 +155,82 @@ pub fn voxels(graph: &mut DualGraph, node: NodeId, cube: Vertex) -> VoxelData {
                 }
             }
         }
+
+        // Planting trees on dirt. Trees consist of a block of wood and a block of leaves.
+        // The leaf block is on the opposite face of the wood block as the dirt block.
+        let loc: usize = 2;
+        let voxel_of_interest_index =
+            index(absolute_subchunk_coords(loc, loc, loc, subchunk_offset));
+        let neighbor_data = voxel_neighbors(na::Vector3::repeat(loc), subchunk_offset, &mut voxels);
```
Ralith, line 164: I'm not sure why this hardcoded location produces good results, but empirically it does. Nice!

PR #37, opened 2020-04-07
Multifractal

DimpyRed:
I added water and snow as a material, and experimented with a different worldgen design.

### DIFF
--- a/client/src/worldgen.rs
+++ b/client/src/worldgen.rs
@@ -127,7 +127,7 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
                 let elev = trilerp(&enviros.max_elevations, cube_coords);
                 let rain = trilerp(&enviros.rainfalls, cube_coords);
                 let temp = trilerp(&enviros.temperatures, cube_coords);
-                let mut elevation_boost = 0_f64;
+                //let mut elevation_boost = 0_f64;
 
 
                 let mut voxel_mat = Material::Dirt;
@@ -156,7 +156,7 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
                     voxel_mat = Material::Wood;
                 }
 
-                max_e -= elevation_boost;
+                //max_e -= elevation_boost;
 
                 if state.surface.voxel_elevation(center, chunk) < max_e / -10.0 {
                     voxels.data_mut()[index(coords)] = voxel_mat;
--- a/client/src/worldgen.rs
+++ b/client/src/worldgen.rs
@@ -152,7 +152,6 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
                     //should not happen.
                     voxel_mat = Material::Wood;
                 }
-                
 
                 if state.surface.voxel_elevation(center, chunk) < max_e / -10.0 {
                     voxels.data_mut()[index(coords)] = voxel_mat;
--- a/client/src/worldgen.rs
+++ b/client/src/worldgen.rs
@@ -144,7 +144,7 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
                     voxel_mat = Material::Snow;
                 }
 
-                if (temp_mod >= 3_f64) && (temp_mod <= 4_f64) {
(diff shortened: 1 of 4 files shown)


client/src/worldgen.rs
```diff
@@ -210,8 +229,10 @@ struct EnviroFactors {
 impl EnviroFactors {
     fn varied_from(parent: Self, spice: u64) -> Self {
         Self {
-            max_elevation: parent.max_elevation + (1 - ((spice % 30) / 10) as i64),
-            temperature: parent.temperature + (1 - ((spice % 15) / 5) as i64),
+            max_elevation: parent.max_elevation
+                + (1 - ((spice % 30) / 10) as i64)
+                + (3 - parent.temperature.rem_euclid(7)),
```
Ralith, line 234: There should be a comment explaining the motivation behind making elevation dependent on temperature. Could the inverse be done instead? That seems more intuitive.

client/src/worldgen.rs
```diff
@@ -124,15 +124,34 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
...
+                    voxel_mat = Material::Stone;
+                }
+
+                //peaks should roughly tend to be snow-covered, and valleys should roughly be watery.
+                let temp_mod = (temp + 0.5_f64).rem_euclid(7_f64);
+                if temp_mod <= 1_f64 {
+                    voxel_mat = Material::Snow;
+                }
+
+                if (temp_mod >= 3_f64) && (temp_mod <= 4_f64) {
```
Ralith, line 147:
```suggestion
                if temp_mod >= 3.0 && temp_mod <= 4.0 {
```

client/src/worldgen.rs
```diff
@@ -124,15 +124,34 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
...
+                    voxel_mat = Material::Water;
+                }
```
Ralith, line 149: Why is this based on temperature? Surely this will lead to physically impossible configurations?

Could we get good results from just replacing void with water when below a certain elevation instead?

client/src/worldgen.rs
```diff
@@ -124,15 +124,34 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
...
+                if rain > 2.5 {
+                    voxel_mat = Material::Leaves;
+                }
+                if rain < -0.5 {
+                    voxel_mat = Material::Stone;
+                }
+
+                //peaks should roughly tend to be snow-covered, and valleys should roughly be watery.
+                let temp_mod = (temp + 0.5_f64).rem_euclid(7_f64);
+                if temp_mod <= 1_f64 {
```
Ralith, line 143:
```suggestion
                if temp_mod <= 1.0 {
```

client/src/worldgen.rs
```diff
@@ -124,15 +124,34 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
...
+                    voxel_mat = Material::Snow;
+                }
+
+                if (temp_mod >= 3_f64) && (temp_mod <= 4_f64) {
+                    voxel_mat = Material::Water;
+                }
+
+                if temp_mod < 0_f64 {
```
Ralith, line 151:
```suggestion
                if temp_mod < 0.0 {
```

client/src/worldgen.rs
DimpyRed: temperature is just a stand-in for an abstract measure of slope.

client/src/worldgen.rs
DimpyRed: The water is there to demonstrate the place where a valley theoretically should be based on the temperature. I also didn't want a minimum water level in this case so the terrain features of the lower areas could be seen.

client/src/worldgen.rs
Ralith: Then it should be renamed to suit its actual use. I think we still want an actual temperature measure to govern biomes, though, so maybe just add a new thing?

client/src/worldgen.rs
Ralith: Perhaps some material other than water would make more sense then? I'm also assuming that "temperature" means "slope" here, which, if so, should certainly be reflected in the code.

PR #38, opened 2020-04-07
Multifractal2

DimpyRed:


client/src/worldgen.rs
```diff
@@ -124,15 +125,40 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex) -> VoxelData {
...
-                    r if r < 1.0 => (Material::Stone, -1.0),
-                    _ => (Material::Sand, -0.5),
-                };
+                let elev = trilerp(&enviros.max_elevations, cube_coords);
+                let rain = trilerp(&enviros.rainfalls, cube_coords);
+                let temp = trilerp(&enviros.temperatures, cube_coords);
+                let slope = trilerp(&enviros.slopeiness, cube_coords);
+
+                let mut voxel_mat = Material::Dirt;
+                let max_e = elev;
```
Ralith, line 134: I don't think we need two different variables for this. Just replace use of max_e with elev?

common/src/world.rs
```diff
@@ -9,10 +9,17 @@ pub enum Material {
     Sand = 3,
     Wood = 4,
     Leaves = 5,
+    Water = 6,
+    Snow = 7,
+    Grass_top = 8,
```
Ralith, line 14:
```suggestion
    Grass = 8,
```

client/src/worldgen.rs
DimpyRed: I disagree. Max_e will eventually be modified by other factors.

DimpyRed: Max_e is an ouput, elev is an input.

PR #43, opened 2020-04-08
Worldgen fixes

Ralith:
Per today's insight:

> If g is a homogeneous scalar field, then the following process will result in a homogeneous field f: for any node W with single descender a, let `f(W) = f(Wa) + (g(W) + g(Wa)) * R`

This change still relies on, for homogeneous scalar field `g(x)`, `g(x) % K` being a homogeneous scalar field. I feel like that's a pretty reasonable judgement, but I'd like @MagmaMcFry's sanity check.


Ralith:
Discussion with @DimpyRed convinced me that the existing logic is probably correct.

PR #69, opened 2020-04-19
Some more terrain gen

danieldeankon:
Temperature is now used in worldgen. There are nine basic terrain types based on combinations of low/medium/high temperature and humidity, and a few extra ones based on the slopiness factor. Terrain varies from smooth to blocky, and the lerp, bilerp, and trilerp functions are replaced with serp, biserp, and triserp functions that produce a more S-shaped curve.

Fog is blue and less intense, and movement speed is faster.


client/shaders/fog.frag
```diff
@@ -15,6 +15,17 @@ void main() {
...
+    float terrain_proportion;
+    float threshold = 1.0;
+    if (dist <= threshold)
+    {
+        terrain_proportion = 1.0;
+    } 
+    else 
+    {
+        terrain_proportion = exp(-(dist - threshold) * fog_density); 
+    }
```
Ralith, line 28: We should try to phrase this in a branch-free way, unless you can show that spirv-opt successfully translates it into one.

client/src/graphics/draw.rs
```diff
@@ -461,10 +461,7 @@ impl Draw {
             Uniforms {
                 view_projection,
                 inverse_projection: projection.try_inverse().unwrap(),
-                // Only 1e-2 of color should be visible from view_distance, assuming
-                // exponential-squared fog
-                fog_density: ((1.0f32 / 1e-2).ln().sqrt()
-                    / self.cfg.local_simulation.view_distance),
+                fog_density: (1.0 / self.cfg.local_simulation.view_distance),
```
Ralith, line 464: The original math is specifically tuned to (mostly) hide stuff outside of view distance, to mask streaming. It should not be replaced with an ad-hoc factor.

common/src/sim_config.rs
```diff
@@ -45,10 +45,10 @@ impl SimConfig {
         let meters_to_absolute = meters_to_absolute(chunk_size, voxel_size);
         SimConfig {
             rate: x.rate.unwrap_or(10),
-            view_distance: x.view_distance.unwrap_or(75.0) * meters_to_absolute,
+            view_distance: x.view_distance.unwrap_or(90.0) * meters_to_absolute,
```
Ralith, line 48: This default should not be changed without a proportionate performance optimization.

client/src/worldgen.rs
```diff
@@ -133,13 +134,14 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex, dimension: u8) ->
     // empirically.
     const ELEVATION_MARGIN: f64 = 0.7;
     let center_elevation = state.surface.elevation(na::Vector3::repeat(0.5), chunk);
-    if center_elevation - ELEVATION_MARGIN > me_max / 10.0 {
+    let threshold = 10.0;
```
Ralith, line 137: This is a scaling factor, not a threshold. Also should probably be a `const`.

client/src/worldgen.rs
```diff
@@ -384,21 +423,51 @@ impl Surface {
     }
 }
 
-fn trilerp<N: na::RealField>(
+fn triserp<N: na::RealField>(
     &[v000, v001, v010, v011, v100, v101, v110, v111]: &[N; 8],
     t: na::Vector3<N>,
+    threshold1: N,
+    threshold2: N,
```
Ralith, line 430: It looks like `threshold2` is *always* `1 - threshold1`, so it should probably be inferred rather than passed everywhere.

client/src/worldgen.rs
```diff
@@ -384,21 +423,51 @@ impl Surface {
...
 ) -> N {
-    fn lerp<N: na::RealField>(v0: N, v1: N, t: N) -> N {
-        v0 * (N::one() - t) + v1 * t
+    fn serp<N: na::RealField>(v0: N, v1: N, t: N, threshold1: N, threshold2: N) -> N {
+        let out: N;
+        if t < threshold1 {
+            out = v0;
+        } else if t < threshold2 {
+            let s = (t - threshold1) / (threshold2 - threshold1);
+            // S-shaped curve that in practice is barely different to a straight line
```
Ralith, line 438: If "barely different" is true, then we shouldn't waste precious CPU time and code readability budget on it.

client/src/worldgen.rs
```diff
@@ -151,39 +153,68 @@ pub fn voxels(graph: &DualGraph, node: NodeId, chunk: Vertex, dimension: u8) ->
...
+
+                let block = triserp(&enviros.blockinesses, cube_coords, 0.2, 0.8);
+                // map real number block to interval (0, 0.25), biased towards 0
+                let triserp_threshold = 2.0f64.powf(block) / (8.0 + 2.0f64.powf(block)) * 0.25;
+                let elev = triserp(
+                    &enviros.max_elevations,
+                    cube_coords,
+                    triserp_threshold,
+                    1.0 - triserp_threshold,
+                );
```
Ralith, line 168: Is the goal of using `triserp`, and introducing `blockiness`, to give the terrain a varying degree of terraced character? If so, we should document that clearly in comments, and retain the simpler and more predictable `trilerp` for the other factors.

Ralith: Additionally, I'm not a huge fan of coupling visible terrain features directly to the topology of the graph, when avoidable. If terraces *are* the goal, would it make more sense to quantize the max elevation value?

client/src/worldgen.rs
```diff
@@ -384,21 +423,51 @@ impl Surface {
...
+fn triserp<N: na::RealField>(
     &[v000, v001, v010, v011, v100, v101, v110, v111]: &[N; 8],
     t: na::Vector3<N>,
+    threshold1: N,
+    threshold2: N,
 ) -> N {
-    fn lerp<N: na::RealField>(v0: N, v1: N, t: N) -> N {
-        v0 * (N::one() - t) + v1 * t
+    fn serp<N: na::RealField>(v0: N, v1: N, t: N, threshold1: N, threshold2: N) -> N {
+        let out: N;
```
Ralith, line 433: You don't need an explicit result variable here, because if statements are expressions: just replace `out = whatever;` with `whatever` (note omitted semicolon)

client/shaders/fog.frag
danieldeankon: I've undone the fog fade, so there is no longer branching.

client/src/graphics/draw.rs
danieldeankon: Changed it back to the original.

common/src/sim_config.rs
danieldeankon: Changed it back to the original.

client/src/worldgen.rs
danieldeankon: Made the suggested changes.

client/src/worldgen.rs
danieldeankon: Made the suggested changes.

client/src/worldgen.rs
danieldeankon: Good point. serp is now piecewise linear.

client/src/worldgen.rs
danieldeankon: Made the change.

client/src/worldgen.rs
danieldeankon: Yes. Specifically, the point of triserp over trilerp is to give the terrain a terraced character, and the point of introducing blockiness is to make the terracing stronger in some regions than others.

As for coupling visible terrain features to the topology of the graph, I think we just have different opinions of how desirable it is. I am in favour of leaning into it or perhaps even embracing it outright, because it seems really difficult to escape it. If we try to make smooth equidistant surfaces out of voxels, we get pentagonal artifacts. No matter how high the subdivision factor is, there will be places where twenty voxels meet at a corner instead of eight voxels---and all voxels get pointer as they get closer to those corners. Hyperbolic planes aligned with voxel faces crisscross through the space at right angles. That said, my opinion might change after seeing the world with a much higher subdivision factor, and quantizing the max elevation value might yield results better than my approach here.

Ralith: You're absolutely right that completely hiding the topology is impossible, at least so long as we're using discrete cubical voxels, and I agree that we shouldn't try to paper over that. However, I'd still like to minimize the extent to which it influences the world to the cases where it's difficult to avoid. if we're not careful, we'll end up *emphasizing* the topology with highly visible patterns rather than merely rolling with it, which I don't think is desirable; we want terrain features at many scales, with no one scale having too much primacy.

I think quantization would produce the intended effect, and would in fact significantly increase the potential variety of terracing intensities. The implementation would also be simpler and more efficient, since we'd be quantizing the interpolated value rather than using an ad-hoc nonlinear interpolation. Given those advantages, and the absence of any clear drawbacks, I'd like to explore that approach.

Ralith:
Great stuff, thanks!

PR #91, opened 2020-05-23
Prevent road supports from having a bottom

DimpyRed:


client/src/worldgen.rs
```diff
@@ -306,7 +303,8 @@ impl ChunkParams {
         }
 
         if (center_elevation + ELEVATION_MARGIN < me_min / ChunkParams::ELEVATION_SCALE)
-            && !(self.is_road || self.is_road_support) //only test for road support here
+            && !(self.is_road || self.is_road_support)
+        //only test for road support here
```
Ralith, line 307: Can you clarify this comment? I don't understand what it's trying to say; it doesn't seem to describe the actual behavior.

PR #92, opened 2020-05-23
Prevent road supports from having a bottom

DimpyRed:


client/src/worldgen.rs
```diff
@@ -40,6 +40,7 @@ enum NodeStateRoad {
     DeepWest,
 }
 use NodeStateRoad::*;
+use term::terminfo::Error::MalformedTerminfo;
```
Ralith, line 43: Stray import

client/src/worldgen.rs
```diff
@@ -256,6 +258,18 @@ impl ChunkParams {
...
     }
 
+    fn generate_road_support(&self, center: na::Vector3<f64>) -> Option<Material> {
+        let plane = -Plane::from(Side::B);
+        let horizontal_distance = plane.elevation(center, self.chunk);
+
+        if horizontal_distance > 0.3 {
+            return Some(Material::WoodPlanks);
+        }
+        else {
```
Ralith, line 268: Please run clippy/rustfmt

client/src/worldgen.rs
```diff
@@ -256,6 +257,17 @@ impl ChunkParams {
         }
     }
 
+    fn generate_road_support(&self, center: na::Vector3<f64>) -> Option<Material> {
+        let plane = -Plane::from(Side::B);
+        let horizontal_distance = plane.elevation(center, self.chunk);
+
+        if horizontal_distance.abs() < 0.3 {
```
Ralith, line 264: You don't actually need to call `abs()` here. Neighboring nodes have coordinate systems reflected across the common face, so the nodes on either side of the road plane have distances with the same sign.

PR #96, opened 2020-05-24
Make supports non-solid

DimpyRed:


client/src/worldgen.rs
```diff
@@ -256,6 +257,64 @@ impl ChunkParams {
...
+            } else {
+                Some(mat)
+            }
+        } else {
+            None
+        }
+    }
+
+    ///carve out a truss-shaped hole
+    fn generate_trussing(&self, coords: na::Vector3<u8>) -> Option<Material> {
```
Ralith, line 283:
```suggestion
    fn trussing_at(&self, coords: na::Vector3<u8>) -> bool {
```
If we're just using this to determine the shape, rather than any material variation, let's use a more specific return type.

client/src/worldgen.rs
```diff
@@ -256,6 +257,64 @@ impl ChunkParams {
...
+        //Generates planar diagonals, but corner is offset
+        let mut criteria_met = 0_u32;
+        let x = coords[0];
+        let y = coords[1];
+        let z = coords[2];
+
+        //straight lines
+        if x == 8 {
```
Ralith, line 291: Let's not hardcode coordinates; that'll be very fragile as we adjust the chunk size in the future. Instead, compute the desired value from `dimension` dynamically; `2 * dimension / 3` perhaps? Opened #97 to help with this.

client/src/worldgen.rs
```diff
@@ -256,6 +257,64 @@ impl ChunkParams {
...
+                None
+            } else {
+                Some(mat)
+            }
+        } else {
+            None
+        }
+    }
+
+    ///carve out a truss-shaped hole
```
Ralith, line 282: Please put a space after the slashes in your comments.

client/src/worldgen.rs
```diff
@@ -257,19 +264,59 @@ impl ChunkParams {
...
+            criteria_met += 1;
+        }
+        if x == z {
+            criteria_met += 1;
+        }
+
+        if criteria_met >= 2 {
+            return true; // will get angry without return keyword.
+        }
+        false
```
Ralith, line 315:
```suggestion
        criteria_met >= 2
```

client/src/worldgen.rs
```diff
@@ -257,19 +264,59 @@ impl ChunkParams {
...
+            criteria_met += 1;
+        }
+        if y == 2 * self.dimension / 3 {
+            criteria_met += 1;
+        }
+        if z == 2 * self.dimension / 3 {
+            criteria_met += 1;
+        }
+
+        //main diagonal
```
Ralith, line 301: Missing space

client/src/worldgen.rs
```diff
@@ -257,19 +264,59 @@ impl ChunkParams {
...
+        let z = coords[2];
+
+        // straight lines.
+        if x == 2 * self.dimension / 3 {
+            criteria_met += 1;
+        }
+        if y == 2 * self.dimension / 3 {
+            criteria_met += 1;
+        }
+        if z == 2 * self.dimension / 3 {
```
Ralith, line 297: Let's compute this value once and store it in a temporary and reuse that.

client/src/worldgen.rs
```diff
@@ -257,19 +264,59 @@ impl ChunkParams {
...
+        //main diagonal
+        if x == y {
+            criteria_met += 1;
+        }
+        if y == z {
+            criteria_met += 1;
+        }
+        if x == z {
+            criteria_met += 1;
+        }
```
Ralith, line 310: These if statements would be more concise if they were all rewritten as something like `criteria_met += u32::from(foo == bar);`

client/src/worldgen.rs
```diff
@@ -257,19 +264,59 @@ impl ChunkParams {
...
+    fn generate_road_support(
+        &self,
+        center: na::Vector3<f64>,
+        coords: na::Vector3<u8>,
+    ) -> Option<Material> {
         let plane = -Plane::from(Side::B);
         let horizontal_distance = plane.elevation(center, self.chunk);
 
-        if horizontal_distance < 0.3 {
+        if (horizontal_distance < 0.3) && self.trussing_at(coords) {
```
Ralith, line 275:
```suggestion
        if horizontal_distance < 0.3 && self.trussing_at(coords) {
```
Comparisons bind tighter than boolean operations.

client/src/worldgen.rs
DimpyRed: What scope should it have? trussing_at, generate_chunks, or global?

DimpyRed: I'm going to default to function-wide

Ralith: Good choice. In this case we're more concerned about readability/maintainability than efficiency, as the compiler can generally manage the latter fine at this scale.

PR #103, opened 2020-05-26
Flatness

DimpyRed:
The makes a lot of the area flat, but retains some deep hills and valleys.

### DIFF
--- a/client/src/worldgen.rs
+++ b/client/src/worldgen.rs
@@ -462,8 +462,9 @@ impl EnviroFactors {
                 + ((((3 - parent.slopeiness.rem_euclid(7)) as f64)
                     * (1.0 - (((parent.flatness as f64) - 20.0) / 10.0).tanh())
                     + ((3 - slopeiness.rem_euclid(7)) as f64)
-                        * (1.0 - (((flatness as f64) - 20.0) / 10.0).tanh())
-                        ) as i64) * rng.sample(&plus_or_minus_one) ,
+                        * (1.0 - (((flatness as f64) - 20.0) / 10.0).tanh()))
+                    as i64)
+                    * rng.sample(&plus_or_minus_one),
             temperature: parent.temperature + rng.sample(&plus_or_minus_one),
             rainfall: parent.rainfall + rng.sample(&plus_or_minus_one),
             blockiness: parent.blockiness + rng.sample(&plus_or_minus_one),
--- a/client/src/worldgen.rs
+++ b/client/src/worldgen.rs
@@ -71,12 +71,12 @@ impl NodeState {
             road_state: NodeStateRoad::ROOT,
             spice: 0,
             enviro: EnviroFactors {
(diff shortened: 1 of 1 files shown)


DimpyRed:
I want to rewrite the ways factors influence bock types now that flatness is a thing, but I want to make sure this is correct first.

client/src/worldgen.rs
```diff
@@ -439,17 +445,38 @@ struct EnviroFactors {
...
+                        + rng.sample(&plus_or_minus_one)
+                        + ((parent.max_elevation < 0) as i64)
+                        - ((parent.max_elevation > 0) as i64)
+                }
+                31..=35 => {
+                    parent.max_elevation
+                        + (rng.sample(&plus_or_minus_one) + rng.sample(&plus_or_minus_one)) / 2
+                } //not sure about the rounding on this one
+                _ => parent.max_elevation,
+            },
```
Ralith, line 479: @MagmaMcFry, can you comment on whether this is sound?

DimpyRed: > Do we really need both "slopeiness" and "flatness"? Those sound like the same thing.

They are different. Slopiness functions as the derivitve of elevation change, while flatness functions as a suppressant for elevation change.

DimpyRed: In other words, slopeiness is a tool used to make terrain hilly, but flatness determines if it will actually be hilly.

Ralith:
It sounds like they both indeed control the same thing. Could we do that with a single value instead?

client/src/worldgen.rs
```diff
@@ -439,17 +445,26 @@ struct EnviroFactors {
...
+            .min(40);
         let slopeiness = parent.slopeiness + rng.sample(&plus_or_minus_one);
         Self {
             slopeiness,
+            flatness,
             max_elevation: parent.max_elevation
-                + ((3 - parent.slopeiness.rem_euclid(7)) + (3 - slopeiness.rem_euclid(7)))
-                + rng.sample(&plus_or_minus_one),
+                + ((((3 - parent.slopeiness.rem_euclid(7)) as f64)
+                    * (1.0 - (((parent.flatness as f64) - 20.0) / 10.0).tanh())
```
Ralith, line 463: `flatness - 20` has a range of -19 to 20 inclusive. That's asymmetrical, and I think therefore unsound; is it what you intended?

DimpyRed: It does not need to be symmetric. The 20.0 can be adjusted artistically. However, I might as well make it symmetric for neatness's sake.

Ralith: On review, the randomized sign at the end probably means any asymmetry here gets worked out in the end, but I still suspect it wasn't intended.

DimpyRed: Besides, 1 - tanh(x) is always more than zero.

PR #106, opened 2020-06-04
Worldgen move3

DimpyRed:
I still have some cleanup to do, but this is the basic structure.


common/Cargo.toml
```diff
@@ -20,6 +19,8 @@ fxhash = "0.2.1"
 tracing = "0.1.10"
 hecs = "0.2.9"
 tracing-subscriber = { version = "0.2.5", default-features = false, features = ["env-filter", "smallvec", "fmt", "ansi", "chrono", "parking_lot"] }
+rand = "0.7.3"
+rand_pcg = "0.2.1"
```
Ralith, line 23: Can we also remove this dependency from `client`?

common/src/chunk.rs
```diff
@@ -0,0 +1,17 @@
+use crate::lru_slab::SlotId;
+use crate::node::VoxelData;
+
+pub enum Chunk {
```
Ralith, line 4: I'd dump this into the `node` module, since it's so tiny.

common/src/node.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub fn get(&self, index: usize) -> Material {
+        match *self {
+            VoxelData::Dense(ref d) => d[index],
+            VoxelData::Solid(mat) => mat,
+        }
+    }
+}
+
+
+use crate::lru_slab::SlotId;
```
Ralith, line 43: This should be at the top with the other `use`s

common/src/node.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub struct Node {
+    pub state: NodeState,
+    /// We can only populate chunks which lie within a cube of populated nodes, so nodes on the edge
+    /// of the graph always have some `None` chunks.
+    pub chunks: Chunks<Chunk>,
+}
+
+#[derive(PartialEq)]
+pub enum VoxelData {
```
Ralith, line 18: Let's sort this type (and its `impl` block) below `Chunk`, since that's where it's used.

DimpyRed: This is what you were asking for, right?

Ralith: Yep, thanks!

Ralith:
Thanks! This will make a great foundation for server-side physics and should make testing and iterating on the worldgen code more convenient.

PR #113, opened 2020-06-08
Enviros redefine

DimpyRed:
This is not quite ready to be pulled into master, and I'm not sure it should even when completed. However, I would like feedback.

### DIFF
--- a/common/src/worldgen.rs
+++ b/common/src/worldgen.rs
@@ -220,10 +220,8 @@ impl ChunkParams {
                 voxel_mat = Material::Dirt;
             } else if rain < 4.0 {
                 voxel_mat = Material::Grass;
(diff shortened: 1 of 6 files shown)


common/src/worldgen.rs
```diff
@@ -190,29 +190,60 @@ impl ChunkParams {
...
+            voxel_mat = Material::Valite;
+        } else if temp - rain < 8.0 {
+            voxel_mat = Material::Sand
+        } else if temp - rain < 10.0 {
+            voxel_mat = Material::Blackstone;
         } else {
-            voxel_mat = Material::Flowergrass;
-        }
+            voxel_mat = Material::Lava
+        };
```
Ralith, line 246:
```suggestion
        }
```

PR #116, opened 2020-06-12
Dithering of rainfall and temperature

DimpyRed:
This is effectively rebased in the current master.

### DIFF
--- a/common/src/worldgen.rs
+++ b/common/src/worldgen.rs
@@ -9,7 +9,6 @@ use crate::{
     Plane,
 };
 
-
 #[derive(Clone, Copy, PartialEq, Debug)]
 enum NodeStateKind {
     Sky,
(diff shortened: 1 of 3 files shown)


common/src/worldgen.rs
```diff
@@ -259,7 +265,7 @@ impl ChunkParams {
                 }
             } else if (slope_mod >= 3_f64) && (slope_mod <= 4_f64) {
                 voxel_mat = match voxel_mat {
-                    Material::Flowergrass => Material::Bigflowergrass,
+                    //Material::Flowergrass => Material::Bigflowergrass,
```
Ralith, line 268: Dead code?

common/src/worldgen.rs
```diff
@@ -167,83 +169,87 @@ impl ChunkParams {
...
-        let elev_raw = trilerp(&self.env.max_elevations, cube_coords);
-        let terracing_scale = 5.0; // This is not wavelength in number of blocks
-        let elev_floor = (elev_raw / terracing_scale).floor();
-        let elev_rem = elev_raw / terracing_scale - elev_floor;
-        let elev = terracing_scale * elev_floor + serp(0.0, terracing_scale, elev_rem, threshold);
-
-        let mut voxel_mat;
-        let max_e;
-
+    fn choose_base_material(rain: f64, temp: f64) -> Material {
```
Ralith, line 172: This doesn't interact with `ChunkParams` in any way, so it probably makes more sense as a module-level function rather than an associated function.

common/src/worldgen.rs
DimpyRed: Is this alive enough to get merged? I could add some actual implementation for valley areas, but that feels like something for another branch.

Ralith:
Subsumed by #118.

PR #118, opened 2020-06-17
New materials, Voronoi method of picking materials. Refactor of worldgen.rs

danieldeankon:
The main intention of these changes was to replace the current nested if-block method of picking materials with something more flexible and extensible. Now, instead of the if-block, I choose materials using helper functions called terraingen_voronoi and surface_adjuster. These functions pick the material with the help of an array of points with associated materials in a 2D space defined by temperature and rainfall. The arguments of rain and temp are used to produce a new point in the space by picking the material of the closest point from the array. In a single one of these helper functions there are actually four arrays, and the array that is chosen is determined by the value of elev that is passed in, representing distance from the guiding plane. The two helper functions are very similar except for the assignment of materials to points in the arrays. Which one of the helper functions is chosen depends on the difference between the distance to the guiding plane and the elevation envirofactor. When the difference is small, the surface_adjuster function is used, otherwise the terraingen_voronoi function is used.

I have also added a bunch of new materials to demonstrate the capabilities of the new system.

Aside from that, the code related to generate_voxels has been refactored. Much of that function's code has been broken out into separate functions which generate_voxels calls in sequence. I did this with the intention of making generate_voxels more modular.

I have already incorporated the dithering-related changes in https://github.com/Ralith/hypermine/pull/116 into this branch.


common/src/worldgen.rs
```diff
@@ -561,22 +486,20 @@ fn chunk_incident_enviro_factors(
...
+    let (e3, t3, r3, b3) = i.next()?.into();
+    let (e4, t4, r4, b4) = i.next()?.into();
+    let (e5, t5, r5, b5) = i.next()?.into();
+    let (e6, t6, r6, b6) = i.next()?.into();
+    let (e7, t7, r7, b7) = i.next()?.into();
+    let (e8, t8, r8, b8) = i.next()?.into();
 
     Some(ChunkIncidentEnviroFactors {
-        max_elevations: [e1, e2, e3, e4, e5, e6, e7, e8],
+        elevations: [e1, e2, e3, e4, e5, e6, e7, e8],
```
Ralith, line 499: `max_elevation` is a very meaningful name, which I don't think should be changed. I think of "elevation" as "distance from the ground plane", which is not something any noise field controls. The actual meaning of this noise field is the maximum elevation at which a solid voxel should be generated.

common/src/terraingen.rs
```diff
@@ -0,0 +1,996 @@
+use crate::world::Material;
+
+pub struct VoronoiInfo {
+    pub location: na::Vector2<f64>,
```
Ralith, line 4: I'd be inclined to use f32 here to keep the tables smaller unless you specifically need high precision for some reason.

common/src/terraingen.rs
```diff
@@ -0,0 +1,996 @@
...
+impl VoronoiInfo {
+    pub fn new(mat: Material, rain: f64, temp: f64) -> VoronoiInfo {
+        VoronoiInfo {
+            location: na::Vector2::new(rain, temp),
+            material: mat,
+        }
+    }
+
+    pub fn terraingen_voronoi(elev: f64, rain: f64, temp: f64) -> Material {
+        let voronoi_choices_high = vec![
```
Ralith, line 16: Every time you call `Vec`, heap memory is allocated. Given that this is constant data, allocating a fresh copy for every single voxel(!) doesn't make sense. Instead, these could be `const` arrays (in which case `new` will have to be a `const fn`). For obscure annoying reasons `na::Vector2` can't be used in constants, so you'll have to replace it with a `[f32; 2]` or similar, but that shouldn't be too much trouble.

common/src/terraingen.rs
```diff
@@ -0,0 +1,996 @@
...
+        } else if elev < 4.0 {
+            &voronoi_choices_med
+        } else {
+            &voronoi_choices_high
+        };
+
+        let y = na::Vector2::new(rain, temp);
+        let mut dist = na::Matrix::norm(&(voronoi_choices[0].location - y));
+        voxel_mat = voronoi_choices[0].material;
+        for vc in voronoi_choices.iter().skip(1) {
```
Ralith, line 495: I'm concerned that we're iterating through a fairly large number of values here brute-force. We do have some spare CPU to throw around in worldgen for the time being, but this feels like squandering it. Could we precompute results into an NxNxN 3D array or similar instead? That would be *much* faster.

Relatedly, it might be useful to create a benchmark for the generation of a chunk (say, the origin chunk).

common/src/worldgen.rs
danieldeankon: > The actual meaning of this noise field is the maximum elevation at which a solid voxel should be generated.

Makes sense. I've changed it back.

common/src/terraingen.rs
danieldeankon: I used f64 because the envirofactors are f64. I've made the struct definition of location use [f32; 2] now, converting using `as` when f64 values are passed in. Perhaps there's an argument for the envirofactors themselves being f32 instead of f64---I don't see why they need to be particularly precise, and the extra memory used adds up across the thousands of nodes.

common/src/terraingen.rs
danieldeankon: Made this change. I've also combined the terraingen_voronoi and surface_adjuster functions into one function which now takes an extra parameter dist representing depth below the ground surface (not the guiding plane).

common/src/terraingen.rs
danieldeankon: I have not done any precomputation. That would probably be something I'd need help with. I am also unsure about the amount of space required for the array. Suppose for temp, rain, and elev in the range of (-10.0, 10.0) you wanted a resolution of 0.02, yielding 1000 samples, then you'd have a billion coordinates, each storing an integer for the Material enum value. Perhaps 500 or even 200 samples would still look good, which would greatly reduce the memory cost. An alternative is to only sample for rain and temp at a small set of elevations, so the storage would be merely 1000x1000xn. 

For now, I have changed the distance calculation to give distance squared.

common/src/worldgen.rs
danieldeankon: I have also updated the commenting around the terrain-generating functions in worldgen.rs and terraingen.rs.

common/src/terraingen.rs
Ralith: The environment factors were 64-bit because we do nothing to ensure they don't grow without bound, basically. That might not be a huge issue if it can be shown that it would take an unreasonable amount of travel to observe any issues, though.

common/src/terraingen.rs
```diff
@@ -0,0 +1,992 @@
...
+    // This function picks the material with the help of arrays of points in
+    // 2D spaces defined by temp and rain with associated materials for each point.
+    // The values of elev and dist determine which of eight arrays is chosen.
+    // dist represents how far from the surface of the terrain a point is.
+    // For values of dist between 0 and TERRAIN_SURFACE_THICKNESS, the four surface strata
+    // arrays are used. Otherwise, the four general terrain strata arrays are used.
+    // Variations between surface and terrain are intended to represent erosion and vegatation.
+    // elev represents distance from the guiding plane. There are four strata of
+    // terrain that are defined by elev thresholds. Variations between strata
+    // are intended to represent the effects of more or less skylight being exposed.
```
Ralith, line 954: Nice explanation!

common/src/worldgen.rs
```diff
@@ -340,6 +320,82 @@ impl ChunkParams {
...
+            self.neighbor(coords, 1, 0, 0, &voxels),
+            self.neighbor(coords, 0, -1, 0, &voxels),
+            self.neighbor(coords, 0, 1, 0, &voxels),
+            self.neighbor(coords, 0, 0, -1, &voxels),
+            self.neighbor(coords, 0, 0, 1, &voxels),
+        ]
+    }
+
+    fn neighbor(
+        &self,
```
Ralith, line 335: Why was this made a method? It doesn't seem to need access to `self`.

common/src/worldgen.rs
```diff
@@ -340,6 +320,82 @@ impl ChunkParams {
         criteria_met >= 2
     }
 
+    fn voxel_neighbors(&self, coords: na::Vector3<u8>, voxels: &VoxelData) -> [NeighborData; 6] {
```
Ralith, line 323: Why does this need to be a method?

common/src/worldgen.rs
danieldeankon: I assume that with this and the other question, you are asking why they are methods as opposed to functions? There's no particular reason they need to be methods. The only part of `self` that they collectively needed was `self.dimension`. I've made the chunk size an additional argument of both functions and moved them out of `impl ChunkParams`.

common/src/worldgen.rs
danieldeankon: Same as above

common/src/worldgen.rs
Ralith: Whoops, I didn't actually see `dimension` on them; it was late. That's a fine reason for them to be methods; feel free to roll back that last commit.

danieldeankon: It's done, also rebased on top of the other PR that you closed.

common/src/worldgen.rs
```diff
@@ -167,154 +169,132 @@ impl ChunkParams {
...
+                    }
+                    if elevation < 0.075 {
+                        // Surface
+                        if horizontal_distance < 0.15 {
+                            // Inner
+                            mat = Material::WhiteBrick;
+                        } else {
+                            // Outer
+                            mat = Material::GreyBrick;
+                        }
```
Ralith, line 263: This whole clause looks like a no-op; `mat` is not read from.

danieldeankon: It can't be a no-op---running the application for the PR branch shows that the road still exists, and the generate_road function is the only place that WhiteBrick and GreyBrick materials are used. The logic was expressed in a confusing way, so I've tidied it up and made it short circuit in that function and in generate_road_support.

common/src/worldgen.rs
```diff
@@ -167,154 +169,132 @@ impl ChunkParams {
...
+                        dist_pre_noise
+                    };
+
+                    let voxel_mat = if dist >= 0.0 {
+                        VoronoiInfo::terraingen_voronoi(elev, rain, temp, dist)
+                    } else {
+                        Material::Void
+                    };
+
+                    voxels.data_mut(self.dimension)[index(self.dimension, coords)] = voxel_mat;
```
Ralith, line 232: `data_mut` implicitly allocates memory for the voxel chunk, so we shouldn't call it unless we know we're going to assign something other than `Void`, to avoid wasting space on the empty chunks that the short circuit didn't catch.

danieldeankon: I've now made it so data_mut() is only called when voxel_mat has been assigned a non-void material.

Issue #73, opened 2020-04-22
Factor graph traversals out of the actual Graph data structure


Ralith:
`nearby_nodes` and `ensure_nearby` can be implemented purely through the public interface of `Graph`, and therefore in the interest of encapsulation should be. This will pave the way for specializing traversals to suit more specific purposes, as in #49 and #50.

danieldeankon:
I'm interested in working on this one. I'll have time to get stuck into it over the weekend.

PR #119, opened 2020-06-21
Graph traversals

danieldeankon:
Closes https://github.com/Ralith/hypermine/issues/73


server/src/sim.rs
```diff
@@ -177,7 +181,7 @@ impl Sim {
     }
 }
 
-enum Empty {}
+pub enum Empty {}
```
Ralith, line 184: Why was this `pub`ed?

server/src/sim.rs
```diff
@@ -193,9 +197,96 @@ fn dump_entity(world: &hecs::World, entity: Entity) -> Vec<Component> {
...
     speed: f32,
 }
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+
+    #[test]
+    fn rebuild_from_tree() {
```
Ralith, line 239: On review, I think these tests still belong in the `common` crate, since they're more about testing graph/cursor logic than the traversal. Maybe move `ensure_nearby` into a new module in `common` and restore these two tests to their original location?

server/src/sim.rs
danieldeankon: When moving ensure_nearby out of the impl block, I made it take a &mut Graph<Empty> as its first parameter because I saw that other code used the Empty type in the template. Then I got the complaint:

```
private type `sim::Empty` in public interface
can't leak private type rustc(E0446)
```

and I made Empty public because I thought that was better than making ensure_nearby not public.

Playing around with the code now, I can see that the function can simply continue to be a template, which is closer to how it originally was, and the enum doesn't need to be made public. Like this: `pub fn ensure_nearby<N>(graph: &mut Graph<N>, start: &Position, distance: f64)` I think I had tried getting that to work earlier but couldn't because I was confused about there being a `Graph` and a `Graph<N>` in two different places in the code. I've made a commit that addresses this point.

server/src/sim.rs
danieldeankon: I've done as you requested. ensure_nearby is now found in traversal.rs in common, and the two tests I moved are back in their original locations in graph and cursor.

I'm not too clear on what the vision is for this change. If ensure_nearby is just going back to a module in common, why have it in a new module instead of just putting it in graph again? Should nearby_nodes also live in traversals?

Ralith: The general idea is to limit the visibility of things as narrowly as possible to the areas that need them. This applies both to the traversals themselves (so it's more obvious what they're for and more natural to make them even more application-specific in the future) and to the internals of `Graph`, which we're protecting from the traversal code that doesn't actually need them by making them no longer methods.

common/src/cursor.rs
```diff
@@ -128,7 +128,7 @@ lazy_static! {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::proto::Position;
+    use crate::{cursor, proto::Position, traversal::ensure_nearby};
```
Ralith, line 131: `crate::cursor` and `super` are the same thing in this context, so there's no need to import `crate::cursor` or qualify things with it below.

danieldeankon: Pushed an amend commit that addresses this.

PR #126, opened 2020-08-11
Cliffs

DimpyRed:
This isn't ready to be pulled, but I think it's at the point where it could use some looking-over.


DimpyRed:
The cliff system is a lot more dynamic now. It should be pretty easy to add displacement noise relative to their faces. The probablilites of generating them are simpler to adjust.

Cliffs will no longer cross.

The occlusion is doesn't get rid of anything it shouldn't, but still removes the center of plateaus. It's on a per-node basis right, now. Performance could further be improved by moving to a to-chunk basis.

PR #132, opened 2020-09-19
Worldgen voxel iterable

danieldeankon:
Syntactic sugar for looping over the voxels in a chunk in world generation. There is also a bit of extra commenting, and reordering of functions in worldgen.rs to place the higher level functions higher up in the code file.

### DIFF
--- a/common/src/worldgen.rs
+++ b/common/src/worldgen.rs
@@ -130,57 +130,38 @@ impl NodeState {
     }
 }
 
-type VoxelCoords = (u8, u8, u8);
-
-struct VoxelIterable {
-    coords: VoxelCoords,
+struct VoxelCoords {
+    counter: u32,
     dimension: u8,
 }
 
-impl VoxelIterable {
+impl VoxelCoords {
     fn new(dimension: u8) -> Self {
-        VoxelIterable {
-            coords: (0, 0, 0),
-            dimension: dimension - 1,
+        VoxelCoords {
+            counter: 0,
+            dimension,
         }
     }
 }
 
-impl Iterator for VoxelIterable {
-    type Item = VoxelCoords;
+impl Iterator for VoxelCoords {
+    type Item = (u8, u8, u8);
 
     fn next(&mut self) -> Option<Self::Item> {
-        let (x, y, z) = self.coords;
-        let mut x_new = x;
-        let mut y_new = y;
-        let z_new;
+        let dim = u32::from(self.dimension);
 
-        let mut finished = false;
-
-        if z >= self.dimension {
-            z_new = 0;
-            if y >= self.dimension {
-                y_new = 0;
-                if x >= self.dimension {
-                    x_new = 0;
-                    finished = true;
-                } else {
-                    x_new = x + 1;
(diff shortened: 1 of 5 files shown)


common/src/worldgen.rs
```diff
@@ -130,19 +130,75 @@ impl NodeState {
     }
 }
 
+type VoxelCoords = (u8, u8, u8);
+
+struct VoxelIterable {
+    coords: VoxelCoords,
```
Ralith, line 136: This would be much simpler if it was just a `counter: u32`, with `Iterator::next` being something like
```rust
if self.counter == u32::from(self.dimension).pow(3) { return None; }
let result = (self.counter % self.dimension, self.counter / self.dimension, self.counter / self.dimension.pow(2));
self.counter += 1;
Some(result)
```

common/src/worldgen.rs
```diff
@@ -169,65 +225,105 @@ impl ChunkParams {
...
-                        // been void are changed to a material---so no floating dirt blocks.
-                        dist_pre_noise
-                    };
-
-                    if dist >= 0.0 {
-                        let voxel_mat = VoronoiInfo::terraingen_voronoi(elev, rain, temp, dist);
-                        voxels.data_mut(self.dimension)[index(self.dimension, coords)] = voxel_mat;
-                    }
-                }
+        for (x, y, z) in &mut VoxelIterable::new(self.dimension) {
```
Ralith, line 282: You don't need `&mut` here (or in any of the similar places below).

Ralith (changes requested):

common/src/worldgen.rs
danieldeankon: Made the change.

common/src/worldgen.rs
danieldeankon: Made this change. 

Is there a way I could have avoided the conversion from u8 to u32 and back to u8? I always feel like I've painted myself into a corner whenever I end up having to do type conversions just for a calculation in one place, but I don't want to do things that may have far-reaching consequences such as changing the type of ChunkParams.dimension.

common/src/worldgen.rs
```diff
@@ -130,19 +130,56 @@ impl NodeState {
     }
 }
 
+struct VoxelIterable {
```
Ralith, line 133:
```suggestion
struct VoxelCoords {
```
Maybe? Just a thought.

Ralith (approved):
Don't stress too hard about numeric conversions, they're unlikely to matter one way or another in these cases; pretty much just a cosmetic matter until you're storing large numbers of something, assuming you keep an eye on ranges.

common/src/worldgen.rs
danieldeankon: I think that would be an improvement. I'm trying to view it from the point of view of someone typing `for (x, y, z) in VoxelCoords::new(self.dimension) {}` in a new worldgen function, making it something that's simple enough to remember without looking it up.

Ralith (approved):
Nice cleanup!

PR #135, opened 2020-10-15
Added plateus/cliffs

DimpyRed:
Replaces my previous pull request on the same subject


DimpyRed:
@Ralith Would you prefer me to focus first on abstracting cliffs to make it easier to add similar structures, or expanding their functionality to allow for things like face-wise bumpiness and improved occlusion?

PR #195, opened 2022-01-24
Update sim_config.rs

stingpie:
You can now choose the curvature on startup.

### DIFF
--- a/common/src/sim_config.rs
+++ b/common/src/sim_config.rs
@@ -39,8 +39,17 @@ pub struct SimConfig {
 }
 
 impl SimConfig {
+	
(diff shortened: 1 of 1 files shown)


Ralith:
Thanks for your interest! This is already possible by specifying `chunk_size` and `voxel_size` in the configuration file. You can infer this from their presence in the configuration file parsing code.

A few notes for future PRs:
- Please maintain good style, e.g. consistent indentation. You can configure most editors to auto-format on save, or just run `cargo fmt` before committing.
- Stopping to read from standard input in random methods is not usually a good idea. For example, it makes automated testing difficult. In hypermine, parameters should be read from the appropriate config file.

PR #204, opened 2022-03-20
Fix outdated and incorrect code documentation (along with a minor code change)

patowen:
These are things I found while studying the code base earlier. I believe that these changes are all correct and will make the code easier to understand, but, needless to say, please check my work before merging.

There is one actual code change as well, which is the reduction of an array size from 3 to 2 in `populate_shorter_neighbors_of_child`. The purpose of the change was to stay consistent with the documentation. This did not cause any bugs based on my manual testing.


Ralith (approved):
Thanks, this looks great!

PR #208, opened 2022-03-22
Update various dependencies

danieldeankon:


client/src/graphics/gltf_mesh.rs
```diff
@@ -417,9 +417,22 @@ async fn load_material(
...
+        let info = color_reader.info();
+        png::OutputInfo {
+            width: info.width,
+            height: info.height,
+            color_type: ct,
+            bit_depth: bits,
+            line_size: color_reader.output_line_size(info.width),
+        }
+    };
+
```
Ralith, line 435: Why not use the result of `color_reader.info()` directly?

danieldeankon: The method `read_info()` on a `Decoder` used to return a `Result<(OutputInfo, Reader<R>), DecodingError>` in version 0.16.8 but now in version 0.17.5 it returns only `Result<Reader<R>, DecodingError>`. Since the following code needed that `OutputInfo`, I had to produce it in the lines following the call to `read_info()`.

I couldn't use the result of `color_reader.info()` because its return type is not the `OutputInfo` struct formerly returned by the `read_info()` method. Instead it's some other `Info` struct which, going by a short dig into the code for the png crate, I think contains a reference to the `buffer: &[u8]` from the argument of the function being modified here.

If I replace lines 424 to 434 with the line `let color_info = color_reader.info();`, then I get this error:

```
    Checking client v0.1.0 (/home/mint/git/hypermine/client)
error[E0759]: `buffer` has an anonymous lifetime `'_` but it needs to satisfy a `'static` lifetime requirement
   --> client/src/graphics/gltf_mesh.rs:382:5
    |
382 |     buffer: &[u8],
    |     ^^^^^^  ----- this data with an anonymous lifetime `'_`...
    |     |
    |     ...is used here...
...
470 |             .run(move |xf, cmd| {
    |              --- ...and is required to live as long as `'static` here
    |
note: `'static` lifetime requirement introduced by this bound
   --> /home/mint/.cargo/git/checkouts/lahar-354255800778016f/fbc889a/src/transfer.rs:21:74
    |
21  |         f: impl FnOnce(&mut TransferContext, vk::CommandBuffer) + Send + 'static,
    |                                                                          ^^^^^^^

For more information about this error, try `rustc --explain E0759`.
error: could not compile `client` due to previous error
```

If I was able to replace the lines with the line `let color_info = color_reader.output_info();`, I believe that would give me what I need, but that method is private. So I copied the code near the end of `read_info()` from version 0.16.8 of the png crate to create the `OutputInfo`.

danieldeankon: I see. I've made that change.

PR #217, opened 2022-06-19
Upgrade ash and lahar

patowen:
This PR contains the following two related changes:
- Upgrade ash from 0.31 to 0.37.0
- Upgrade lahar from from fbc889a (May 22 2020) to 88abd75e (June 19 2022)

To limit the scope of these changes, functionality removed from lahar has been added to a module in `client` called `lahar_deprecated` with minor changes for compatibility.

While making these changes, I had attempted several other things which I have since reverted, which makes the commit history messy, but I believe the plan is to squash these commits, so I made no attempt to clean that up.


Ralith (approved):
Looks reasonable, thanks!

PR #220, opened 2022-06-19
Tokio and quinn upgrade

patowen:
This PR contains the following related changes:
- Upgrade tokio from 0.2.13 to 1.18.2
- Upgrade quinn from 0.6.1 to 0.8.3
- Upgrade rustls from 0.17.0 to 0.20.6

Due to non-backwards-compatible changes in the way quinn/rustls handles certificates, some logic in that area needed to change.

A few items still need to be completed before I'm fully comfortable merging this.
- [x] Test that a client can connect to a server when the server gets a certificate from a file.
- [x] Consider documenting certificate generation commands that can be used to test.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -19,7 +19,7 @@ ash-window = "0.10.0"
 directories = "4.0.1"
 vk-shader-macros = "0.2.5"
 na = { package = "nalgebra", version = "0.19" }
-tokio = { version = "0.2.13", features = ["rt-threaded", "sync", "macros"] }
+tokio = { version = "1.18.2", features = ["rt-multi-thread", "sync", "macros"] }
 png = "0.17.5"
 anyhow = "1.0.26"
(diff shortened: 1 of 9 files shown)


patowen:
I was able to get a certificate pair that the server accepted with the help of the following openssl command in Windows git bash:
```
MSYS_NO_PATHCONV=1 openssl req -new -x509 -nodes -days 365 -text -subj "/CN=localhost" -extensions v3_req \
     -keyout server.key -out server.crt
```
followed by converting `server.crt` into a pem file.

Given that I believe multiplayer is very much a work in progress, I'm happy just using this PR comment as documentation of a working certificate generation command.

patowen:
True. I had tested that first, but the ability to set up your own certificate pair is a feature that I could have broken, so I wanted to test that to ensure that this PR didn't cause a regression.

Ralith:
Fair enough; thanks for the rigor!

PR #221, opened 2022-06-25
Ensure that recv_whole respects size_limit

patowen:
There is currently a bug in `recv_whole` that allows `cursor` to overshoot `size_limit`, causing `recv_whole` to be able to keep receiving data indefinitely (until `buf.reserve` panics).

I was able to reproduce this bug by artificially reducing `size_limit` and using `println!` to see what's going on. Unfortunately, without making more major changes to `codec.rs`, I am unable to add unit tests, as there isn't an easy way to generate a `quinn::RecvStream` in a test context.

I'm not familiar with the `quinn` library, so it may now be possible to make `recv_whole` practically a one-liner, it which case it would probably make the most sense to do that refactor now instead of just fixing the bug.

The actual bug fix is in the assignment to `slice`. I believe the change to the `buf.reserve` call makes the code feel more logical to me, and may be better in edge cases (`size_limit == isize::MAX`), but due to the lack of guarantees of the `reserve` method, it does not address the underlying issue this PR addresses.

EDIT: This pull request now just replaces the entire implementation with a call to the built in `RecvStream::read_to_end` method.

### DIFF
--- a/common/src/codec.rs
+++ b/common/src/codec.rs
@@ -47,34 +47,8 @@ pub async fn send_whole<T: Serialize + ?Sized>(
 /// Receive the entirety of `stream` as a `T`
 pub async fn recv_whole<T: DeserializeOwned>(
     size_limit: usize,
-    mut stream: quinn::RecvStream,
+    stream: quinn::RecvStream,
 ) -> Result<T> {
-    let mut buf = Vec::<u8>::with_capacity(1024);
-    let mut cursor = 0;
-    loop {
-        let slice = unsafe {
-            std::slice::from_raw_parts_mut(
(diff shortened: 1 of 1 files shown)


Ralith:
Good catch! This code was actually unsound from the start due to constructing a reference to uninitialized memory. We should probably just replace everything with [`RecvStream::read_to_end`](https://docs.rs/quinn/latest/quinn/struct.RecvStream.html#method.read_to_end).

patowen:
Good point. Unsafe Rust still trips me up, so I thought the code was sound despite having already learned about this kind of thing as a common gotcha. I went ahead and followed your suggestion, as it has the additional benefit of no longer being harmed by a lack of unit tests.

PR #230, opened 2022-08-13
Upgrade all dependencies mentioned by dependabot

patowen:
### Summary
This PR upgrades the following crates (comments in parentheses):
- metrics (complicated)
- ash-window (mechanical)
- hecs (mechanical)
- nalgebra (mechanical. Was blocked by [a performance regression in nalgebra](https://github.com/dimforge/nalgebra/issues/1139), but this has now been fixed.)
- winit (mechanical)
- approx (trivial)
- rand (trivial)
- rand_distr (trivial)
- rand_pcg (trivial)
- rcgen (trivial)
- tracing (trivial)
- tracing-subscriber (trivial, apart from changing "chrono" feature to "time")
- webpki (trivial)
- metrics-core (removed, not upgraded, since "metrics" no longer uses it)

One semi-unrelated change is adding `debug-assertions = true` in the top-level Cargo.toml. [Debug assertions are only enabled by default in non-optimized builds](https://doc.rust-lang.org/std/macro.debug_assert.html), so for them to ever be triggered, they need to be enabled manually. I discovered this while debugging the nalgebra upgrade.

The below sections summarize the code changes for each crate.

Although a lot of crates were upgraded, I don't believe the diff is too complicated to review. However, the individual commits are relatively clean, so looking at these may help simplify the code review. I did create the PR under the assumption that the commits would be squashed, though.

### metrics
This is the most complicated change, as the API was significantly adjusted in the metrics crate. The `timing!` macro was removed and replaced with a `histogram!` macro. The `histogram!` macro both "registers" a histogram and records to it. I found the documentation of the crate a bit lacking, but my understanding is that registration is meant to be idempotent.

I made registration practically a no-op, just returning an object that had enough information to implement "recording" the same way as it was implemented before. Fortunately, the PR diff highlights this well.

One other major change is that the metrics crate uses `f64` instead of `u64` now, and the conversion of an `Interval` to an `f64` using this crate uses a unit of seconds, so I had the calls to `histogram!` use seconds, while I had the `hdrhistogram` use nanoseconds as before.

Fortunately, besides switching `metrics_core::Key` to `metrics::Key`, I didn't need to change the `Recorder` struct itself.

Another minor note: the `std` feature option was removed in version [0.18.0](https://github.com/metrics-rs/metrics/blob/64d5c5ae63babafbf3875f723ed263efe9332647/metrics/CHANGELOG.md#removed).

### ash-window
The required parameters for `ash_window::enumerate_required_extensions` and `ash_window::create_surface` have changed to use `RawDisplayHandle` and `RawWindowHandle` objects directly instead of relying on traits. This required adding an additional dependency: `raw_window_handle`. Implementing this change just involved following the example provided in the ash_window git repo.

### hecs
The main change in the hecs crate was the replacement of `get::<T>` and `get_mut::<T>` with `get::<&T>` and `get::<&mut T>` respectively. I believe this change was introduced in [0.8.0](https://github.com/Ralith/hecs/blob/3f530401e24baa862369b14e74aed69cb887e5c8/CHANGELOG.md#changed-1).

### nalgebra
The most repetitive code modification was changing `na::RealField` to `na::RealField + Copy`, as the `Copy` trait was removed from `na::RealField`, and we depended on that in a lot of places. I systematically made this change in all usages. This was a [breaking change in 0.29.0](https://github.com/dimforge/nalgebra/blob/daa2ea33b21b083d781a0f3c169cf15381d1e37e/CHANGELOG.md#breaking-changes-2).

Another code change was replacing some occurrences of `na::U3` with `3`, which is due to [version 0.26.0](https://github.com/dimforge/nalgebra/blob/daa2ea33b21b083d781a0f3c169cf15381d1e37e/CHANGELOG.md#0260) introducing the usage of const generics.

The final code change was that `nalgebra` is now imported in `Cargo.toml` as itself instead of `na` to make its procedural macros introduced in 0.27.0 usable. An `extern crate` declaration was added to continue allowing the use of `na`. Actually using these macros is out of scope for this PR.

Unfortunately, 0.26.0 also came with a performance regression in nalgebra that makes Hypermine laggy with a 5x CPU slowdown on my computer. The fix is in https://github.com/dimforge/nalgebra/pull/1140, which is now present in 0.31.2.

### winit
`set_cursor_grab` takes an enum insteasd of a boolean, but the change is relatively straightforward. The code for grabbing the cursor came straight from the method documentation. `ash-window` was not compatible with winit 0.27.1, but patch 0.27.2 added in backwards compatibility, making the upgrade possible.

### tracing-subscriber
The "chrono" feature option was made unavailable, and based on https://github.com/tokio-rs/tracing/pull/1646, I believe it was replaced with "time". The justification was that the maintenance status of "chrono" was uncertain, causing support for it to be dropped.

### All other crates
For all other crates, only `cargo.toml` was modified, making the upgrade easy. Due to interdependence between crates, I believe some had to be updated at the same time (like nalgebra and rand), but the upgrade process itself was trivial.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -15,7 +15,7 @@ tracing = "0.1.10"
 ash = { version = "0.37.0", features = ["loaded"] }
 lahar = { git = "https://github.com/Ralith/lahar", rev = "88abd75e41d04c3a4199d95f581cb135f5962844" }
 winit = "0.26.1"
-ash-window = "0.10.0"
+ash-window = "0.11.0"
 directories = "4.0.1"
 vk-shader-macros = "0.2.5"
 na = { package = "nalgebra", version = "0.31.2" }
@@ -30,7 +30,7 @@ downcast-rs = "1.1.1"
 quinn = "0.8.3"
 futures-util = "0.3.1"
 rustls = { version = "0.20.6", features = ["dangerous_configuration"] }
-webpki = "0.21.0"
+webpki = "0.22.0"
 hecs = "0.9.0"
 rcgen = { version = "0.9.2", default-features = false }
 memoffset = "0.6"
@@ -44,7 +44,7 @@ default = ["use-repo-assets"]
 use-repo-assets = []
 
 [dev-dependencies]
-approx = "0.3.2"
+approx = "0.5.1"
 bencher = "0.1.5"
 renderdoc = "0.10.1"
 
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -14,7 +14,7 @@ server = { path = "../server" }
 tracing = "0.1.10"
 ash = { version = "0.37.0", features = ["loaded"] }
 lahar = { git = "https://github.com/Ralith/lahar", rev = "88abd75e41d04c3a4199d95f581cb135f5962844" }
-winit = "0.26.1"
+winit = "0.27.2"
 ash-window = "0.11.0"
 directories = "4.0.1"
 vk-shader-macros = "0.2.5"
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -35,8 +35,7 @@ hecs = "0.9.0"
 rcgen = { version = "0.9.2", default-features = false }
 memoffset = "0.6"
 gltf = { version = "1.0.0", default-features = false, features = ["utils"] }
-metrics = { version = "0.12.1", features = ["std"] }
-metrics-core = "0.5.2"
+metrics = { version = "0.20.1" }
 hdrhistogram = { version = "7", default-features = false }
 
 [features]
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -18,7 +18,7 @@ winit = "0.26.1"
 ash-window = "0.10.0"
 directories = "4.0.1"
 vk-shader-macros = "0.2.5"
-na = { package = "nalgebra", version = "0.19" }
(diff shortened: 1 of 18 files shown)


client/Cargo.toml
```diff
@@ -11,7 +11,7 @@ license = "Apache-2.0 OR Zlib"
 [dependencies]
 common = { path = "../common" }
 server = { path = "../server" }
-tracing = "0.1.10"
+tracing = "0.1.36"
```
Ralith, line 14: There's rarely any need to update patch versions manually; newer patch versions are always considered compatible and preferred.

patowen: Thanks for pointing that out. I had known that before, but I must have missed that for `tracing`. Should I revert this change, or leave it as-is, since it's already part of the PR?

Ralith: It's harmless, so I don't really care, but it should also be very easy to drop via interactive rebase. I'm assuming this PR is subject to change regardless as you have it marked as a draft.

patowen: Makes sense to me. I believe the PR is mostly complete, as the only change that's left is to upgrade the version of nalgebra. Everything should be fair game for review. I'll go ahead and do the rebase.

patowen:
Now that nalgebra has been updated, this is finally ready to review and merge.

patowen:
Thanks for merging!

PR #232, opened 2022-08-20
Refactor dodeca.rs to make math easier to understand

patowen:
This PR has a major breaking change in dodeca.rs's API. In chunk coordinates, 0 and 1 are swapped from how they were before. This has a few advantages:
- The origin in chunk coordinates is at the center of the Beltrami-Klein model, making the mapping of these coordinates to Hyperbolic space more intuitive.
- Relatedly, the chunk coordinates give an idea of how distorted we are from a Euclidean cube, as higher-magnitude chunk coordinates have higher distortion, and everything is nicely behaved near the origin.
- Collision math should be slightly simpler due to the above points.

A potentially notable disadvantage is that the origin of a node and the origin of a chunk no longer coincide. However, I believe this is minor.

A few other helpful features have been added to dodeca.rs
- `Side` has a `normal()` method which can retrieve the outward facing normal rather than relying on the entire reflection matrix.
- In addition to "node" and "chunk" coordinates, there are also "dual" coordinates, named that way because the origin is mapped to the center of the cube in the cubic honeycomb (dual to the dodecahedral honeycomb), and the axes point directly towards the faces of the cube.
    - The conversion between "node" and "dual" coordinates is an isometry, which makes it easier to understand and work with than "chunk" coordinates.
    - The conversion between "dual" and "chunk" coordinates simply requires scaling the `x`, `y`, and `z` values by a constant factor: `Vertex::dual_to_chunk_factor()`.

I also renamed `faces` to `is_facing` because that change is relatively small and non-controversial, so it makes sense to tag it along during a dodeca.rs refactor.

The following screenshot has a 2D example (in the Beltrami-Klein model) of what chunk coordinates will look like after this PR is merged. It is centered at a node's origin. If it was centered at a chunk's origin instead, the grid of that chunk would be uniform:
image

### DIFF
--- a/common/src/dodeca.rs
+++ b/common/src/dodeca.rs
@@ -133,14 +133,12 @@ impl Vertex {
 
     /// Transform from euclidean chunk coordinates to hyperbolic node space
     pub fn chunk_to_node(self) -> na::Matrix4<f64> {
-        let origin = na::Vector4::new(0.0, 0.0, 0.0, 1.0);
-        let [a, b, c] = self.canonical_sides();
-        na::Matrix4::from_columns(&[
-            a.reflection().column(3) - origin,
-            b.reflection().column(3) - origin,
-            c.reflection().column(3) - origin,
-            origin,
-        ]) * na::Matrix4::new_scaling(0.5)
+        self.dual_to_node() * na::Matrix4::new_scaling(1.0 / Self::dual_to_chunk_factor())
+    }
+
+    /// Transform from hyperbolic node space to euclidean chunk coordinates
+    pub fn node_to_chunk(self) -> na::Matrix4<f64> {
+        na::Matrix4::new_scaling(Self::dual_to_chunk_factor()) * self.node_to_dual()
     }
 
     /// Transform from cube-centric coordinates to dodeca-centric coordinates
@@ -151,6 +149,13 @@ impl Vertex {
     /// Transform from dodeca-centric coordinates to cube-centric coordinates
     pub fn node_to_dual(self) -> &'static na::Matrix4<f64> {
         &NODE_TO_DUAL[self as usize]
+    }
+
+    /// Scale factor used in conversion from cube-centric coordinates to euclidean chunk coordinates.
+    /// Scaling the x, y, and z components of a vector in cube-centric coordinates by this value
+    /// and dividing them by the w coordinate will yield euclidean chunk coordinates.
+    pub fn dual_to_chunk_factor() -> f64 {
+        (2.0 + 5.0f64.sqrt()).sqrt()
     }
 
     /// Convenience method for `self.chunk_to_node().determinant() < 0`.
@@ -328,7 +333,7 @@ mod tests {
 
     #[test]
     fn radius() {
-        let corner = Vertex::A.chunk_to_node() * na::Vector4::repeat(1.0);
+        let corner = Vertex::A.chunk_to_node() * math::origin();
         assert_abs_diff_eq!(
             BOUNDING_SPHERE_RADIUS,
             math::distance(&corner, &math::origin()),
@@ -338,6 +343,28 @@ mod tests {
         assert_abs_diff_eq!(
             BOUNDING_SPHERE_RADIUS,
             (1.5 * phi).sqrt().asinh(),
+            epsilon = 1e-10
+        );
+    }
+
+    #[test]
+    fn chunk_to_node() {
+        // Chunk coordinates of (1, 1, 1) should be at the center of a dodecahedron.
+        let mut chunk_corner_in_node_coordinates =
+            Vertex::A.chunk_to_node() * na::Vector4::new(1.0, 1.0, 1.0, 1.0);
+        chunk_corner_in_node_coordinates /= chunk_corner_in_node_coordinates.w;
+        assert_abs_diff_eq!(
+            chunk_corner_in_node_coordinates,
+            na::Vector4::new(0.0, 0.0, 0.0, 1.0),
+            epsilon = 1e-10
+        );
+    }
+
+    #[test]
+    fn node_to_chunk() {
+        assert_abs_diff_eq!(
+            Vertex::A.chunk_to_node().try_inverse().unwrap(),
+            Vertex::A.node_to_chunk(),
             epsilon = 1e-10
         );
     }
--- a/common/src/dodeca.rs
+++ b/common/src/dodeca.rs
@@ -57,7 +57,7 @@ impl Side {
 
     /// Whether `p` is opposite the dodecahedron across the plane containing `self`
     #[inline]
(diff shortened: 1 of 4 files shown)


Ralith:
> 0 and 1 are swapped from how they were before.

Where is the origin now in relation to the dodecahedral honeycomb?

> The origin in chunk coordinates is at the center of the Beltrami-Klein model

I'm not sure what this means. IIRC chunks have no particular relationship to the Klein model to begin with, that's only used for projection. We do most of our work in the hyperboloid model. 

> The conversion between "node" and "dual" coordinates is an isometry
> The conversion between "dual" and "node" coordinates simply requires scaling the x, y, and z values

This is a contradiction.

patowen:
> Where is the origin now in relation to the dodecahedral honeycomb?

The origin is on the corresponding vertex of the docecahedron.

> I'm not sure what this means. IIRC chunks have no particular relationship to the Klein model to begin with, that's only used for projection. We do most of our work in the hyperboloid model.

If the Hyperboloid coordinates are treated as homogeneous coordinates, they are equivalent to the Beltrami-Klein model. In other words, (x/w, y/w, z/w) are the coordinates in the Beltrami-Klein model.

What I mean by that statement is that the relationship between the Euclidean coordinates and Hyperbolic coordinates is that if you took the Euclidean coordinates, inserted a sphere around the origin and made that the boundary of the Beltrami Klein model, the resulting Hyperbolic coordinates are defined with respect to that model.

In other words, to visualize a chunk in the Beltrami-Klein model, just draw an evenly-divided grid with the chunk's origin at the ball's origin.

> This is a contradiction.

Indeed it is. I missed that in proofreading. The second bullet point should have read "chunk", not "node".

common/src/plane.rs
```diff
@@ -81,7 +80,7 @@ impl Plane<f64> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::math::translate_along;
+    use crate::math::{origin, translate_along};
```
Ralith, line 83: This change seems unrelated to the commit message.

common/src/worldgen.rs
```diff
@@ -266,7 +266,8 @@ impl ChunkParams {
             let trilerp_coords = center.map(|x| (1.0 - x) * 0.5);
 
             let rain = trilerp(&self.env.rainfalls, trilerp_coords) + rng.sample(&normal.unwrap());
-            let temp = trilerp(&self.env.temperatures, trilerp_coords) + rng.sample(&normal.unwrap());
+            let temp =
+                trilerp(&self.env.temperatures, trilerp_coords) + rng.sample(&normal.unwrap());
```
Ralith, line 270: This too.

common/src/dodeca.rs
```diff
@@ -346,4 +346,15 @@ mod tests {
             epsilon = 1e-10
         );
     }
+
+    #[test]
+    fn dual_to_chunk_factor_correct() {
```
Ralith, line 351: This should be included in the commit that introduced the factor.

Ralith (changes requested):
Rather than having "fix tests" as a standalone commit, please ensure the tests are passing after each individual commit, modifying them as needed.

patowen:
> Rather than having "fix tests" as a standalone commit, please ensure the tests are passing after each individual commit, modifying them as needed.

I had assumed that commits would generally get squashed, so I set up the commits in the order that I worked on things. Next time, I'm happy to squash the commits myself before creating a pull request. However, at least in this PR, it's probably possible to separate the changes into distinct units that work on their own, so I'll go ahead and do that. This way, the PR could be merge-able with or without a squash.

common/src/plane.rs
patowen: Surprisingly, this change is related. Since `origin` is no longer used above in `Plane<f64>::from`, `rustc` throws a warning about the import, but a test still uses it, so the import has to be moved down here.

common/src/worldgen.rs
patowen: However, this one is related to ignoring the consistency of individual commits assuming they would be squashed. I'll fix that.

common/src/dodeca.rs
patowen: I decided to replace this test with easier-to-read tests that check that the dodecahedron's origin is in the right place in chunk coordinates. This tests `Vertex::dual_to_chunk_factor()` indirectly.

patowen:
I've updated this PR based on your suggestions. There are now only 3 commits, and all tests/lints should be passing in each of the commits.

Ralith (approved):
Thanks! Logical, self-contained commits that can be rebased directly are easier to review, and if necessary, dig through in the future.

common/src/dodeca.rs
```diff
@@ -127,14 +133,29 @@ impl Vertex {
...
     pub fn chunk_to_node(self) -> na::Matrix4<f64> {
-        let origin = na::Vector4::new(0.0, 0.0, 0.0, 1.0);
-        let [a, b, c] = self.canonical_sides();
-        na::Matrix4::from_columns(&[
-            a.reflection().column(3) - origin,
-            b.reflection().column(3) - origin,
-            c.reflection().column(3) - origin,
-            origin,
-        ]) * na::Matrix4::new_scaling(0.5)
+        self.dual_to_node() * na::Matrix4::new_scaling(1.0 / Self::dual_to_chunk_factor())
```
patowen, line 136: Leaving this comment for posterity:

Before the rebase, there was an intermediate commit where this function was defined as
```rust
let [a, b, c] = self.canonical_sides();
na::Matrix4::from_columns(&[
    a.normal() * a.normal()[3],
    b.normal() * b.normal()[3],
    c.normal() * c.normal()[3],
    math::origin(),
])
```
This still used the old convention but tried to take advantage of the normals. Since the definition got changed again, this version was omitted from history after the rebase.

common/src/dodeca.rs
```diff
@@ -306,4 +346,26 @@ mod tests {
             epsilon = 1e-10
         );
     }
+
+    #[test]
+    fn chunk_to_node() {
```
patowen, line 351: Leaving this comment for posterity:

Before the rebase, instead of these two tests, there was only one test: `dual_to_chunk_factor_correct()` defined as follows:
```rust
// The center of a dodeca should have chunk coordinates of (1, 1, 1)
let corner = Vertex::A.node_to_dual() * math::origin();
assert_abs_diff_eq!(
    Vertex::dual_to_chunk_factor(),
    1.0 / (corner.x / corner.w),
    epsilon = 1e-10
);
```
This test is no longer in the commit history. It was removed because the `chunk_to_node` test is just as effective at catching bugs but is easier to understand.

patowen:
> Thanks! Logical, self-contained commits that can be rebased directly are easier to review, and if necessary, dig through in the future.

That was what I figured. I'll keep that in mind for future PRs.

PR #240, opened 2022-12-03
Package builds in CI

Ralith:


patowen:
The lint is failing because of an ash update. We'll need to increase the required version to 0.37.1 instead of 0.37.0 (since not doing so may break builds for people who already installed 0.37.0) and fix the lint error.

PR #241, opened 2022-12-04
Collision demo

patowen:
# Known bugs
- Walking direction is counterintuitive on slopes.
- On some computers, placing blocks causes chunks to flash with contents of a different chunk.

Note: This won't be merged as-is, since for the purposes of prototyping, a lot of the main code has been messed up. A rewrite is planned once the rest of the bugs and things to experiment on this branch are sorted out.

### DIFF
--- a/client/src/graphics/voxels/mod.rs
+++ b/client/src/graphics/voxels/mod.rs
@@ -103,6 +103,7 @@ impl Voxels {
             sim.graph.get_mut(chunk.node).as_mut().unwrap().chunks[chunk.chunk] =
                 Chunk::Populated {
                     surface: None,
+                    old_surface: None,
                     voxels: chunk.voxels,
                 };
         }
@@ -173,6 +174,7 @@ impl Voxels {
                     }
                     Populated {
                         ref mut surface,
+                        ref old_surface,
                         ref voxels,
                     } => match (surface, voxels) {
                         (&mut Some(slot), _) => {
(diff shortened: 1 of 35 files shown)


client/src/graphics/voxels/mod.rs
```diff
@@ -184,6 +186,15 @@ impl Voxels {
...
                         (&mut ref mut surface @ None, &VoxelData::Dense(ref data)) => {
+                            // Render the previous surface
+                            if let Some(old_slot) = old_surface {
+                                self.states.get_mut(*old_slot).refcount += 1;
+                                frame.drawn.push(*old_slot);
+                                // Transfer transform
+                                frame.surface.transforms_mut()[old_slot.0 as usize] =
+                                    node_transform * chunk.chunk_to_node().map(|x| x as f32);
+                            }
+
```
patowen, line 197: I think I just discovered the root cause of
> On some computers, placing blocks causes chunks to flash with contents of a different chunk.

The issue is that in the `if let Some(lru) = removed {` code below here (Github won't let me put a comment there), Hypermine will set `surface` to `None` for chunks associated with surfaces removed from the LRU. However, this demo doesn't un-associate LRU surfaces with chunks after block updates, so even if `surface` now refers to a different LRU slot, it will still be set to `None`. This can cause a fallback to `old_surface`, which there is no similar set-to-None logic for, so if it's invalidated, `old_surface` might draw an unrelated chunk.

There are many possible fixes, but given how minor the bug is, I don't plan on fixing it here. I'll just keep it in mind when working on the actual PR I hope to merge.

patowen:
Closing as obsolete with the latest Hypermine release: https://github.com/Ralith/hypermine/releases/tag/v0.1.0

PR #244, opened 2022-12-11
Simplify a few functions in math.rs

patowen:
This PR makes the following changes to math.rs that will hopefully make it more convenient to work with. I believe these changes are useful on their own, but the main purpose is to prepare for collision-related changes:

- `math::translate` now requires Lorentz-normalized vectors. Also, it uses a more direct formula rather than composing two reflections. This should hopefully be more efficient, but additional efficiency can be gained by having a special hardcoded function for translating the origin to a particular point.
- `math::translate_along` now only takes a single vector parameter, whose magnitude is how far to translate. Most users of this function benefit from the simplification. For instance, the `sanitize_motion_input` function is much shorter. Internally, it has also been simplified to call `math::translate`.
- A new function `math::mtranspose` has been added to replace `.try_inverse().unwrap()` for isometries. This has the dual benefit of being more efficient (transpose with some minus signs vs general 4x4 matrix inverse) and reducing the number of `unwrap` calls in the code.
- Internally, for readability, `i31` has been replaced with `minkowski_outer_product`, as a counterpart to `mip`.
- Internally, `renormalize_isometry` now uses `translate` instead of `translate_along` to reduce redundant computation.

For the math that `translate` uses, I took the formula of the original `translate_along` function and tried to express it in terms of the origin vector, the vector the origin is translated to, inner/outer products, and the identity matrix, and it ended up simplifying quite well. I'm not sure what the theory behind the formula is, but by symmetry arguments and manual testing, I'm confident that it's correct.

### DIFF
--- a/client/src/graphics/draw.rs
+++ b/client/src/graphics/draw.rs
@@ -268,7 +268,7 @@ impl Draw {
         let draw_started = Instant::now();
         let view = sim.view();
         let projection = frustum.projection(0.01);
-        let view_projection = projection.matrix() * view.local.try_inverse().unwrap();
+        let view_projection = projection.matrix() * math::mtranspose(&view.local);
         self.loader.drive();
 
         let device = &*self.gfx.device;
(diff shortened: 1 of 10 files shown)


Ralith:
CI indicates this has substantially changed the results in at least one case?

Ralith (approved):

PR #247, opened 2022-12-17
Add persistent velocity as a step towards player physics

patowen:
This PR does the following
- Adds persistent velocity and acceleration, which is off by default and can be toggled with "V".
- Puts player movement logic into `character_controller.rs` to separate it from the netcode and prediction logic.
- Various changes/refactors to make this separation of logic possible, some of which will likely need revision.

While I believe this PR is technically in a mergeable state, I do not have that much confidence in the current state of this PR and am definitely open to suggestions on how to improve the readability/maintainability of this code. I have the following specific concerns:

- [x] I have the server pass a `SimConfig` object to the client, which reduces some code repetition, but it also likely adds unwanted coupling between config file organization and the actual simulation configuration.
- [x] The prediction logic and server communication doesn't seem to mesh well with the entity-component abstraction, and it seems like fields are being passed around ad-hoc.
- [x] A lambda is passed into `PredictedMotion`, and a type parameter is used, both in order to allow mocking for the `wraparound()` test. However, the lambda setup seems quite ad-hoc and fragile, and it's not obvious that the type `T` refers to a tick of input.
- [x] The `handle_net` function has a large amount of code repetition. The `update_position` function should probably be inlined into `handle_net` to make the code look more consistent. This is mostly a symptom of the ad-hoc non-entity-component setup of message passing.
- [x] The purpose of `CharacterControllerPass` is to stuff contextual information into the `self` parameter and rely on regular arguments in helper functions for information relevant to the function. However, I'm not that familiar with this design pattern and am not sure if it's an antipattern. I have a more complete example at [collision-demo](https://github.com/patowen/hypermine/blob/2d2f38b1cdb51a79c9ce6480e682183b74b072ed/client/src/sim.rs#L617-L885).
- [x] The `no_clip` toggle option's current state is used as part of the input struct rather than toggling `no_clip` being a kind of input. I believe this likely results in cleaner logic, but it's a bit unintuitive.

### DIFF
--- a/client/src/graphics/window.rs
+++ b/client/src/graphics/window.rs
@@ -230,6 +230,9 @@ impl Window {
                         VirtualKeyCode::F => {
                             down = state == ElementState::Pressed;
                         }
+                        VirtualKeyCode::V if state == ElementState::Pressed => {
+                            self.sim.toggle_no_clip();
+                        }
                         VirtualKeyCode::Escape => {
                             let _ = self.window.set_cursor_grab(CursorGrabMode::None);
                             self.window.set_cursor_visible(true);
--- a/client/src/graphics/window.rs
+++ b/client/src/graphics/window.rs
@@ -132,11 +132,12 @@ impl Window {
                         * (right as u8 as f32 - left as u8 as f32)
                         + na::Vector3::y() * (up as u8 as f32 - down as u8 as f32)
                         + na::Vector3::z() * (back as u8 as f32 - forward as u8 as f32);
-                    self.sim.velocity(if move_direction.norm_squared() > 1.0 {
-                        move_direction.normalize()
-                    } else {
-                        move_direction
-                    });
+                    self.sim
+                        .set_movement_input(if move_direction.norm_squared() > 1.0 {
+                            move_direction.normalize()
+                        } else {
+                            move_direction
+                        });
 
                     self.sim.rotate(&na::UnitQuaternion::from_axis_angle(
                         &-na::Vector3::z_axis(),
--- a/client/src/prediction.rs
+++ b/client/src/prediction.rs
@@ -96,6 +96,7 @@ mod tests {
         let mock_graph = DualGraph::new();
         let mock_character_input = CharacterInput {
             movement: na::Vector3::x(),
+            no_clip: true,
         };
 
         let mut pred = PredictedMotion::new(pos());
--- a/client/src/prediction.rs
+++ b/client/src/prediction.rs
@@ -18,6 +18,7 @@ pub struct PredictedMotion {
     log: VecDeque<CharacterInput>,
     generation: u16,
     predicted_position: Position,
+    predicted_velocity: na::Vector3<f32>,
 }
 
 impl PredictedMotion {
@@ -26,6 +27,7 @@ impl PredictedMotion {
             log: VecDeque::new(),
             generation: 0,
             predicted_position: initial_position,
+            predicted_velocity: na::Vector3::zeros(),
         }
     }
 
@@ -36,6 +38,7 @@ impl PredictedMotion {
             cfg,
             graph,
             &mut self.predicted_position,
+            &mut self.predicted_velocity,
             input,
             cfg.step_interval.as_secs_f32(),
         );
@@ -51,6 +54,7 @@ impl PredictedMotion {
         graph: &DualGraph,
         generation: u16,
         position: Position,
+        velocity: na::Vector3<f32>,
     ) {
         let first_gen = self.generation.wrapping_sub(self.log.len() as u16);
         let obsolete = usize::from(generation.wrapping_sub(first_gen));
@@ -60,12 +64,14 @@ impl PredictedMotion {
         }
         self.log.drain(..obsolete);
         self.predicted_position = position;
+        self.predicted_velocity = velocity;
 
         for input in self.log.iter() {
             character_controller::run_character_step(
                 cfg,
                 graph,
                 &mut self.predicted_position,
+                &mut self.predicted_velocity,
                 input,
                 cfg.step_interval.as_secs_f32(),
             );
@@ -75,6 +81,10 @@ impl PredictedMotion {
     /// Latest estimate of the server's state after receiving all `push`ed inputs.
     pub fn predicted_position(&self) -> &Position {
         &self.predicted_position
+    }
+
+    pub fn predicted_velocity(&self) -> &na::Vector3<f32> {
+        &self.predicted_velocity
     }
 }
 
@@ -105,7 +115,13 @@ mod tests {
         let push =
             |pred: &mut PredictedMotion| pred.push(&mock_cfg, &mock_graph, &mock_character_input);
         let reconcile = |pred: &mut PredictedMotion, generation| {
-            pred.reconcile(&mock_cfg, &mock_graph, generation, pos())
+            pred.reconcile(
+                &mock_cfg,
+                &mock_graph,
+                generation,
+                pos(),
+                na::Vector3::zeros(),
+            )
         };
 
         pred.generation = u16::max_value() - 1;
--- a/client/src/prediction.rs
+++ b/client/src/prediction.rs
@@ -1,6 +1,11 @@
 use std::collections::VecDeque;
 
-use common::{math, proto::Position};
+use common::{
+    character_controller,
+    node::DualGraph,
+    proto::{CharacterInput, Position},
+    SimConfig,
+};
 
 /// Predicts the result of motion inputs in-flight to the server
 ///
@@ -10,7 +15,7 @@ use common::{math, proto::Position};
 /// determine which inputs have been integrated into the server's state and no longer need to be
 /// predicted.
 pub struct PredictedMotion {
-    log: VecDeque<Input>,
+    log: VecDeque<CharacterInput>,
     generation: u16,
     predicted_position: Position,
 }
@@ -26,16 +31,27 @@ impl PredictedMotion {
 
     /// Update for input about to be sent to the server, returning the generation it should be
     /// tagged with
-    pub fn push(&mut self, movement: &na::Vector3<f32>) -> u16 {
-        let transform = math::translate_along(movement);
-        self.predicted_position.local *= transform;
-        self.log.push_back(Input { transform });
+    pub fn push(&mut self, cfg: &SimConfig, graph: &DualGraph, input: &CharacterInput) -> u16 {
+        character_controller::run_character_step(
+            cfg,
+            graph,
+            &mut self.predicted_position,
+            input,
+            cfg.step_interval.as_secs_f32(),
+        );
+        self.log.push_back(input.clone());
         self.generation = self.generation.wrapping_add(1);
         self.generation
     }
 
     /// Update with the latest state received from the server and the generation it was based on
-    pub fn reconcile(&mut self, generation: u16, position: Position) {
+    pub fn reconcile(
+        &mut self,
+        cfg: &SimConfig,
+        graph: &DualGraph,
+        generation: u16,
+        position: Position,
+    ) {
         let first_gen = self.generation.wrapping_sub(self.log.len() as u16);
         let obsolete = usize::from(generation.wrapping_sub(first_gen));
         if obsolete > self.log.len() || obsolete == 0 {
@@ -43,21 +59,23 @@ impl PredictedMotion {
             return;
         }
         self.log.drain(..obsolete);
-        self.predicted_position.node = position.node;
-        self.predicted_position.local = self
-            .log
-            .iter()
-            .fold(position.local, |acc, x| acc * x.transform);
+        self.predicted_position = position;
+
+        for input in self.log.iter() {
+            character_controller::run_character_step(
+                cfg,
+                graph,
+                &mut self.predicted_position,
+                input,
+                cfg.step_interval.as_secs_f32(),
+            );
+        }
     }
 
     /// Latest estimate of the server's state after receiving all `push`ed inputs.
     pub fn predicted_position(&self) -> &Position {
         &self.predicted_position
     }
-}
-
-struct Input {
-    transform: na::Matrix4<f32>,
 }
 
 #[cfg(test)]
@@ -74,17 +92,32 @@ mod tests {
 
     #[test]
     fn wraparound() {
+        let mock_cfg = SimConfig::from_raw(&common::SimConfigRaw::default());
+        let mock_graph = DualGraph::new();
+        let mock_character_input = CharacterInput {
+            movement: na::Vector3::x(),
+        };
+
         let mut pred = PredictedMotion::new(pos());
+
+        // Helper functions to make test more readable
+        let push =
+            |pred: &mut PredictedMotion| pred.push(&mock_cfg, &mock_graph, &mock_character_input);
+        let reconcile = |pred: &mut PredictedMotion, generation| {
+            pred.reconcile(&mock_cfg, &mock_graph, generation, pos())
+        };
+
         pred.generation = u16::max_value() - 1;
-        assert_eq!(pred.push(&na::Vector3::x()), u16::max_value());
-        assert_eq!(pred.push(&na::Vector3::x()), 0);
+
+        assert_eq!(push(&mut pred), u16::max_value());
+        assert_eq!(push(&mut pred), 0);
         assert_eq!(pred.log.len(), 2);
 
-        pred.reconcile(u16::max_value() - 1, pos());
+        reconcile(&mut pred, u16::max_value() - 1);
         assert_eq!(pred.log.len(), 2);
-        pred.reconcile(u16::max_value(), pos());
+        reconcile(&mut pred, u16::max_value());
         assert_eq!(pred.log.len(), 1);
-        pred.reconcile(0, pos());
+        reconcile(&mut pred, 0);
         assert_eq!(pred.log.len(), 0);
     }
 }
--- a/client/src/prediction.rs
+++ b/client/src/prediction.rs
@@ -12,23 +12,23 @@ use common::{math, proto::Position};
 pub struct PredictedMotion {
     log: VecDeque<Input>,
     generation: u16,
-    predicted: Position,
+    predicted_position: Position,
 }
 
 impl PredictedMotion {
-    pub fn new(initial: Position) -> Self {
+    pub fn new(initial_position: Position) -> Self {
         Self {
             log: VecDeque::new(),
             generation: 0,
-            predicted: initial,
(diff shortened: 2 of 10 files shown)


client/src/graphics/window.rs
```diff
@@ -229,6 +229,11 @@ impl Window {
                         VirtualKeyCode::F => {
                             down = state == ElementState::Pressed;
                         }
+                        VirtualKeyCode::V => {
+                            if state == ElementState::Pressed {
+                                self.sim.toggle_no_clip();
+                            }
+                        }
```
Ralith, line 236: nit: fold the `if` into a match guard (`VirtualKeyCode::V if state == ElementState::Pressed`)

client/src/sim.rs
```diff
@@ -75,6 +80,10 @@ impl Sim {
         self.instantaneous_velocity = v;
     }
 
+    pub fn toggle_no_clip(&mut self) {
+        self.toggle_no_clip = true;
```
Ralith, line 87: Why not toggle `no_clip` directly here, as the name suggests?

client/src/sim.rs
```diff
@@ -162,14 +182,31 @@ impl Sim {
...
+                                    input,
+                                    graph,
+                                    config: &params.sim_config,
+                                    dt_seconds: 1.0 / params.sim_config.rate as f32,
+                                }
+                                .step()
+                            },
+                        );
+                    }
+                }
```
Ralith, line 226: Why was this inlined out of `update_position`?

client/src/sim.rs
```diff
@@ -243,33 +280,50 @@ impl Sim {
...
-        let generation = self.prediction.push(
-            &(velocity
-                * params.movement_speed
-                * self.params.as_ref().unwrap().step_interval.as_secs_f32()),
-        );
+        let player_input = CharacterInput {
+            movement: velocity,
+            no_clip: self.no_clip,
+        };
+        let graph = &self.graph;
```
Ralith, line 309: Is this borrow splitting needed with modern Rust?

common/src/proto.rs
```diff
@@ -54,9 +48,15 @@ pub struct Spawns {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Command {
     pub generation: u16,
+    pub player_input: CharacterInput,
```
Ralith, line 52: `character_input` for consistency?

client/src/prediction.rs
```diff
@@ -13,21 +13,31 @@ pub struct PredictedMotion<T: Clone> {
     log: VecDeque<T>,
     generation: u16,
     predicted_position: Position,
+    predicted_velocity: na::Vector3<f32>,
 }
 
 impl<T: Clone> PredictedMotion<T> {
-    pub fn new(initial_position: Position) -> Self {
+    pub fn new(initial_position: Position, initial_velocity: na::Vector3<f32>) -> Self {
```
Ralith, line 20: Are we ever going to want to pass anything other than the zero vector for `initial_velocity`? If you can't immediately think of a case, just bake that in for simplicity for now; easy to change later if needed.

client/src/sim.rs
```diff
@@ -163,12 +166,29 @@ impl Sim {
...
+                        my_velocity = Some(velocity)
+                    }
+                    match self.entity_ids.get(&id) {
+                        None => debug!(%id, "character orientation update for unknown entity"),
+                        Some(&entity) => match self.world.get::<&mut Character>(entity) {
+                            Ok(mut ch) => {
+                                ch.velocity = velocity;
+                            }
+                            Err(e) => {
+                                error!(%id, "character orientation update for non-character entity {}", e)
```
Ralith, line 187: `e` will display as something like "missing component" or "no such entity", so this wording is pretty strange.

common/src/proto.rs
```diff
@@ -34,6 +34,7 @@ pub struct StateDelta {
     /// Highest input generation received prior to `step`
     pub latest_input: u16,
     pub positions: Vec<(EntityId, Position)>,
+    pub character_velocities: Vec<(EntityId, na::Vector3<f32>)>,
     pub character_orientations: Vec<(EntityId, na::UnitQuaternion<f32>)>,
```
Ralith, line 38: Since these two fields always cover exactly the same entities, it'd be cleaner/lighter/more convenient to fold them into a single `Vec<(EntityId, CharacterState)>`.

common/src/proto.rs
```diff
@@ -54,9 +48,15 @@ pub struct Spawns {
...
+    pub player_input: CharacterInput,
     pub orientation: na::UnitQuaternion<f32>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CharacterInput {
     /// Relative to the character's current position, excluding orientation
-    pub velocity: na::Vector3<f32>,
+    pub movement: na::Vector3<f32>,
+    pub no_clip: bool,
```
Ralith, line 60: It's strange for this to be added with no logic that actually depends on it in the same commit.

Ralith:
>  I have the server pass a SimConfig object to the client, which reduces some code repetition, but it also likely adds unwanted coupling between config file organization and the actual simulation configuration.

This is fine, and matches my design in other projects. There's already `SimConfigRaw` specifically to decouple the details of the config file from the in-memory form, and we can add other intermediate forms in the future if needed.

>  The prediction logic and server communication doesn't seem to mesh well with the entity-component abstraction, and it seems like fields are being passed around ad-hoc.

Not sure what you're getting at here; they seem to relate to the ECS in the same way any logic would?

> A lambda is passed into PredictedMotion, and a type parameter is used, both in order to allow mocking for the wraparound() test. However, the lambda setup seems quite ad-hoc and fragile, and it's not obvious that the type T refers to a tick of input.

This decision puzzled me as well. I'd leave `PredictedMotion` and its methods monomorphic for simplicity, and maybe see if you can bundle up the state that it cares about into a single long-lived structure inside `Sim` that can be passed as one mutable reference.

>  The handle_net function has a large amount of code repetition

This will be mitigated if you refactor the message as my review recommends.

> The purpose of CharacterControllerPass is to stuff contextual information into the self parameter and rely on regular arguments in helper functions for information relevant to the function

See discussion of `PredictedMotion` above. In general this pattern isn't terrible, but it's a clue that there may be a better way to organize your code.

> The no_clip toggle option's current state is used as part of the input struct rather than toggling no_clip being a kind of input.

This is correct. How input is physically expressed by the user should be kept decoupled from what the input represents. See also why we have a motion vector instead of `w: bool, a: bool, s: bool, d: bool` or whatever.

Ralith (changes requested):

Ralith:
General note: I can give much better code review if you make changes in the form of small, independent, self-contained commits. Reviewing a commit that changes a bunch of unrelated things for reasons that will only become clear in the context of code in another commit, or code that hasn't even been pushed yet, is difficult.

client/src/sim.rs
patowen: If `no_clip` is toggled immediately until waiting after the next input was sent, there would be a "retcon" for the frames after the last tick, which would cause a jolt if no-clip is toggled while moving. The simplest fix for anything like this is to delay the toggling until the next tick. The main downside is input lag, but this doesn't matter for non-canon things like no-clip toggling.

It might be easier to read this if there's a field called something like `queued_no_clip` or `no_clip_next_tick`, so the setting rather than the changes to be made to a setting are queued up.

client/src/sim.rs
patowen: The purpose of the `update_position` method was a bit unclear to me. It had the dual purpose of setting the position component of all relevant entities as well as updating the player entity's predictions. These purposes meshed less well together when velocity was introduced, as velocity affects position but doesn't live in the position component. I inlined it out to separate out this second purpose with the idea that it probably belongs in a different method.

I never moved it to a different method because my concern about the ad-hoc nature of passing component data around meant that it would add to the complexity of this PR, and I wasn't certain I was going the right direction.

I made a mistake when splitting my changes into two commits, since this change makes less sense outside this context.

client/src/sim.rs
patowen: If I unsplit this, I get the following error:
```
error[E0502]: cannot borrow `self.prediction` as mutable because it is also borrowed as immutable
   --> client\src\sim.rs:287:26
    |
287 |           let generation = self.prediction.push(&player_input, |position, input| {
    |                            ^               ----                ----------------- immutable borrow occurs here
    |                            |               |
    |  __________________________|               immutable borrow later used by call
    | |
288 | |             CharacterControllerPass {
289 | |                 position,
290 | |                 input,
291 | |                 graph: &self.graph,
    | |                         ---------- first borrow occurs due to use of `self` in closure
...   |
295 | |             .step()
296 | |         });
    | |__________^ mutable borrow occurs here
```

I'm not yet sure why this is the case, but my guess is that while modern rust has gotten good about splitting fields, `self` might still be borrowed as a single unit in the implicit closure struct.

client/src/sim.rs
patowen: This was copied from the existing `Err(e) => error!(%id, "position update for unpositioned entity {}", e),` in `update_position(...)`. I'll update both.

common/src/proto.rs
patowen: That makes sense. Would it also make sense to use `Character` instead of a new struct `CharacterState`, and have the "name" live in its own component? This seems more inline with ECS, and it would allow `Character` to be used for something like an NPC.

common/src/proto.rs
patowen: I was unsure whether to put these changes in the first or second commit, so thanks for this feedback!

patowen:
> General note: I can give much better code review if you make changes in the form of small, independent, self-contained commits. Reviewing a commit that changes a bunch of unrelated things for reasons that will only become clear in the context of code in another commit, or code that hasn't even been pushed yet, is difficult. This does usually mean having to rearrange your changes after you first draft them, especially when you're still developing the habit, but the result is much more likely to work as intended.

Good to know. Believe it or not, this is actually the revised version. My first draft lives in [collision-wip](https://github.com/Ralith/hypermine/compare/master...patowen:hypermine:collision-wip), which has smaller commits, but their far less organized. The first commit felt like none of its changes made sense without the context of the other changes in that commit, and it wasn't clear how I could split it further, or whether it made sense to try. I'll keep trying to build up the skill of crafting a sensible commit history.

Nevertheless, your review was exactly what I was looking for, as I should be able to find my footing again. Thanks for the detailed review!

patowen:
> > The prediction logic and server communication doesn't seem to mesh well with the entity-component abstraction, and it seems like fields are being passed around ad-hoc.
> 
> Not sure what you're getting at here; they seem to relate to the ECS in the same way any logic would?

One possibility I was picturing was that predicted information could live in a component rather than being pure internal system data. This could potentially be useful if multiple entities are involved in a single prediction. Also, in engines like Unity, entities have a "transform" component which directly affects draw calls, but in this PR, the draw calls depend on `PredictedMotion` with additional extrapolation on top of that.

Your response suggests that this is nothing to worry about, and I have no disagreements.

Everything else you said in https://github.com/Ralith/hypermine/pull/247#issuecomment-1356476698 makes sense to me.

client/src/sim.rs
Ralith: That makes sense. Please always justify non-obvious behavior like this in comments at minimum.

client/src/sim.rs
Ralith: What version of rustc are you using?

patowen: I'm using 1.66.0. I'm certain my IDE is also using it because it detected the new clippy lints. (I didn't notice the clippy lints before creating this PR because I was still on 1.65 then, but I can still reproduce this error.)

client/src/sim.rs
patowen: Will do. Sometimes, it's a bit unclear to me what's obvious and what isn't, but hopefully, I can get better at this with experience.

common/src/proto.rs
Ralith: The existing `Character` struct should remain, and should have a `CharacterState` member.

> This seems more inline with ECS

The network data model will never and should never exactly match the in-memory data model. Different components of the system have different concerns. For example, the network layer should only send updates for data that is subject to change, but static and dynamic data that are always associated with each other should not be needlessly separated in the ECS.

Ralith:
> predicted information could live in a component rather than being pure internal system data

This isn't an unreasonable thought, but when there's always exactly one predicted character, shoving the data into the ECS just adds unnecessary ceremony. If we add other forms of prediction which are less narrow (e.g. player-fired projectiles) then that state should definitely live in components.

> entities have a "transform" component which directly affects draw calls, but in this PR, the draw calls depend on PredictedMotion with additional extrapolation on top of that.

We should move towards a normalized transform component eventually, both for prediction of the local character and for interpolation of server-governed states. That way downstream logic has a homogeneous data model to work with. However, since there isn't really any downstream logic other than the renderer, the simpler existing pattern isn't doing any harm; one step at a time.

client/src/sim.rs
Ralith: Weird. Well, no matter.

patowen: Oh, I looked further into this. If we switch to the 2021 edition of Rust, I believe the error goes away. Currently, we're using the 2018 edition. I assume that means that this smarter capture logic is not backwards compatible.

Ralith: Ahah, nice catch! Edition updates are usually pretty painless, but no rush.

client/src/sim.rs
```diff
@@ -261,16 +260,16 @@ impl Sim {
...
     pub fn view(&self) -> Position {
-        let mut result = *self.prediction.predicted();
+        let mut result = *self.prediction.predicted_position();
         result.local *= self.orientation.to_homogeneous();
         if let Some(ref params) = self.params {
             // Apply input that hasn't been sent yet
-            let velocity = sanitize_motion_input(self.average_velocity);
+            let movement = sanitize_motion_input(self.average_movement_input);
             // We multiply by the entire timestep rather than the time so far because
             // self.average_velocity is always over the entire timestep, filling in zeroes for the
```
Ralith: Comment needs updating here.

client/src/sim.rs
```diff
@@ -75,6 +80,13 @@ impl Sim {
         self.movement_input = movement_input;
     }
 
+    /// Prepares no_clip to be toggled after the next step. We avoid toggling it immediately, as
+    /// that would cause a discontinuity when predicting the player's position within a given step,
+    /// causing an undesirable jolt.
```
Ralith, line 85: nit: doc-comments should only contain information relevant to the caller, whereas information on the (reasoning behind the) internals should be relegated to regular `//` comments, and perhaps placed lexically inside the function.

common/src/character_controller.rs
```diff
@@ -35,11 +38,29 @@ impl<T> CharacterControllerPass<'_, T> {
...
+                *self.velocity += current_to_target_velocity.normalize() * max_delta_velocity;
+            } else {
+                *self.velocity += current_to_target_velocity;
+            }
+
+            // Update position using the average between the old and new velocity to avoid the following two issues:
+            // 1. Input lag, which would occur if only the old velocity was used
+            // 2. Discontinuities, which would occur if only the new velocity was used
+            self.position.local *=
+                math::translate_along(&((*self.velocity + old_velocity) * 0.5 * self.dt_seconds));
```
Ralith, line 63: I found this difficult to follow. Is the idea that, by averaging the velocities together, we're actually modeling a velocity that increases linearly over the timestep? That makes sense, but isn't clearly communicated by the comment. In particular, I don't see how using the new velocity directly would be strictly more discontinuous, since it might model a temporary spike in velocity by the same token that this models a linear ramp to the end velocity.

common/src/proto.rs
```diff
@@ -81,4 +82,5 @@ pub struct FreshNode {
 pub struct Character {
     pub name: String,
     pub orientation: na::UnitQuaternion<f32>,
+    pub velocity: na::Vector3<f32>,
```
Ralith, line 85: These fields should be replaced with a single `CharacterState` field, since otherwise we're just manually inlining it.

server/src/sim.rs
```diff
@@ -204,9 +212,3 @@ fn dump_entity(world: &hecs::World, entity: Entity) -> Vec<Component> {
     }
     components
 }
-
-struct Character {
```
Ralith, line 197: Since this is being removed in favor of using `proto::Character`, which we're importing and referencing without qualification, we should take care to replace existing qualified references to `proto::Character` (e.g. in `dump_entity`) with plain `Character` to avoid giving the impression of multiple types. Further, `dump_entity` could now be `Clone`ing the component directly rather than enumerating its fields.

Ralith (changes requested):
Thank you, this was a pleasure to review! Exactly what I had in mind.

common/src/character_controller.rs
patowen: Your wording of treating velocity as increasing linearly over the timestep makes more sense than my wording, and I believe that is the general idea.

As for discontinuities, I made a mistake, as I had encountered what felt like discontinuities in the past, but that's unrelated to the code here and was due to a now-fixed bug. The main issue I encounter now with only using the new velocity is that when you let go of the "forward" button, you go backwards a little bit at the end. I assume you also accelerate more quickly at the beginning, but that's harder to notice.

I'll update the comment.

client/src/sim.rs
patowen: Since the only doc comments that are relevant to the caller are that the function toggles no_clip, I opted to remove them entirely.

patowen:
I have applied all your suggestions, so this should be ready for another review. The diff of my latest changes are in https://github.com/Ralith/hypermine/commit/6902a955788111bfceae1edf2658e5e2e503a007. This lives in my collision-wip branch, which I've used to (temporarily) keep track of the true history of changes I've made over time. While the branch as a whole is confusing and not really useful for reviewing, this single commit representing the minor changes I most recently made may be useful.

patowen:
@Ralith: My plan for future PR revisions is to push the changes to GitHub, leave a summary comment of the changes, and re-request the review (without making a comment "mentioning" you like this one). I believe this strikes a good balance in ensuring you see my changes without being discourteous.

However, I'm not 100% sure that GitHub notifications are reliable enough for this to work. To avoid accidentally being in a situation where both of us are waiting for each other, please can you confirm that leaving a comment and re-requesting the review (the things I did three hours ago) are sufficient?

Ralith:
Yep, that's a fine workflow and should reliably notify me in ways I'll typically notice within a day or two. Also feel free to bump a PR or reach out via side-channels if I haven't responded to something within a week.

Ralith (approved):

PR #249, opened 2022-12-18
Upgrade to Rust 2021 edition

patowen:
As Hypermine is designed to work with the latest version of Rust, it also makes sense to upgrade to the latest edition as well. There are some borrow checker improvements between these two editions.

I tested this manually, and it still works as before. No changes are needed for 2021 edition compatibility.


Ralith (approved):
Thanks!

PR #250, opened 2022-12-18
Fix ineffective renormalize_isometry function

patowen:
This fixes a bug I introduced in https://github.com/Ralith/hypermine/pull/244.

`math::translate` was changed to require Lorentz-normalized vectors, but `renormalize_isometry` did not fulfill this new requirement. This bug wasn't caught it any of the tests, and since it takes a while to manifest itself, I didn't notice until now. I've added a new test that should hopefully prevent this issue from reoccurring after a later refactor.


Ralith (approved):
Nice catch!

PR #251, opened 2023-01-03
Allow the server to generate terrain instead of just the client

patowen:
Relying on the consistency of terrain-generation code between runs and between different computers, we have the server generate chunks in a smaller radius of nodes around the player. This, in addition to future changes, will allow the server to run code that depends on the current state of the world, such as collision checking, terrain modification, NPC spawns, etc.

As no code currently depends on this server-generated terrain, testing that these changes work as intended is not possible without additional code modification. However, even if this is broken, this should not result in any regressions, so to limit the scope of this PR, the lack of I believe the lack of testability is acceptable.

One minor concrete change this PR makes is that terrain generation for the client no longer depends on which direction the player is looking. Surface extraction is unaffected and is still affected as before.

### DIFF
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -7,15 +7,16 @@ use rand::{Rng, SeedableRng};
 use tracing::{error_span, info, trace};
 
 use common::{
-    character_controller,
+    character_controller, dodeca,
     graph::{Graph, NodeId},
     math,
-    node::DualGraph,
+    node::{populate_fresh_nodes, Chunk, DualGraph},
     proto::{
         Character, CharacterInput, CharacterState, ClientHello, Command, Component, FreshNode,
         Position, Spawns, StateDelta,
     },
-    traversal::ensure_nearby,
+    traversal::{ensure_nearby, nearby_nodes},
+    worldgen::ChunkParams,
     EntityId, SimConfig, Step,
 };
 
@@ -157,7 +158,42 @@ impl Sim {
                 })
                 .collect(),
         };
-        self.graph.clear_fresh();
+        populate_fresh_nodes(&mut self.graph);
+
+        // We want to load all chunks that a player can interact with in a single step, so chunk_generation_distance
+        // is set up to cover that distance.
+        // TODO: Use actual max speed instead of max ground speed.
+        // TODO: Account for the radius of the player's collision sphere
+        let chunk_generation_distance = dodeca::BOUNDING_SPHERE_RADIUS
+            + self.cfg.max_ground_speed as f64 * self.cfg.step_interval.as_secs_f64()
+            + 0.001;
+
+        // Load all chunks around entities corresponding to clients, which correspond to entities
+        // with a "Character" component.
+        for (_, (position, _)) in self.world.query::<(&Position, &Character)>().iter() {
+            let nodes = nearby_nodes(&self.graph, position, chunk_generation_distance);
+            for &(node, _) in &nodes {
+                for chunk in dodeca::Vertex::iter() {
+                    if let Chunk::Fresh = self
+                        .graph
+                        .get(node)
+                        .as_ref()
+                        .expect("all nodes must be populated before loading their chunks")
+                        .chunks[chunk]
+                    {
+                        if let Some(params) =
+                            ChunkParams::new(self.cfg.chunk_size, &self.graph, node, chunk)
+                        {
+                            self.graph.get_mut(node).as_mut().unwrap().chunks[chunk] =
+                                Chunk::Populated {
+                                    voxels: params.generate_voxels(),
+                                    surface: None,
+                                };
+                        }
+                    }
+                }
+            }
+        }
 
         // TODO: Omit unchanged (e.g. freshly spawned) entities (dirty flag?)
         let delta = StateDelta {
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -10,6 +10,7 @@ use common::{
     character_controller,
     graph::{Graph, NodeId},
     math,
+    node::DualGraph,
     proto::{
         Character, CharacterInput, CharacterState, ClientHello, Command, Component, FreshNode,
         Position, Spawns, StateDelta,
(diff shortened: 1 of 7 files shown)


common/Cargo.toml
```diff
@@ -16,6 +16,7 @@ anyhow = "1.0.26"
 quinn = "0.8.3"
 lazy_static = "1.4.0"
 fxhash = "0.2.1"
+tokio = "1.18.2"
```
Ralith, line 19: We should specify the features needed by `common` (probably just `rt`) and disable default features.

common/src/chunk_loader.rs
```diff
@@ -39,8 +39,32 @@ impl ChunkLoader {
...
+                if let Chunk::Fresh = graph
+                    .get(node)
+                    .as_ref()
+                    .expect("all nodes must be populated before loading their chunks")
+                    .chunks[chunk]
+                {
+                    if let Some(params) = ChunkParams::new(dimension, graph, node, chunk) {
+                        if self.load(node, params) {
+                            graph.get_mut(node).as_mut().unwrap().chunks[chunk] = Chunk::Generating;
+                        }
```
Ralith, line 59: Pre-existing issue (and hence best fixed in a separate commit, and not necessarily in this PR), but we should immediately `return` if there's no capacity available.

server/src/sim.rs
```diff
@@ -156,7 +164,27 @@ impl Sim {
...
+            let mut nodes =
+                nearby_nodes(&self.graph, position, self.cfg.simulation_distance as f64);
+            // Sort nodes by distance to the view to prioritize loading closer data and improve early Z
+            // performance
+            let view_pos = position.local * math::origin();
+            nodes.sort_unstable_by(|&(_, ref xf_a), &(_, ref xf_b)| {
+                math::mip(&view_pos, &(xf_a * math::origin()))
+                    .partial_cmp(&math::mip(&view_pos, &(xf_b * math::origin())))
+                    .unwrap_or(std::cmp::Ordering::Less)
+            });
```
Ralith, line 181: The server doesn't care about the order in which chunks are loaded, nor does it need to load chunks that do not overlap with players. We should only load the bare minimum without sorting, and expand in the future if necessary.

Ralith (changes requested):

common/Cargo.toml
patowen: It looks like `tokio` doesn't have any default features. I believe the necessary features are being added because "common" isn't being built on its own. (I was unaware of https://doc.rust-lang.org/nightly/cargo/reference/features.html#feature-unification).

If I build "common" on its own, I get ``error[E0432]: unresolved import `tokio::sync::mpsc` ``. I can get rid of this error by enabling the "sync" feature, but the "rt" feature is already enabled because of the "quinn" crate, making it a bit difficult to notice that it wasn't enabled when importing the tokio crate.

I'll go ahead and add "sync" and "rt".

server/src/sim.rs
patowen: Since the idea is for the server to only load the bare necessary chunks, the idea of loading chunks synchronously may be worth entertaining. However, we would miss out on the advantage of multiple threads.

However, I don't think only loading chunks that overlap with players is sufficient. We want sufficient room for collision checking to be able to evaluate if a player can move to a specific spot, which means that we at least need to cover where the player is and where it can move in one frame. Multiple frames would allow for wiggle room if something temporarily slows down chunk generation.

Ralith: > It may be as simple as removing the simulation_distance configuration option and setting it to the max node radius plus some buffer to account for player radius and movement speed.

That's what I had in mind. The distance should be the player radius plus the distance a player could traverse in one timestep at maximum speed, maybe plus a small margin (say, 1cm) to account for rounding error.

> Multiple frames would allow for wiggle room if something temporarily slows down chunk generation.

A larger radius means an exponential increase in load, so I'd prefer keep things minimal.

> the idea of loading chunks synchronously may be worth entertaining. However, we would miss out on the advantage of multiple threads.

For a one-step movement bound to work, we need to finish loading chunks in the same step that we deem them necessary, before running character controllers for that step. This is appealing as it guarantees consistent behavior, with the server running slow under heavy load rather than behaving strangely. We'd also want to remove the limit on in-flight chunks. Parallelism can still be employed, but we would need to block the timestep until all outstanding work is complete. At that point it might not make sense to share the same loader between client and server, since their requirements are significantly different.

Alternatively, we could stick with the loose async approach, treat unloaded chunks as solid obstacles, and allow for two steps of motion rather than one, so that under good conditions the next step has all the chunks it needs in the worst case. This might make for better use experience under bad conditions, with the server still running in real time but occasional glitches when entering new territory. On the other hand, those glitches might feel much worse than a server running slowly, and the larger margin means higher load, and therefore odds of becoming CPU-bound, to begin with. I think the approach in the previous paragraph is therefore preferable.

patowen: I agree that it makes sense to go with the synchronous approach, which will require me to make significant changes to this PR. For now, I think it makes sense to keep it single-threaded, since the exponential growth you mentioned goes both ways, so the client generates significantly more chunks than the server, making this unlikely to be a bottleneck.

To support falling (and potentially flying), any solution we have should be able to accommodate a relatively large maximum speed. IRL terminal velocity is about 60 m/s, which is 6 meters per time step (for the current defaults), which I don't expect to be large enough to cause a bottleneck.

Ralith: SGTM. Note that bottleneck risk mainly arises in the face of large numbers of clients, but a strictly serial approach should indeed be fine for a good long time.

patowen:
All requested changes should now be done. Due to your suggestions, the list of changes is much smaller than before, so it's been reduced to three commits. It's possible that chunk generation code should live somewhere outside the `step` method, but until the logic gets more complicated, any abstraction I try to add feels like a guess, so I believe that just putting the code in the `step` method is fine for now, but I'm open to suggestions.

server/src/sim.rs
```diff
@@ -157,7 +158,42 @@ impl Sim {
                 })
                 .collect(),
         };
-        self.graph.clear_fresh();
+        populate_fresh_nodes(&mut self.graph);
+
+        // We want to load all chunks that a player can interact with in a single frame, so chunk_generation_distance
```
Ralith, line 163: nit: step, not frame

Ralith (changes requested):

Ralith (approved):

PR #252, opened 2023-01-03
Use libm to help prevent future terrain generation consistency bugs

patowen:
To prevent requiring the server to send lots of data to the client, terrain generation should be run on both the server and the client separately. To avoid inconsistencies in terrain generation caused by differing implementations of transcendental functions, we use the `libm` library, which provides its own implementation of these functions.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -19,7 +19,8 @@ ash-window = "0.12.0"
 raw-window-handle = "0.5.0"
 directories = "4.0.1"
 vk-shader-macros = "0.2.5"
-nalgebra = "0.31.2"
+nalgebra = { version = "0.31.2", features = ["libm-force"] }
+libm = "0.2.6"
 tokio = { version = "1.18.2", features = ["rt-multi-thread", "sync", "macros"] }
 png = "0.17.5"
 anyhow = "1.0.26"
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -10,11 +10,12 @@ license = "Apache-2.0 OR Zlib"
 
 [dependencies]
 serde = { version = "1.0.104", features = ["derive"] }
-nalgebra = { version = "0.31.2", features = ["rand", "serde-serialize"] }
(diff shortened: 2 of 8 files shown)


Ralith (changes requested):
You need to actually replace calls to trancedental functions from std with calls into libm.

patowen:
Oh yeah. That makes sense in retrospect.

patowen:
@Ralith: Actually, this seems potentially dangerous. As far as I can tell, there's no way to un-import the std math functions (without something drastic like no_implicit_prelude). Mixing std math and libm math in the codebase seems troublesome to get correct. Do you have a suggestion for how to get libm used reliably?

Ralith:
LD_PRELOAD something which stubs out the libc versions with ones that call `abort`? Good discipline will be necessary. It's not the end of the world if there's an occasional mistake.

patowen:
That may work. One other potential concern is that libm's performance may be significantly worse than std's. Since we only need libm for terrain generation, and other things would be self-correcting, we could be leaving performance on the table.

That being said, I'm hopeful that this won't be the bottleneck.

patowen:
> LD_PRELOAD something which stubs out the libc versions with ones that call `abort`? Good discipline will be necessary. It's not the end of the world if there's an occasional mistake.

Given that it's not the end of the world if there's an occasional mistake, and only worldgen needs consistency, I'm inclined to just not enforce this and instead rely on code reviews to catch human error.

Ralith:
> One other potential concern is that libm's performance may be significantly worse than std's

This is unlikely; host libc isn't magic.

patowen:
I believe the main performance benefit of std would be potential hardware implementations of common functions like sqrt. However, I haven't found anything noteworthy during my testing, so I agree that this is nothing to worry about.

I'm noticing that far fewer changes are needed than I originally thought. Anything using `RealField` is affected by nalgebra's `libm-force` feature, so all our generic code uses `libm` without any modifications.

Ralith (approved):

PR #256, opened 2023-01-28
Fix clippy warnings for Rust 1.67.0

patowen:
~~This PR fixes 5 instances of [needless_borrowed_reference](https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrowed_reference) and 6 instances of [uninlined_format_args](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args).~~

This PR allows the [needless_borrowed_reference](https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrowed_reference) lint and fixes 6 instances of [uninlined_format_args](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args).

### DIFF
--- a/client/src/graphics/window.rs
+++ b/client/src/graphics/window.rs
@@ -282,7 +282,7 @@ impl Window {
                         self.swapchain_needs_update = true;
                     }
                     Err(e) => {
-                        panic!("acquire_next_image: {}", e);
+                        panic!("acquire_next_image: {e}");
                     }
                 }
             };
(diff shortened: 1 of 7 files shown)


Ralith:
Let's allowlist the overzelaous (https://github.com/rust-lang/rust-clippy/issues/10245) `needless_borrowed_reference` lint rather than churning perfectly good code.

patowen:
Will do. I'm personally fine with the lint, as every code pattern is hard to read until you're used to it, and it taught me an ergonomic feature I didn't know about before. It would also keep this code in sync with unrelated Rust projects (a major advantage of opinionated linters). However, I agree there's a good chance this is just worse, a fault with clippy, that we should work around, so I will ignore it.

What are your thoughts on the other lint, uninlined_format_args? I'm not a fan of the inline format args feature because it's too limited, only allowing variable names. Because of that, common changes/refactors that were trivial, such as combining local variable assignment and use, require switching between inlined and uninlined format args, and the resulting code can look needlessly inconsistent.

Ralith:
For the common case where you are just formatting a variable name, I think the inlined version is a lot easier to keep track of. It's not perfect but I don't mind it.

Issue #127, opened 2020-08-15
Walking


suhr:
Looking around is already cool, but even more cool would being able to *walk* on this terrain.

Depends on https://github.com/Ralith/hypermine/issues/34 I guess.

dranorter:
The issue with walking is gravity. Hyperblock recently decided to use a very Escher-like gravity: https://www.youtube.com/watch?v=NzK0MCwgLE8 but other solutions would be possible; nothing is perfectly intuitive.

Ralith:
To the contrary, the issue is collision detection. It's easy enough to come up with a consistent rule for gravity, but that's no use if you just fall forever.

Issue #34, opened 2020-04-06
Voxel vs. player collision detection


Ralith:

Ralith:
Implemented in #262.

PR #262, opened 2023-02-11
Add initial sphere-voxel collision detection 

patowen:
This PR adds collision detection logic. There is a `shape_cast` method in the `common::collision` module that answers the question of "How far can a sphere of a given radius move in a given direction, up to a given distance, before colliding with a voxel? Does it collide at all?" If there is a collision, it also finds the normal, which can be used for collision resolution.

For testing purposes, a basic form of collision resolution is also included with these changes. As long as noclip is disabled by pressing "V", if a character collides with a wall, it will stop. This has the advantage of simplicity, but it has the disadvantage of making walls "sticky", so smoother movement is planned in a future PR.

### DIFF
--- a/client/src/graphics/voxels/mod.rs
+++ b/client/src/graphics/voxels/mod.rs
@@ -21,7 +21,7 @@ use common::{
     graph::NodeId,
     lru_slab::SlotId,
     math,
-    node::{Chunk, VoxelData},
+    node::{Chunk, ChunkId, VoxelData},
     traversal::nearby_nodes,
     LruSlab,
 };
@@ -146,14 +146,13 @@ impl Voxels {
             }
 
             use Chunk::*;
-            for chunk in Vertex::iter() {
+            for vertex in Vertex::iter() {
+                let chunk = ChunkId::new(node, vertex);
                 // Fetch existing chunk, or extract surface of new chunk
                 match sim
                     .graph
-                    .get_mut(node)
-                    .as_mut()
+                    .get_chunk_mut(chunk)
                     .expect("all nodes must be populated before rendering")
-                    .chunks[chunk]
                 {
                     Generating => continue,
                     Fresh => {
@@ -161,12 +160,10 @@ impl Voxels {
                         if let Some(params) = common::worldgen::ChunkParams::new(
                             self.surfaces.dimension() as u8,
                             &sim.graph,
-                            node,
                             chunk,
                         ) {
                             if self.worldgen.load(ChunkDesc { node, params }).is_ok() {
-                                sim.graph.get_mut(node).as_mut().unwrap().chunks[chunk] =
-                                    Generating;
+                                sim.graph[chunk] = Generating;
                             }
                         }
                         continue;
@@ -181,7 +178,7 @@ impl Voxels {
                             frame.drawn.push(slot);
                             // Transfer transform
                             frame.surface.transforms_mut()[slot.0 as usize] =
-                                node_transform * chunk.chunk_to_node().map(|x| x as f32);
+                                node_transform * vertex.chunk_to_node().map(|x| x as f32);
                         }
                         (&mut ref mut surface @ None, &VoxelData::Dense(ref data)) => {
                             // Extract a surface so it can be drawn in future frames
@@ -203,7 +200,7 @@ impl Voxels {
                             frame.extracted.push(scratch_slot);
                             let slot = self.states.insert(SurfaceState {
                                 node,
-                                chunk,
+                                chunk: vertex,
                                 refcount: 0,
                             });
                             *surface = Some(slot);
@@ -224,7 +221,7 @@ impl Voxels {
                                 indirect_offset: self.surfaces.indirect_offset(slot.0),
                                 face_offset: self.surfaces.face_offset(slot.0),
                                 draw_id: slot.0,
-                                reverse_winding: chunk.parity() ^ node_is_odd,
+                                reverse_winding: vertex.parity() ^ node_is_odd,
                             });
                         }
                         (None, &VoxelData::Solid(_)) => continue,
--- a/common/src/character_controller.rs
+++ b/common/src/character_controller.rs
@@ -2,7 +2,7 @@ use tracing::info;
 
 use crate::{
     graph_collision, math,
-    node::DualGraph,
+    node::{ChunkLayout, DualGraph},
     proto::{CharacterInput, Position},
     sanitize_motion_input, SimConfig,
 };
@@ -107,7 +107,14 @@ impl CharacterControllerPass<'_> {
         let ray = graph_collision::Ray::new(math::origin(), displacement_normalized);
         let tanh_distance = displacement_norm.tanh();
 
-        let cast_hit = graph_collision::sphere_cast(self.position, &ray, tanh_distance);
+        let cast_hit = graph_collision::sphere_cast(
+            self.cfg.character_radius,
+            self.graph,
+            &ChunkLayout::new(self.cfg.chunk_size as usize),
+            self.position,
+            &ray,
+            tanh_distance,
+        );
 
         let allowed_displacement = math::translate(
             &ray.position,
--- a/common/src/character_controller.rs
+++ b/common/src/character_controller.rs
@@ -1,4 +1,4 @@
-use tracing::info;
+use tracing::{error, info};
 
 use crate::{
     graph_collision, math,
@@ -115,6 +115,14 @@ impl CharacterControllerPass<'_> {
             &ray,
             tanh_distance,
         );
+
+        let cast_hit = match cast_hit {
+            Ok(r) => r,
+            Err(e) => {
+                error!("Collision checking returned {:?}", e);
+                return CollisionCheckingResult::stationary();
+            }
+        };
 
         let allowed_displacement = math::translate(
             &ray.position,
--- a/common/src/character_controller.rs
+++ b/common/src/character_controller.rs
@@ -1,5 +1,7 @@
+use tracing::info;
+
 use crate::{
-    math,
+    graph_collision, math,
     node::DualGraph,
     proto::{CharacterInput, Position},
     sanitize_motion_input, SimConfig,
@@ -56,15 +58,25 @@ impl CharacterControllerPass<'_> {
                 *self.velocity += current_to_target_velocity;
             }
 
-            // Update position by using the average of the old velocity and new velocity, which has
-            // the effect of modeling a velocity that changes linearly over the timestep. This is
-            // necessary to avoid the following two issues:
+            // Set expected displacement by using the average of the old velocity and new velocity,
+            // which has the effect of modeling a velocity that changes linearly over the timestep.
+            // This is necessary to avoid the following two issues:
             // 1. Input lag, which would occur if only the old velocity was used
             // 2. Movement artifacts, which would occur if only the new velocity was used. One
-            //    example of such an artifact is the player moving backwards slightly when they
+            //    example of such an artifact is the character moving backwards slightly when they
             //    stop moving after releasing a direction key.
-            self.position.local *=
-                math::translate_along(&((*self.velocity + old_velocity) * 0.5 * self.dt_seconds));
+            let expected_displacement = (*self.velocity + old_velocity) * 0.5 * self.dt_seconds;
+
+            // Update position with collision checking
+            let collision_checking_result = self.check_collision(&expected_displacement);
+            self.position.local *= collision_checking_result.allowed_displacement;
+            if let Some(collision) = collision_checking_result.collision {
+                *self.velocity = na::Vector3::zeros();
+                // We are not using collision normals yet, so print them to the console to allow
+                // sanity checking. Note that the "orientation" quaternion is not used here, so the
+                // numbers will only make sense if the character doesn't look around.
+                info!("Collision: normal = {:?}", collision.normal);
+            }
         }
 
         // Renormalize
@@ -75,6 +87,75 @@ impl CharacterControllerPass<'_> {
         if next_node != self.position.node {
             self.position.node = next_node;
             self.position.local = transition_xf * self.position.local;
+        }
+    }
+
+    /// Checks for collisions when a character moves with a character-relative displacement vector of `relative_displacement`.
+    fn check_collision(&self, relative_displacement: &na::Vector3<f32>) -> CollisionCheckingResult {
+        // Split relative_displacement into its norm and a unit vector
+        let relative_displacement = relative_displacement.to_homogeneous();
+        let displacement_sqr = relative_displacement.norm_squared();
+        if displacement_sqr < 1e-16 {
+            // Fallback for if the displacement vector isn't large enough to reliably be normalized.
+            // Any value that is sufficiently large compared to f32::MIN_POSITIVE should work as the cutoff.
+            return CollisionCheckingResult::stationary();
+        }
+
+        let displacement_norm = displacement_sqr.sqrt();
+        let displacement_normalized = relative_displacement / displacement_norm;
+
+        let ray = graph_collision::Ray::new(math::origin(), displacement_normalized);
+        let tanh_distance = displacement_norm.tanh();
+
+        let cast_hit = graph_collision::sphere_cast(self.position, &ray, tanh_distance);
+
+        let allowed_displacement = math::translate(
+            &ray.position,
+            &math::lorentz_normalize(
+                &ray.ray_point(
+                    cast_hit
+                        .as_ref()
+                        .map_or(tanh_distance, |hit| hit.tanh_distance),
+                ),
+            ),
+        );
+
+        CollisionCheckingResult {
+            allowed_displacement,
+            collision: cast_hit.map(|hit| Collision {
+                // `CastEndpoint` has its `normal` given relative to the character's original position,
+                // but we want the normal relative to the character after the character moves to meet the wall.
+                // This normal now represents a contact point at the origin, so we omit the w-coordinate
+                // to ensure that it's orthogonal to the origin.
+                normal: na::UnitVector3::new_normalize(
+                    (math::mtranspose(&allowed_displacement) * hit.normal).xyz(),
+                ),
+            }),
+        }
+    }
+}
+
+struct CollisionCheckingResult {
+    /// Multiplying the character's position by this matrix will move the character as far as it can up to its intended
+    /// displacement until it hits the wall.
+    allowed_displacement: na::Matrix4<f32>,
+    collision: Option<Collision>,
+}
+
+struct Collision {
+    /// This collision normal faces away from the collision surface and is given in the perspective of the character
+    /// _after_ it is transformed by `allowed_displacement`. The 4th coordinate of this normal vector is assumed to be
+    /// 0.0 and is therefore omitted.
+    normal: na::UnitVector3<f32>,
+}
+
+impl CollisionCheckingResult {
+    /// Return a CollisionCheckingResult with no movement and no collision; useful if the character is not moving
+    /// and has nothing to check collision against. Also useful as a last resort fallback if an unexpected error occurs.
+    fn stationary() -> CollisionCheckingResult {
+        CollisionCheckingResult {
+            allowed_displacement: na::Matrix4::identity(),
+            collision: None,
         }
     }
 }
--- a/common/src/graph_collision.rs
+++ b/common/src/graph_collision.rs
@@ -1,0 +1,108 @@
+use crate::{math, proto::Position};
+
+/// Performs ray casting against a fixed plane.
+///
+/// This is a temporary implementation, as it is designed to perfrom sphere casting (swept collision query) against
+/// the voxels in a `DualGraph` when it is ready for use.
+///
+/// The `ray` parameter and any resulting hit normals are given in the local coordinate system of `position`.
+///
+/// The `tanh_distance` is the hyperbolic tangent of the cast_distance, or the distance along the ray to check for hits.
+pub fn sphere_cast(position: &Position, ray: &Ray, tanh_distance: f32) -> Option<GraphCastHit> {
+    let node_ray = position.local * ray;
+
+    // As a temporary placeholder implementation, check for collision against the z=0 plane going in the -z direction.
+
+    const EPSILON: f32 = 1e-4;
+    let normal = math::mtranspose(&position.local) * na::Vector4::z();
+    if node_ray.direction.z >= 0.0 {
+        // Going the wrong way. Don't report a collision.
+        None
+    } else if (-EPSILON..=0.0).contains(&node_ray.position.z) {
+        // Extra logic to ensure precision issues don't allow a collider to clip through a surface
+        Some(GraphCastHit {
+            tanh_distance: 0.0,
+            normal,
+        })
+    } else {
+        // Compute needed tanh_distance to reach z=0 plane.
+        let new_tanh_distance = -node_ray.position.z / node_ray.direction.z;
+        if new_tanh_distance >= 0.0 && new_tanh_distance < tanh_distance {
+            // Report a collision if the needed tanh_distance is in the right range
+            Some(GraphCastHit {
+                tanh_distance: new_tanh_distance,
+                normal,
+            })
+        } else {
+            None
+        }
+    }
+}
+
+/// Information about the intersection at the end of a ray segment.
+#[derive(Debug)]
+pub struct GraphCastHit {
+    /// The tanh of the distance traveled along the ray to result in this hit.
+    pub tanh_distance: f32,
+
+    /// Represents the normal vector of the hit surface in the original coordinate system
+    /// of the sphere casting. To get the actual normal vector, project it so that it is orthogonal
+    /// to the endpoint in Lorentz space.
+    pub normal: na::Vector4<f32>,
+}
+
+/// A ray in hyperbolic space. The fields must be lorentz normalized, with `mip(position, position) == -1`,
+/// `mip(direction, direction) == 1`, and `mip(position, direction) == 0`.
+#[derive(Debug)]
+pub struct Ray {
+    pub position: na::Vector4<f32>,
+    pub direction: na::Vector4<f32>,
+}
+
+impl Ray {
+    pub fn new(position: na::Vector4<f32>, direction: na::Vector4<f32>) -> Ray {
+        Ray {
+            position,
+            direction,
+        }
+    }
+
+    /// Returns a point along this ray `atanh(tanh_distance)` units away from the origin. This point
+    /// is _not_ lorentz normalized.
+    pub fn ray_point(&self, tanh_distance: f32) -> na::Vector4<f32> {
+        self.position + self.direction * tanh_distance
+    }
+}
+
+impl std::ops::Mul<&Ray> for na::Matrix4<f32> {
+    type Output = Ray;
+
+    #[inline]
+    fn mul(self, rhs: &Ray) -> Self::Output {
+        Ray {
+            position: self * rhs.position,
+            direction: self * rhs.direction,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use approx::assert_abs_diff_eq;
+
+    use super::*;
+
+    #[test]
+    fn temp_shape_cast_example() {
+        // Directly hit the z=0 plane with a ray 0.5 units away.
+        let ray = math::translate_along(&na::Vector3::new(0.0, 0.0, 0.5))
+            * &Ray::new(math::origin(), -na::Vector4::z());
+        assert_abs_diff_eq!(
+            sphere_cast(&Position::origin(), &ray, 0.9)
+                .unwrap()
+                .tanh_distance,
+            0.5_f32.tanh(),
+            epsilon = 1e-4
+        );
+    }
+}
--- a/common/src/graph_collision.rs
+++ b/common/src/graph_collision.rs
@@ -1,42 +1,50 @@
-use crate::{math, proto::Position};
+use crate::{
+    chunk_collision::chunk_sphere_cast,
+    dodeca::Vertex,
+    math,
+    node::{Chunk, ChunkId, ChunkLayout, DualGraph},
+    proto::Position,
+};
 
-/// Performs ray casting against a fixed plane.
-///
-/// This is a temporary implementation, as it is designed to perfrom sphere casting (swept collision query) against
-/// the voxels in a `DualGraph` when it is ready for use.
+/// Performs sphere casting (swept collision query) against the voxels in the `DualGraph`
 ///
 /// The `ray` parameter and any resulting hit normals are given in the local coordinate system of `position`.
 ///
 /// The `tanh_distance` is the hyperbolic tangent of the cast_distance, or the distance along the ray to check for hits.
-pub fn sphere_cast(position: &Position, ray: &Ray, tanh_distance: f32) -> Option<GraphCastHit> {
-    let node_ray = position.local * ray;
+///
+/// This implementaion is incomplete, and it will only detect intersection with voxels in the "A" chunk of the position's node.
+pub fn sphere_cast(
+    collider_radius: f32,
+    graph: &DualGraph,
+    layout: &ChunkLayout,
+    position: &Position,
+    ray: &Ray,
+    tanh_distance: f32,
+) -> Option<GraphCastHit> {
+    let chunk = ChunkId::new(position.node, Vertex::A);
+    let node_transform = position.local;
+    let Chunk::Populated {
+            voxels: ref voxel_data,
+            ..
+        } = graph[chunk] else {
+            // Collision checking on unpopulated chunk
+            panic!("Collision checking on unpopulated chunk");
+        };
+    let local_ray = chunk.vertex.node_to_dual().cast::<f32>() * node_transform * ray;
 
-    // As a temporary placeholder implementation, check for collision against the z=0 plane going in the -z direction.
-
-    const EPSILON: f32 = 1e-4;
-    let normal = math::mtranspose(&position.local) * na::Vector4::z();
-    if node_ray.direction.z >= 0.0 {
-        // Going the wrong way. Don't report a collision.
-        None
-    } else if (-EPSILON..=0.0).contains(&node_ray.position.z) {
-        // Extra logic to ensure precision issues don't allow a collider to clip through a surface
-        Some(GraphCastHit {
-            tanh_distance: 0.0,
-            normal,
-        })
-    } else {
-        // Compute needed tanh_distance to reach z=0 plane.
-        let new_tanh_distance = -node_ray.position.z / node_ray.direction.z;
-        if new_tanh_distance >= 0.0 && new_tanh_distance < tanh_distance {
-            // Report a collision if the needed tanh_distance is in the right range
-            Some(GraphCastHit {
-                tanh_distance: new_tanh_distance,
-                normal,
-            })
-        } else {
-            None
-        }
-    }
+    // Check collision within a single chunk
+    chunk_sphere_cast(
+        collider_radius,
+        voxel_data,
+        layout,
+        &local_ray,
+        tanh_distance,
+    )
+    .map(|hit| GraphCastHit {
+        tanh_distance: hit.tanh_distance,
+        chunk,
+        normal: math::mtranspose(&node_transform) * chunk.vertex.dual_to_node().cast() * hit.normal,
+    })
 }
 
 /// Information about the intersection at the end of a ray segment.
@@ -44,6 +52,9 @@ pub fn sphere_cast(position: &Position, ray: &Ray, tanh_distance: f32) -> Option
 pub struct GraphCastHit {
     /// The tanh of the distance traveled along the ray to result in this hit.
     pub tanh_distance: f32,
+
+    /// Which chunk in the graph the hit occurred in
+    pub chunk: ChunkId,
 
     /// Represents the normal vector of the hit surface in the original coordinate system
     /// of the sphere casting. To get the actual normal vector, project it so that it is orthogonal
@@ -83,26 +94,5 @@ impl std::ops::Mul<&Ray> for na::Matrix4<f32> {
             position: self * rhs.position,
             direction: self * rhs.direction,
         }
-    }
-}
-
-#[cfg(test)]
-mod tests {
-    use approx::assert_abs_diff_eq;
-
-    use super::*;
-
-    #[test]
-    fn temp_shape_cast_example() {
-        // Directly hit the z=0 plane with a ray 0.5 units away.
-        let ray = math::translate_along(&na::Vector3::new(0.0, 0.0, 0.5))
-            * &Ray::new(math::origin(), -na::Vector4::z());
-        assert_abs_diff_eq!(
-            sphere_cast(&Position::origin(), &ray, 0.9)
-                .unwrap()
-                .tanh_distance,
-            0.5_f32.tanh(),
-            epsilon = 1e-4
-        );
     }
 }
--- a/common/src/graph_collision.rs
+++ b/common/src/graph_collision.rs
@@ -1,6 +1,10 @@
+use std::collections::VecDeque;
+
+use fxhash::FxHashSet;
+
 use crate::{
     chunk_collision::chunk_sphere_cast,
-    dodeca::Vertex,
+    dodeca::{self, Vertex},
     math,
     node::{Chunk, ChunkId, ChunkLayout, DualGraph},
     proto::Position,
@@ -12,7 +16,9 @@ use crate::{
 ///
 /// The `tanh_distance` is the hyperbolic tangent of the cast_distance, or the distance along the ray to check for hits.
 ///
-/// This implementaion is incomplete, and it will only detect intersection with voxels in the "A" chunk of the position's node.
+/// This function may return a `SphereCastError` if not enough chunks are generated, even if the ray never reaches an
+/// ungenerated chunk. To prevent these errors, make sure that the distance between the ray's start point and the center of
+/// the closest node with ungenerated chunks is greater than `cast_distance + collider_radius + dodeca::BOUNDING_SPHERE_RADIUS`
 pub fn sphere_cast(
     collider_radius: f32,
     graph: &DualGraph,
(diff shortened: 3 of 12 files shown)


common/src/ray_tracing.rs
```diff
@@ -0,0 +1,105 @@
+use crate::math;
+
+/// Performs ray tracing against the z=0 plane in node coordinates. This plane is one-sided, so the ray must be going in the -z
+/// direction to detect an intersection.
+///
+/// The `start_node_transform` parameter determines which coordinate system the `ray` parameter and any resulting intersection
+/// normals are given in. Specifically, the `start_node_transform` matrix converts this coordinate system to node coordinates.
+///
+/// The `tanh_distance` is the hyperbolic tangent of the distance along the ray to check for intersections.
+pub fn trace_ray(
```
Ralith, line 19: This function *demands* a decent suite of unit tests.

common/src/character_controller.rs
```diff
@@ -77,4 +89,66 @@ impl CharacterControllerPass<'_> {
...
+    /// Multiplying the character's position by this matrix will move the character as far as it can up to its intended
+    /// displacement until it hits the wall.
+    allowed_displacement: na::Matrix4<f32>,
+    collision: Option<Collision>,
+}
+
+struct Collision {
+    /// This collision normal faces away from the collision surface and is given in the perspective of the character
+    /// _after_ it is transformed by `allowed_displacement`.
+    normal: na::Vector3<f32>,
```
Ralith, line 162: Why is this a `Vector3` when `RayTracingIntersection`'s is a `Vector4`?

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,294 @@
+use crate::{
```
Ralith, line 1: Approximately everything in this file needs unit tests.

common/src/ray_tracing.rs
```diff
@@ -103,3 +178,82 @@ impl std::ops::Mul<&Ray> for na::Matrix4<f32> {
         }
     }
 }
+
+/// Represents a discretized region in the voxel grid contained by an axis-aligned bounding box.
+pub struct CubicVoxelRegion {
```
Ralith, line 258: `VoxelAABB`?

common/src/ray_tracing.rs
```diff
@@ -19,39 +32,81 @@ pub fn trace_ray(
...
+
+    /// Returns the radius of the sphere of influence of the collider. `trace_ray` might not be called on
+    /// chunks outside this sphere of influence.
+    fn max_radius(&self) -> f32;
+}
+
+/// Contains all the immutable data needed for `ChunkRayTracer` to perform its logic
+pub struct ChunkRayTracingContext<'a> {
+    pub dimension: usize,
+    pub dimension_f32: f32,
```
Ralith, line 155: Storing these separately is not necessary. Seems like mostly code wants the `f32`, so let's stick with that.

Ralith (changes requested):
Initial incomplete pass; more to come over coming days, but this should get things rolling.

common/src/ray_tracing.rs
```diff
@@ -1,13 +1,26 @@
...
+    math,
+    node::{Chunk, ChunkId, DualGraph, VoxelData},
+    world::Material,
+};
 
-/// Performs ray tracing against the z=0 plane in node coordinates. This plane is one-sided, so the ray must be going in the -z
-/// direction to detect an intersection.
+/// Performs ray tracing against the voxels in the `DualGraph`. This function is suitable for collision checking with
+/// any different collider shape. Each collider would have a different implemenation of `ChunkRayTracer`, which describes
+/// how the collision checking math works within a single chunk.
```
Ralith, line 12: Do we have immediate plans for any other shapes to be tested against voxels? If not, a trait seems very premature.

common/src/ray_tracing.rs
```diff
@@ -1,13 +1,26 @@
...
+    dodeca::Vertex,
+    math,
+    node::{Chunk, ChunkId, DualGraph, VoxelData},
+    world::Material,
+};
 
-/// Performs ray tracing against the z=0 plane in node coordinates. This plane is one-sided, so the ray must be going in the -z
-/// direction to detect an intersection.
+/// Performs ray tracing against the voxels in the `DualGraph`. This function is suitable for collision checking with
+/// any different collider shape. Each collider would have a different implemenation of `ChunkRayTracer`, which describes
```
Ralith, line 11: Is this a ray trace or a shape cast (aka swept collision query)? The name suggests the former, but this comment, and the presence of collision shape information other than the voxels, suggests the latter.

Ralith (changes requested):

common/src/ray_tracing.rs
```diff
@@ -0,0 +1,105 @@
...
+        result.intersection = Some(RayTracingIntersection {
+            normal: math::mtranspose(&start_node_transform) * na::Vector4::z(),
+        });
+    }
+
+    result
+}
+
+/// Contains all information that was discovered as a result of ray tracing.
+pub struct RayTracingResult {
```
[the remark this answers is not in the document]
patowen: I thought of `RaySegment` but decided against it because it implied that it contained both endpoints. I like the name you suggested.

common/src/ray_tracing.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// A ray in hyperbolic space. The fields must be lorentz normalized, with `mip(position, position) == -1`,
+/// `mip(direction, direction) == 1`, and `mip(position, direction) == 0`.
+pub struct Ray {
+    pub position: na::Vector4<f32>,
+    pub direction: na::Vector4<f32>,
+}
+
+impl Ray {
+    pub fn new(position: na::Vector4<f32>, direction: na::Vector4<f32>) -> Ray {
```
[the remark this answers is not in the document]
patowen: Possibly, although I don't think we `debug_assert` most normalization invariants in the rest of the codebase.

common/src/ray_tracing.rs
```diff
@@ -0,0 +1,105 @@
...
+        Ray {
+            position,
+            direction,
+        }
+    }
+
+    /// Returns a point along this ray `atanh(tanh_distance)` units away from the origin. This point
+    /// is _not_ lorentz normalized.
```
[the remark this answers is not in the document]
patowen: Maybe? One thing in that vein that I had considered is that we could have a type for a "vector with Lorentzian semantics", which would wrap an `na::Vector4<N>` and have member functions for operations that make sense in hyperbolic geometry, like `mip`. Then, the "Unit" version of this type would be Lorentz-normalized. There could even be separate types for point-like unit vectors (mip with itself is `-1`) and direction-like vectors (mip with itself is `+1`).

common/src/ray_tracing.rs
```diff
@@ -0,0 +1,105 @@
+use crate::math;
+
+/// Performs ray tracing against the z=0 plane in node coordinates. This plane is one-sided, so the ray must be going in the -z
+/// direction to detect an intersection.
```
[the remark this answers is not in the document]
patowen: It would be the -z direction based on the Hyperboloid model. This would also be the direction represented by the `(0, 0, -1, 0)` vector, which is a valid direction vector, as it's tangent to the hyperboloid whenever we're on the z=0 plane.

common/src/ray_tracing.rs
```diff
@@ -0,0 +1,105 @@
...
+///
+/// The `start_node_transform` parameter determines which coordinate system the `ray` parameter and any resulting intersection
+/// normals are given in. Specifically, the `start_node_transform` matrix converts this coordinate system to node coordinates.
```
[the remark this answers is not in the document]
patowen: If we separated these concerns, I don't think it would save that much work. The reason why `ray_tracing` handles these transformations might not immediately be clear until the graph traversal commit, when there are multiple `node_transform`s.

common/src/ray_tracing.rs
patowen: The logic that definitely needs unit tests in this function gets moved to `sphere_collider.rs` later on. My plan for now is to create a unit test in this commit and move/adapt it in the future commit. It's possible we'll still want this method to be tested directly in the final result, but I'll hold off on that for now.

common/src/character_controller.rs
patowen: The normal vector here is in relation to the character, and the implication is that the vector represents a direction starting at the character's center. Because of that, the `w` coordinate is always 0, so we enforce that invariant by making it a `Vector3` instead of a `Vector4`. Converting a `Vector4` to this `Vector3` effectively performs the necessary projection to ensure this invariant holds.

common/src/ray_tracing.rs
patowen: We will want to test points for the ray-casting used to allow block placing and breaking.

common/src/ray_tracing.rs
patowen: This is a shape cast, assuming the collider used isn't a point collider. I am not familiar with the terms shape-casting and "swept collision query", so ray tracing was the best term I could come up with.

I've been visualizing it as ray tracing against an abstract shape where a constant-width outline is added around all the voxels.

common/src/ray_tracing.rs
patowen: If we want to keep just one of them, I think it makes more sense to keep `dimension`, since it is used in some places, and converting a usize to a float and back to a usize without any other changes seems not ideal.

This context class isn't permanently stored anywhere, so there doesn't seem to be much of a downside to keeping both versions of `dimension`.

That being said, I'm also happy to just use `dimension as f32` everywhere.

common/src/ray_tracing.rs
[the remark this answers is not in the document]
Ralith: Might not be a bad time to start.

common/src/ray_tracing.rs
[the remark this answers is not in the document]
Ralith: I like the sound of that. Something for hyperbolic isometries, too. Later, though.

common/src/ray_tracing.rs
Ralith: Playtesting is an extremely inefficient way to test. We should cover at least basic sanity checks, and ideally interesting corner cases as well.

common/src/ray_tracing.rs
Ralith: Sure, sounds good.

> there doesn't seem to be much of a downside to keeping both versions of dimension.

The downside is increased verbosity, decreased readability, and room for error.

common/src/ray_tracing.rs
Ralith: Let's call it "shape cast" then.

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,294 @@
...
+            let ray_endpoint = ctx.ray.ray_point(tanh_distance);
+            result.update(ctx, tanh_distance, ray_endpoint - vertex_position);
+        }
+    }
+}
+
+/// Finds the tanh of the distance a sphere will have to travel along a ray before it
+/// intersects the given point.
+///
+/// Returns NaN if there's no such intersection
```
Ralith, line 176: It's very easy to handle possibly-`NaN` data incorrectly. Should this and below to return `Option<f32>` instead?

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,294 @@
...
+    // We use an alternative quadratic formula to ensure that we return a positive number if `constant_term > 0.0`.
+    // Otherwise, the edge case of a small positive `constant_term` could be mishandled.
+    // Note that discriminant can be negative, which allows this function to return NaN when there is no solution.
+    constant_term / (-half_linear_term + discriminant.sqrt())
+}
+
+/// Converts from t-u-v coordinates to x-y-z coordinates. t-u-v coordinates are a permuted version of x-y-z coordinates.
+/// `t_axis` determines which of the three x-y-z coordinates corresponds to the t-coordinate. This function works with
+/// any indexable entity with at least three entries. Any entry after the third entry is ignored.
+fn tuv_to_xyz<T: std::ops::IndexMut<usize, Output = N>, N: Copy>(t_axis: usize, tuv: T) -> T {
```
Ralith, line 265: `fn tuv_to_xyz<T: Copy, const D: usize>(t_axis: usize, tuv: [T; D])`, then calling `.into()` on vectors where needed, might be a little more straightforward. Your call, though.

common/src/ray_tracing.rs
```diff
@@ -103,3 +178,82 @@ impl std::ops::Mul<&Ray> for na::Matrix4<f32> {
...
+
+            // We convert to usize here instead of earlier because out-of-range voxel coordinates can be negative.
+            bounds[axis] = [voxel_min.floor() as usize, voxel_max.floor() as usize];
+        }
+
+        Some(CubicVoxelRegion { bounds })
+    }
+
+    /// Creates an iterator over grid points contained in the region, represented as ordered triples
+    pub fn grid_point_iterator(
```
Ralith, line 306: nit: `grid_points` is more concise and no less clear; similarly below.

common/src/ray_tracing.rs
patowen: I'm happy to create a few basic sanity checks. However, the fact that we want test suites on functions that act on a populated `DualGraph` suggests that a future enhancement outside the scope of this PR would be the ability to select a "test world" for worldgen, and this can be the same world that is used for these kinds of tests.

For now, I can create a basic test world with only two or three solid blocks, although this will still be a fair amount of setup code.

common/src/ray_tracing.rs
patowen: The use case already exists in the collision-demo branch. Any logic involving rays can benefit from the graph traversal algorithm in ray_tracing.rs.

However, I understand that the design can change far more than originally anticipated, so I'm okay with getting rid of this abstraction.

common/src/ray_tracing.rs
Ralith: Specifying geometry to test against is certainly necessary. We should work towards factoring out common code for that as test cases grow.

common/src/ray_tracing.rs
```diff
@@ -24,48 +24,118 @@ pub fn trace_ray(
...
+                    .magnitude()
+                    .acosh();
+                let ray_length = result.tanh_distance.atanh();
+                if ray_node_distance - ray_length
+                    > dodeca::BOUNDING_SPHERE_RADIUS as f32 + chunk_ray_tracer.max_radius()
+                {
+                    // Ray cannot intersect node
+                    continue;
+                }
+                // If we have to do collision checking on nodes that don't exist in the graph, we cannot have a conclusive result.
```
Ralith, line 113: Could we treat missing nodes/unpopulated chunks as solid walls for collision detection purposes?

Ralith (changes requested):

common/src/ray_tracing.rs
Ralith: It sounds like we'll eventually want to factor out the graph traversal algorithm into a more abstract swept broadphase, then. I'd definitely favor leaving the abstraction for that until we actually need it.

common/src/ray_tracing.rs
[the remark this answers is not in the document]
Ralith: Checks out.

common/src/sphere_collider.rs
patowen: Possibly. It would make the code calling it slightly easier to read, and any performance cost should be minor compared to the vector arithmetic required.

common/src/ray_tracing.rs
patowen: It's definitely possible, but it would be a compromise. Server-side, it's always a logic error if we collide with an unpopulated chunk, since we generate synchronously. This may be worth discussing pros and cons on Discord.

common/src/ray_tracing.rs
Ralith: "swept broadphase" is a terse description of the graph traversal algorithm.

common/src/ray_tracing.rs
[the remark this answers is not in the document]
patowen: I'm having second thoughts about implementing this.
* Approximate equality is needed for this debug assertion.
* `approx` is currently a dev-dependency, so it only works in tests.
* `approx` has no `debug_assert_abs_diff_eq!` method, so I would need to use `debug_assert!(abs_diff_eq!(...))` instead, making error messages less useful unless additional boilerplate code is added.
* I currently highly doubt this will be a common source of bugs. Trying to make this more airtight might be better done when we start having separate types for various normalization states as you commented below.

If you would still prefer me to add this, which approach would you recommend. Should I add the dependency, add new boilerplate code somewhere (where would you recommend?), or just do something like the following?
```rust
debug_assert!(
    (math::mip(&position, &position) + 1.0).abs() < 1e-5,
    "mip(position, position) == {}",
    math::mip(&position, &position)
);
```

patowen:
I've updated this PR and added most of the necessary tests. The main shape casting method in the collision module still needs test, as I've saved that for last, and it might take some time. I believe there are enough new changes that it's worth a review (especially reviewing the design of the test). When I find time, I'll add the remaining missing tests.

I believe everything other than `collision::shape_cast` is well-covered.

common/src/ray_tracing.rs
patowen: I managed to add a test case for this method which tests a fair amount of its functionality. To save my own time, my plan is to wait until you've reviewed some of my tests before I add some for corner cases. That way, I can get a better sense of your preferred coding style when it comes to tests for this project.

common/src/ray_tracing.rs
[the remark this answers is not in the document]
Ralith: I had imagined we'd do something like your example, but I agree it's low risk and redundant to more rigorous mechanisms involving the type system.

client/src/graphics/voxels/mod.rs
```diff
@@ -146,27 +146,24 @@ impl Voxels {
...
                         if let Some(params) = common::worldgen::ChunkParams::new(
                             self.surfaces.dimension() as u8,
                             &sim.graph,
-                            node,
                             chunk,
                         ) {
                             if self.worldgen.load(ChunkDesc { node, params }).is_ok() {
-                                sim.graph.get_mut(node).as_mut().unwrap().chunks[chunk] =
-                                    Generating;
+                                *sim.graph.get_chunk_mut(chunk).unwrap() = Generating;
```
Ralith, line 166: nit: Perhaps we should `impl Index(Mut)<ChunkId>` for panicing assets to save writing `unwrap` a bunch.

common/src/collision.rs
```diff
@@ -0,0 +1,129 @@
...
+            normal: math::mtranspose(&start_node_transform) * na::Vector4::z(),
+        });
+    }
+
+    endpoint
+}
+
+/// Where a cast ray ended, and all information about the hit at the end if such a hit occurred.
+#[derive(Debug)]
+pub struct RayEndpoint {
```
Ralith, line 175: s/`Ray`/`Cast`/ for consistency

common/src/collision.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub tanh_distance: f32,
+
+    /// Information about the intersection at the end of the ray segment. If this is `None`, there
+    /// are no intersections.
+    pub hit: Option<RayHit>,
+}
+
+/// Information about the intersection at the end of a ray segment.
+#[derive(Debug)]
+pub struct RayHit {
```
Ralith, line 192: as above

common/src/collision.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn mul(self, rhs: &Ray) -> Self::Output {
+        Ray {
+            position: self * rhs.position,
+            direction: self * rhs.direction,
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
Ralith: This is a good start, but we should always also test *negative* cases. For example, a ray parallel to the plane (though I guess that's less interesting in H3, heh) and a ray pointing away from it.

common/src/collision.rs
```diff
@@ -1,14 +1,25 @@
...
-/// Temporarily performs ray casting instead for testing purposes, as if the sphere has radius 0.
 ///
 /// The `start_node_transform` parameter determines which coordinate system the `ray` parameter and any resulting hit
 /// normals are given in. Specifically, the `start_node_transform` matrix converts this coordinate system to the coordinate
 /// system of `start_chunk`'s node.
 ///
 /// The `tanh_distance` is the hyperbolic tangent of the distance along the ray to check for hits.
+///
+/// This implementaion is incomplete, and it will only detect intersection with voxels in `start_chunk`.
 pub fn shape_cast(
```
Ralith, line 18: nit: Since this is no longer generic, `sphere_cast` would be a bit clearer.

common/src/collision.rs
```diff
@@ -109,21 +178,140 @@ impl std::ops::Mul<&Ray> for na::Matrix4<f32> {
...
+        let aabb = VoxelAABB::from_ray_segment_and_radius(
+            dimension,
+            dual_to_grid_factor,
+            &ray,
+            tanh_distance,
+            radius,
+        )
+        .unwrap();
+
+        let covered_planes: HashSet<_> = aabb.grid_planes(0).collect();
```
Ralith, line 437: `collect`ing this into a `HashSet` feels a bit silly when it starts out as a `Range` which is both smaller and faster to query against. Perhaps returning `impl Iterator` is excessively opaque in this case?

Ralith, line 437: Should we test the other axes too?

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,800 @@
+use crate::{
+    collision::{ChunkShapeCastingContext, Ray, RayEndpoint},
+    math,
+    world::Material,
+};
+
+/// Handles collisions for a single chunk.
+pub fn chunk_shape_cast(ctx: &ChunkShapeCastingContext, endpoint: &mut RayEndpoint) {
```
Ralith, line 8: nit: `sphere_cast`

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,800 @@
...
+
+    let discriminant = half_linear_term * half_linear_term - quadratic_term * constant_term;
+    if discriminant < 0.0 {
+        return None;
+    }
+
+    // We use an alternative quadratic formula to ensure that we return a positive number if `constant_term > 0.0`.
+    // Otherwise, the edge case of a small positive `constant_term` could be mishandled.
+    // Note that discriminant can be negative, which allows this function to return NaN when there is no solution.
+    Some(constant_term / (-half_linear_term + discriminant.sqrt()))
```
Ralith, line 241: Can both `half_linear_term` and `discriminant` be zero here?

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,800 @@
...
+
+            let mut test_ctx = TestShapeCastContext {
+                collider_radius,
+                dimension,
+                dual_to_grid_factor: Vertex::dual_to_chunk_factor() as f32 * dimension as f32,
+                voxel_data: VoxelData::Solid(Material::Void),
+                transform,
+            };
+
+            // Populate voxels. Consists of a single cube with grid coordinates from (1, 1, 1) to (2, 2, 2)
```
Ralith, line 313: What space are the coordinates in this comment? The context suggests voxel coordinates, but only a single voxel is set, not a box of eight.

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,800 @@
...
+        let corrected_normal =
+            math::lorentz_normalize(&(normal + ray.position * math::mip(&normal, &ray.position)));
+
+        // Check that the normal and ray are pointing opposite directions
+        assert!(math::mip(&corrected_normal, &ray.direction) < 0.0);
+    }
+
+    /// Tests that a suitable collision is found when approaching a single voxel from various angles.
+    #[test]
+    fn chunk_shape_cast_examples() {
```
Ralith, line 495: A few negative cases here wouldn't go amiss.

common/src/collision.rs
```diff
@@ -23,54 +23,122 @@ pub fn shape_cast(
...
     // This `endpoint` variable gets updated over time before being returned.
     let mut endpoint = RayEndpoint {
         tanh_distance,
         hit: None,
     };
 
+    // Start a breadth-first search of the graph's chunks, performing collision checks in each relevant chunk.
+    // The `chunk_queue` contains ordered pairs containing the `ChunkId` and the transformation needed to switch
+    // from the original node coordinates to the current chunk's node coordinates.
+    let mut visited_chunks: HashSet<ChunkId> = HashSet::new();
```
Ralith, line 37: nit: `FxHashSet` is a better choice when the input isn't at risk of being carefully crafted to produce a DoS.

Ralith (changes requested):
This is looking great, thanks!

common/src/sphere_collider.rs
```diff
@@ -0,0 +1,800 @@
...
+    }
+
+    let discriminant = half_linear_term * half_linear_term - quadratic_term * constant_term;
+    if discriminant < 0.0 {
+        return None;
+    }
+
+    // We use an alternative quadratic formula to ensure that we return a positive number if `constant_term > 0.0`.
+    // Otherwise, the edge case of a small positive `constant_term` could be mishandled.
+    // Note that discriminant can be negative, which allows this function to return NaN when there is no solution.
```
Ralith, line 240: Outdated comment.

common/src/sphere_collider.rs
Ralith: Also, could `half_linear_term` be equal to `discriminant.sqrt()`?

Ralith (changes requested):

common/src/sphere_collider.rs
patowen: They're in "grid coordinates". the voxel coordinates would just be (2, 2, 2).

common/src/sphere_collider.rs
patowen: To get an infinity from this, `half_linear_term` has to be non-negative. This would only happen if the character is moving away from a wall, and no collision should be detected.

Infinity is outside the bounds of any range we choose, so this edge case shouldn't give us problems.

common/src/collision.rs
patowen: I'm undecided on this. The advantage of `impl Iterator` is that it's consistent with the other `VoxelAABB` functions.

The speed disadvantage is minor, since it's still very fast, and tests need not be optimized unless they're slowing down CI.

If I test with lines or points, I would need a hashset.

patowen: We can. I consider all axes equivalent with their logic, so it's not entirely necessary. However, we should probably test with a few different rays, and we might as well have those tests look at different axes.

common/src/collision.rs
patowen: Oh yeah. I didn't try that hard to add extensive tests here because the code was going to be deleted.

common/src/collision.rs
patowen: I had left it as `Ray` for consistency with the `Ray` struct, since this class can act as a companion to it. Still happy to rename, but I would like your confirmation.

common/src/sphere_collider.rs
patowen: I do have negative cases in separate tests, but I can add one here as well.

common/src/collision.rs
patowen: I originally chose `HashSet` because this is a test. I'll switch it over.

common/src/sphere_collider.rs
Ralith: Is that scheme used anywhere else?

common/src/collision.rs
Ralith: `pub fn shape_cast` is a test?

common/src/collision.rs
Ralith: I think a more generic name makes sense. You can cast rays and spheres and other things, and it's weird to use ray-oriented nomenclature to refer to a sphere-cast result which isn't, in fact, on the ray.

common/src/collision.rs
Ralith: No sense overthinking support code in a test I suppose.

Ralith: It's not obvious to me that it'd be impossible to have bugs that break some, but not all, axes, and it seems easy enough to just slap a for loop around this whole thing.

common/src/sphere_collider.rs
Ralith: I suppose infinity is a relatively sensible/harmless result.

common/src/collision.rs
patowen: Oh, whoops. I got this mixed up with the `VoxelAABB` test.

common/src/sphere_collider.rs
patowen: It was one of the coordinate systems discussed in Discord, and it's used for the iterators in `VoxelAABB`.

common/src/collision.rs
patowen: Good point. Since the ray isn't equal in all coordinates, it would give us more test cases for free.

common/src/collision.rs
patowen: That's fair. I'll use the term `Cast` for everything except the `Ray` struct.

common/src/sphere_collider.rs
patowen: I definitely appreciate the call-out. Anywhere infinities can get involved, you can also end up with large numbers where you didn't intend them. I did need to think about it to be sure that we were fine after reading your question.

I think this is probably worth a code comment.

common/src/sphere_collider.rs
Ralith: Let's revisit this in the future I guess. Probably mostly a doc problem.

common/src/sphere_collider.rs
patowen: Actually, thinking about this again, I think this does need to be addressed. If the numerator is also zero, we get NaN, and we cannot necessarily rule that out. I'll take a closer look at this.

patowen: It looks like the answer is to restrict the solution space to non-negative lower solutions that are not double-roots (We don't care about double-roots because they correspond to moving tangent to a surface, which doesn't have to be a collision). This forces the constant term to be non-negative and the linear term to be negative. Because the linear term is negative, the denominator cannot be zero.

client/src/graphics/voxels/mod.rs
patowen: That seems like a good idea. There is the risk of reduced mindfulness when making `unwrap` implicit. However, this tradeoff has already been considered in Rust standard library, and they decided in favor of allowing this, so I am willing to believe the benefits outweigh the costs.

common/src/ray_tracing.rs
patowen: Based on your review, it doesn't seem like there's anything fundamentally wrong with how I set up these tests, so I'll add some edge case tests here before I request another review.

patowen: I found a bug while testing an edge case, so you were definitely right that this needed testing. When setting `ray_node_distance`, I used the wrong formula for "distance to the origin". For a normalized hypervector, it should either by `.xyz().magnitude().asinh()` or `.w.cosh()`, but I used `.xyz().magnitude().acosh()`.

I tried to test for this manually, but I guess my aim wasn't good enough. Thanks for insisting on unit tests!

common/src/collision.rs
patowen: Actually, looking at this again, the test for planes isn't quite complete, as an implementation that passes this test for all rays could still be buggy. The planar bounds of a sphere's path is not determined just by the start and the end point, as cylinders bulge in the middle. This assumption only works in Euclidean geometry.

Rather than try some difficult-to-verify calculation to see whether a plane is in the sphere's path, I think the right move is to just test the union of a bunch of spheres along this path, since then we just need sphere/plane intersection queries. This would also let me test lines and points without hitting the wall I was hitting earlier.

Ralith: Sounds reasonable to me.

patowen:
Another round of changes has been complete. This time, I have added all the tests that I believe should be added, but I am open to suggestions if you think I missed something important. I also believe all your other suggestions have been addressed.

I made a few other minor changes as well that I noticed while rereading the code. Probably the most important change is that `sphere_cast` now takes a Position `position` instead of a matrix `start_node_transform` and a ChunkId `start_chunk`. There are at least two benefits to this: (1) The caller doesn't have to awkwardly pass in an arbitrary vertex, as that is handled internally, and (2) It's much easier to explain what the `position` parameter means in the documentation.

patowen:
@Ralith, looking at this code again, I think it might make more sense to try to make sphere_collider.rs more stand-alone. It doesn't need a `ChunkId` or a transformation matrix to compute anything internally; that's only used on the result. Is this something worth pursuing in this PR, or would you prefer it as future refactor in a separate PR?

Ralith:
That sounds like a nice improvement. I don't think it needs to block this PR, but if you get it into a state you're happy with before we merge this, then you might as well fold them together.

patowen:
> That sounds like a nice improvement. I don't think it needs to block this PR, but if you get it into a state you're happy with before we merge this, then you might as well fold them together.

Sounds good to me. Once you're reviewed it to your satisfaction and believe it is ready to merge as-is, I think the right move is to merge it, with the spirit of not letting "perfect" be the enemy of "good".

However, if I find time to make this adjustment before we reach that point, I'll go ahead and add it to this PR, adjusting the commit history as if this was the plan from the start. If we go this route, it may be a good idea for me to prepare another diff for you to reference so that it's easier for you to see what changed.

Ralith:
> it may be a good idea for me to prepare another diff for you to reference so that it's easier for you to see what changed.

Nah, no need to go that far.

patowen:
> Nah, no need to go that far.

Got it. I won't go that far, especially if you won't use it. The real diff history will be available in https://github.com/patowen/hypermine/tree/collision-wip, since that's the branch I usually work on, but I won't spend any effort to clean that up.

patowen:
> That sounds like a nice improvement. I don't think it needs to block this PR, but if you get it into a state you're happy with before we merge this, then you might as well fold them together.

@Ralith, I ended up making the changes I proposed. It does make some code more verbose and repetitive (especially `chunk_collision::chunk_sphere_cast`), but I would say it's an overall win for readability.

common/src/ray_tracing.rs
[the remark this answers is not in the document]
patowen: Created https://github.com/Ralith/hypermine/issues/264.

common/src/node.rs
```diff
@@ -93,6 +93,52 @@ impl VoxelData {
...
+    }
+
+    /// Scale by this to convert dual coordinates to homogeneous grid coordinates.
+    pub fn dual_to_grid_factor(&self) -> f32 {
+        self.dual_to_grid_factor
+    }
+
+    /// Converts a single coordinate from dual coordinates in the Klein-Beltrami model to an integer coordinate
+    /// suitable for voxel lookup. Margins are included. Returns `None` if the coordinate is outside the chunk.
+    #[inline]
```
Ralith, line 123: If we're gunna bother to mark things inline, definitely do so for the pure getters above.

common/src/character_controller.rs
```diff
@@ -107,7 +107,14 @@ impl CharacterControllerPass<'_> {
         let ray = graph_collision::Ray::new(math::origin(), displacement_normalized);
         let tanh_distance = displacement_norm.tanh();
 
-        let cast_hit = graph_collision::sphere_cast(self.position, &ray, tanh_distance);
+        let cast_hit = graph_collision::sphere_cast(
+            self.cfg.character_radius,
+            self.graph,
+            &ChunkLayout::new(self.cfg.chunk_size as usize),
```
Ralith, line 113: Should we just store the `ChunkLayout` in the config structure directly?

common/src/graph_collision.rs
```diff
@@ -12,39 +16,121 @@ use crate::{
...
+        )
+        .map_or(hit, |hit| {
+            Some(GraphCastHit {
+                tanh_distance: hit.tanh_distance,
+                chunk,
+                normal: math::mtranspose(&node_transform)
+                    * chunk.vertex.dual_to_node().cast()
+                    * hit.normal,
+            })
+        });
```
Ralith, line 75: Shouldn't this only replace `hit` if the newly detected hit is closer? Consider e.g. a spherecast roughly parallel to and overlapping wit the edge of a chunk. This would be a good test case, too.

common/src/graph_collision.rs
```diff
@@ -12,39 +16,121 @@ use crate::{
...
+        for axis in 0..3 {
+            // Check for neighboring nodes
+            if klein_ray_start[axis] <= klein_lower_boundary
+                || klein_ray_end[axis] <= klein_lower_boundary
+            {
+                let side = chunk.vertex.canonical_sides()[axis];
+                let next_node_transform = side.reflection().cast::<f32>() * node_transform;
+                // Crude check to ensure that the neighboring chunk's node can be in the path of the ray. For simplicity, this
+                // check treats each node as a sphere and assumes the ray is pointed directly towards its center. The check is
+                // needed because chunk generation uses this approximation, and this check is not guaranteed to pass near corners.
```
Ralith, line 94: Does this mean `BOUNDING_SPHERE_RADIUS` is incorrect? It sounds like this comment is describing a false negative case, which seems bad.

common/src/graph_collision.rs
```diff
@@ -12,39 +16,121 @@ use crate::{
...
+                let ray_node_distance = (next_node_transform * ray.position).w.acosh();
+                let ray_length = current_tanh_distance.atanh();
+                if ray_node_distance - ray_length
```
Ralith, line 97: Should we also account for the collider radius here?

Ralith (changes requested):
Good progress; loved the refactor!

common/src/graph_collision.rs
patowen: That is handled by the fact that `chunk_sphere_cast` only detects collisions at `current_tanh_distance` or closer. I agree with adding a test case for this, though.

common/src/graph_collision.rs
patowen: That is handled by `> dodeca::BOUNDING_SPHERE_RADIUS as f32 + collider_radius`. However, I should take another look at which terms are on which side of the inequality sign for minimal confusion.

common/src/graph_collision.rs
Ralith: Oh, that's clever, and obvious in retrospect. Didn't read closely enough I guess!

common/src/graph_collision.rs
patowen: `BOUNDING_SPHERE_RADIUS` is correct as far as I can tell. The issue is that finding the exact set of chunks subject to a sphere cast is needlessly complicated, so we use an AABB-based conservative approximation, but this approximation can include chunks outside the set of generated chunks under specific scenarios. This additional check fine-tunes the set of chunks based on a sphere-based conservative approximation instead of an AABB-based one.

common/src/character_controller.rs
patowen: Possibly. The idea of the `ChunkLayout` struct is that it would eagerly pre-compute things that depend on the dimension, just like `dodeca` pre-computes things that don't depend on the dimension. Using it more widely would have the additional benefit of allowing experimentation with nonlinear layouts where cube-corner voxels are made a bit smaller, and cube-center voxels are made a bit larger.

This design choice has far-reaching implications, and I'm not sure whether it's within scope for this PR. I had decided to put the seam between the current approach and this new `ChunkLayout` approach here, but if you're a fan of this idea, I can add it to `sim_config` in place of `chunk_size`.

common/src/graph_collision.rs
Ralith: Ah, I see. Do I understand correctly that we want both because the AABB-based approximation can be much less brute force than doing the sphere-based approximation for every neighbor?

common/src/graph_collision.rs
Ralith: Whoops, not sure how I missed that. LGTM.

common/src/character_controller.rs
Ralith: Let's err on the side of avoiding scope creep.

common/src/graph_collision.rs
patowen: Yeah. The sphere-based approximation alone is quite brute-force, as it doesn't depend on the direction of the ray at all. It checks whether the bounding sphere and the ray start intersects the bounding sphere of the node, and it doesn't distinguish between chunks within a given node.

The AABB approximation is nice because it checks an adjacent chunk whenever a ray gets too close to the corresponding face of the current chunk, which is quite intuitive and not very expensive to calculate.

common/src/graph_collision.rs
patowen: I went ahead and added the test case.

common/src/graph_collision.rs
patowen: There isn't a clear reason why `collider_radius` is on the right side instead of the left, so it is pretty well hidden because of that. I think it's a holdover from when I visualized this as ray tracing with inflated voxels instead of sphere casting with regular voxels.

I've moved it over to the left side where it's easier to see.

common/src/graph_collision.rs
patowen: I adjusted the comments a bit to hopefully make this a bit less confusing for future readers of this code.

common/src/graph_collision.rs
```diff
@@ -0,0 +1,559 @@
...
+
+/// Performs sphere casting (swept collision query) against the voxels in the `DualGraph`
+///
+/// The `ray` parameter and any resulting hit normals are given in the local coordinate system of `position`.
+///
+/// The `tanh_distance` is the hyperbolic tangent of the cast_distance, or the distance along the ray to check for hits.
+///
+/// This function may return a `SphereCastError` if not enough chunks are generated, even if the ray never reaches an
+/// ungenerated chunk. To prevent these errors, make sure that the distance between the ray's start point and the closest
+/// ungenerated chunk's center is less than `cast_distance + collider_radius + dodeca::BOUNDING_SPHERE_RADIUS`
```
Ralith, line 21: Should this say "greater than", and with respect to the node center, rather than chunk center?

Ralith (changes requested):

common/src/graph_collision.rs
patowen: Good catch. I was picturing as a ray needing to be short enough, hence the "less than", but I did get it backwards.

patowen:
Ah, I just today realized that Github has a "Compare" button that works with force pushes. No wonder you didn't need me to prepare a diff.

Ralith (approved):

patowen:
Thanks for merging! I'm optimistic that we're past the hardest part of getting player physics that works similarly to the collision demo.

Besides code style, the biggest remaining open questions are how to make sure prediction logic feels right, especially when jumping, but assuming these questions can get resolved, we should be pretty close to being able to close https://github.com/Ralith/hypermine/issues/127. Depending on what criteria we're looking for, we might be already be able to close https://github.com/Ralith/hypermine/issues/34.

PR #265, opened 2023-03-01
Allow the character to slide against walls

patowen:
As a follow-up to https://github.com/Ralith/hypermine/pull/262, this PR improves the collision resolution logic of the character so that it slides against walls instead of getting stuck against them.

Unfortunately, the precision of the original collision logic was insufficient to avoid noticeable artifacts made visible by the updated collision resolution logic, so this PR also tweaks this collision logic to use more numerically stable formulas at the cost of being (hopefully only slightly) less efficient for sphere/vertex collisions.

### DIFF
--- a/client/src/graphics/draw.rs
+++ b/client/src/graphics/draw.rs
@@ -265,7 +265,7 @@ impl Draw {
     ) {
         let draw_started = Instant::now();
         let view = sim.view();
-        let projection = frustum.projection(0.01);
+        let projection = frustum.projection(1.0e-4);
         let view_projection = projection.matrix() * math::mtranspose(&view.local);
         self.loader.drive();
 
--- a/common/src/character_controller.rs
+++ b/common/src/character_controller.rs
@@ -58,18 +58,19 @@ impl CharacterControllerPass<'_> {
                 *self.velocity += current_to_target_velocity;
             }
 
-            // Set expected displacement by using the average of the old velocity and new velocity,
+            // Estimate the average velocity by using the average of the old velocity and new velocity,
             // which has the effect of modeling a velocity that changes linearly over the timestep.
             // This is necessary to avoid the following two issues:
             // 1. Input lag, which would occur if only the old velocity was used
             // 2. Movement artifacts, which would occur if only the new velocity was used. One
             //    example of such an artifact is the character moving backwards slightly when they
             //    stop moving after releasing a direction key.
-            let expected_displacement = (*self.velocity + old_velocity) * 0.5 * self.dt_seconds;
+            let estimated_average_velocity = (*self.velocity + old_velocity) * 0.5;
 
             // Update position with collision checking
+            let expected_displacement = estimated_average_velocity * self.dt_seconds;
             let collision_checking_result = self.check_collision(&expected_displacement);
-            self.position.local *= collision_checking_result.allowed_displacement;
+            self.position.local *= collision_checking_result.displacement_transform;
             if let Some(collision) = collision_checking_result.collision {
                 *self.velocity = na::Vector3::zeros();
                 // We are not using collision normals yet, so print them to the console to allow
@@ -124,26 +125,24 @@ impl CharacterControllerPass<'_> {
             }
         };
 
-        let allowed_displacement = math::translate(
-            &ray.position,
-            &math::lorentz_normalize(
-                &ray.ray_point(
-                    cast_hit
-                        .as_ref()
-                        .map_or(tanh_distance, |hit| hit.tanh_distance),
-                ),
-            ),
-        );
+        let distance = cast_hit
+            .as_ref()
+            .map_or(tanh_distance, |hit| hit.tanh_distance)
+            .atanh();
+
+        let displacement_vector = displacement_normalized.xyz() * distance;
+        let displacement_transform = math::translate_along(&displacement_vector);
 
         CollisionCheckingResult {
-            allowed_displacement,
+            displacement_vector,
+            displacement_transform,
             collision: cast_hit.map(|hit| Collision {
                 // `CastEndpoint` has its `normal` given relative to the character's original position,
                 // but we want the normal relative to the character after the character moves to meet the wall.
                 // This normal now represents a contact point at the origin, so we omit the w-coordinate
                 // to ensure that it's orthogonal to the origin.
                 normal: na::UnitVector3::new_normalize(
-                    (math::mtranspose(&allowed_displacement) * hit.normal).xyz(),
+                    (math::mtranspose(&displacement_transform) * hit.normal).xyz(),
                 ),
             }),
         }
@@ -151,9 +150,15 @@ impl CharacterControllerPass<'_> {
 }
 
 struct CollisionCheckingResult {
+    ///The displacement allowed by the character before hitting a wall. The result of
+    /// `math::translate_along(&displacement_vector)` is `displacement_transform`.
+    #[allow(dead_code)]
+    displacement_vector: na::Vector3<f32>,
+
     /// Multiplying the character's position by this matrix will move the character as far as it can up to its intended
     /// displacement until it hits the wall.
-    allowed_displacement: na::Matrix4<f32>,
+    displacement_transform: na::Matrix4<f32>,
+
     collision: Option<Collision>,
 }
 
@@ -169,7 +174,8 @@ impl CollisionCheckingResult {
     /// and has nothing to check collision against. Also useful as a last resort fallback if an unexpected error occurs.
     fn stationary() -> CollisionCheckingResult {
         CollisionCheckingResult {
-            allowed_displacement: na::Matrix4::identity(),
+            displacement_vector: na::Vector3::zeros(),
+            displacement_transform: na::Matrix4::identity(),
             collision: None,
         }
     }
--- a/common/src/character_controller.rs
+++ b/common/src/character_controller.rs
@@ -1,4 +1,4 @@
-use tracing::{error, info};
+use tracing::{error, warn};
 
 use crate::{
     graph_collision, math,
@@ -67,17 +67,7 @@ impl CharacterControllerPass<'_> {
             //    stop moving after releasing a direction key.
             let estimated_average_velocity = (*self.velocity + old_velocity) * 0.5;
 
-            // Update position with collision checking
-            let expected_displacement = estimated_average_velocity * self.dt_seconds;
-            let collision_checking_result = self.check_collision(&expected_displacement);
-            self.position.local *= collision_checking_result.displacement_transform;
-            if let Some(collision) = collision_checking_result.collision {
-                *self.velocity = na::Vector3::zeros();
-                // We are not using collision normals yet, so print them to the console to allow
-                // sanity checking. Note that the "orientation" quaternion is not used here, so the
-                // numbers will only make sense if the character doesn't look around.
-                info!("Collision: normal = {:?}", collision.normal);
-            }
+            self.apply_velocity(&estimated_average_velocity);
         }
 
         // Renormalize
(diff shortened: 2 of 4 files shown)


client/src/graphics/draw.rs
```diff
@@ -265,7 +265,7 @@ impl Draw {
     ) {
         let draw_started = Instant::now();
         let view = sim.view();
-        let projection = frustum.projection(0.01);
+        let projection = frustum.projection(0.005);
```
Ralith, line 268: Should this be derived from the radius, or even the default value, directly? A hard-coded constant is liable to become obsolete again.

common/src/character_controller.rs
```diff
@@ -124,36 +125,40 @@ impl CharacterControllerPass<'_> {
...
-                    (math::mtranspose(&allowed_displacement) * hit.normal).xyz(),
+                    (math::mtranspose(&displacement_transform) * hit.normal).xyz(),
                 ),
             }),
         }
     }
 }
 
 struct CollisionCheckingResult {
+    ///The displacement allowed by the character before hitting a wall. The result of
```
Ralith, line 223: "allowed by" or "allowed for"? Also, leading space.

common/src/character_controller.rs
```diff
@@ -91,6 +81,46 @@ impl CharacterControllerPass<'_> {
...
+    /// Updates the position based on the given average velocity while handling collisions. Also updates the velocity
+    /// based on collisions that occur.
+    fn apply_velocity(&mut self, estimated_average_velocity: &na::Vector3<f32>) {
+        // To prevent an unbounded runtime, we only allow a limited number of collisions to be processed in
+        // a single step. If the player encounters excessively complex geometry, it is possible to hit this limit,
+        // in which case further movement processing is delayed until the next time step.
+        const MAX_COLLISION_ITERATIONS: u32 = 5;
+
+        let mut expected_displacement = estimated_average_velocity * self.dt_seconds;
+        let mut active_normals = Vec::<na::UnitVector3<f32>>::with_capacity(3);
```
Ralith, line 93: Is accumulating and simultaneously solving multiple normals necessary? Would it be sufficient to instead treat each collision independently, each time subtracting the portion of the remaining `expected_displacement` that would move into that individual normal?

common/src/character_controller.rs
```diff
@@ -149,10 +179,49 @@ impl CharacterControllerPass<'_> {
     }
 }
 
+/// Modifies the `subject` by a linear combination of the `normals` to ensure that it is approximately
+/// orthogonal to all the normals. The normals are assumed to be linearly independent, and, assuming the final
+/// result is nonzero, a small correction is applied to ensure that the subject is moving away from the surfaces
+/// the normals represent even when floating point approximation is involved.
```
Ralith, line 185: What if the character is moving along a concave curve? I think that could produce an arbitrarily large number of linearly dependent vectors.

common/src/chunk_collision.rs
```diff
@@ -288,40 +313,42 @@ fn solve_sphere_plane_intersection(
...
     ray: &Ray,
-    line_position: &na::Vector4<f32>,
-    line_direction: &na::Vector4<f32>,
-    cosh_radius: f32,
+    line_normal0: &na::Vector4<f32>,
+    line_normal1: &na::Vector4<f32>,
+    sinh_radius: f32,
 ) -> Option<f32> {
-    // This could be made more numerically stable by using a formula that depends on sinh_radius,
-    // but the precision requirements of collision should be pretty lax.
```
Ralith, line 296: Why is sinh more precise? Can we capture the required precision in a test that the cosh version fails?

Ralith (changes requested):

common/src/chunk_collision.rs
patowen: Might be a good idea. I'll see if I can write a test for this.

To answer your question, `cosh(x)` is approximately `1 + x^2`, while `sinh(x)` is approximately `x`. When `x` is small, then `cosh(x)` is very close to 1, and solving for `x` geometrically is like trying to find the intersection point of two almost-parallel lines.

common/src/character_controller.rs
patowen: Each step of the concave curve pushes the player away from the previous normal, so only the most recent normal gets applied. This function then only gets called with one normal, satisfying its constraint.

common/src/character_controller.rs
patowen: This causes problems if the player moves across an inner edge with an acute angle. If each normal is treated one-at-a-time, the first surface will push the player towards the second surface, and the second surface will then push the player back towards the first surface, an infinite loop that makes forward progress impossible.

client/src/graphics/draw.rs
patowen: I'm not sure. One option is to make it extremely small. I believe `frustum.rs` uses a trick to take advantage of floating point to better distribute z-values in the depth buffer, which means that we should be able to go much smaller without z-fighting.

If we do want to make it configurable, it probably makes sense to put it in the client-specific config, and the near-plane distance can be given in "meters". However, there is an argument for putting it in sim_config so that the server can tune it based on other parameters that live in sim_config.

common/src/character_controller.rs
Ralith: Ah, missed the `retain`.

common/src/character_controller.rs
Ralith: Ah, missed the `retain`.

common/src/character_controller.rs
Ralith: Not an infinite loop, since the maximum iterations are bounded, and motion perpendicular to the axis of the angle should still be possible.

patowen: Right, the loop wouldn't be infinite, but if we reach the bound, the behavior usually isn't ideal.

The issue is that the motion vector after being affected by one normal at a time (without any consideration of the other normal) will always point towards at least one of the walls of the wedge.

Without this extra logic, it would go as follows in Euclidean space: Some forward progress is made originally before hitting the first wall. Then, some more forward progress is made, sliding against the first wall, before hitting the second wall. However, now that the player is now touching both walls at once, as long as the `expected_displacement` faces one of the two walls, a collision is always reported immediately without any forward movement (`tanh_distance == 0`), so the player gets stuck.

In hyperbolic space, since the wall diverges a bit from the player, the player wouldn't quite hit both walls at once, so some forward progress would be made in every step, but that forward progress would exponentially decay, preventing collision from getting resolved before `MAX_COLLISION_ITERATIONS`. At the scales we normally deal with, it basically feels exactly like the Euclidean case.

common/src/chunk_collision.rs
patowen: I've added a test to demonstrate the difference for edges (and I believe it's safe to extrapolate this reasoning for vertices). The test is a bit lengthy and convoluted, so I'm definitely open to feedback on it.

client/src/graphics/draw.rs
patowen: It looks like the link that explains the reverse depth buffer in `frustum.rs` is broken. I can find it in the Wayback Machine, but an alternative link that describes the same thing is at https://developer.nvidia.com/content/depth-precision-visualized.

Even setting znear to 1.0e-30, I don't notice any z-fighting. It does break if I set it to 0, although there may be a workaround. Even if there's a workaround, it could potentially result in some pathological visual scenarios when going through a wall.

In short, for now, I would be in favor of just setting it to 1.0e-4 or something for now.

Ralith: I wouldn't read too much into visual tests using our current scene, since voxel geometry won't tend to have overlapping nearby faces very often.

> If we do want to make it configurable

I didn't mean to suggest that it be independently configurable, but rather that it be the largest value that the player collision geometry (governed by the server) ensures won't easily result in clipping.

> it looks like the the main fundamental limitation to decreasing znear is that typical depth values are scaled proportional to znear

Does that analysis account for the perspective divide?

patowen: > I wouldn't read too much into visual tests using our current scene, since voxel geometry won't tend to have overlapping nearby faces very often.

That's a fair point. I've mainly gone off of mathematical analysis, and the visual tests were just final sanity checks. The resource at https://web.archive.org/web/20220807195222/http://dev.theomader.com/depth-precision/ has an interactive module that helps illustrate this benefit.

> I didn't mean to suggest that it be independently configurable, but rather that it be the largest value that the player collision geometry (governed by the server) ensures won't easily result in clipping.

My main concern there is that it causes configuration values to have second-order effects. Unless we have a clear reason that we want `znear` to be as large as possible without causing clipping, keeping it fixed leaves things more predictable.

> Does that analysis account for the perspective divide?

It does. With `zfar == 1`, we have `za == zb == -znear / (znear - 1.0)`. Looking at only the z and w coordinates, the projection matrix is
```
  za   zb
-1.0  0.0
```
If you multiply this by the column vector `[-z*w, w]`, where `w` is arbitrary, and `z` is between 0 and 1 (I negated z in `-z*w` because our projection matrix operates on vectors with negative z values), then you get the column vector
```
znear*w*(1-z) / (1-znear)
z*w
```
which divides out to `znear*(1-z) / (1-znear)`, which is roughly proportional to `znear` when `znear` is small.

Ralith: Good enough for me! Can always revisit if we see issues in practice.

common/src/character_controller.rs
Ralith: > as long as the `expected_displacement` faces one of the two walls, a collision is always reported immediately without any forward movement

Why would `expected_displacement` face either wall, after the component facing them was subtracted out in the first two iterations? Isn't it at that point left necessarily either perpendicular to or away from them?

common/src/chunk_collision.rs
Ralith: LGTM. I wouldn't mind trimming out the parts that demonstrate the less precise method; my main goal is just having some insurance against future regression.

patowen: That makes sense to me. I definitely agree with you here, so I'll go ahead and get rid of the other checks.

common/src/character_controller.rs
patowen: Here's an image which demonstrates the issue. It shows 4 collision steps of the sphere approaching a wedge between two planes. There is movement in the z-direction that we don't see in this diagram because I made it 2D for simplicity. The z-component of these vectors never get redirected.

image

In each of the images, the sphere initially has an expected displacement equal to the blue vector, and subtracting out the wall normal turns it into the green vector. The green vector in the previous image corresponds to the blue vector in the next image.

In stage 1, the sphere just hit one of the two walls, causing it to be redirected to slide against this wall.

In stage 2, the sphere just hit the other wall, causing it to get redirected downwards again. I draw the vectors starting on the relevant contact point to help better illustrate which wall is triggering this redirection, but I could have just as easily drawn them anywhere else.

In stage 3, the sphere immediately hits the lower wall without moving, causing the expected displacement to redirect upwards again. It cannot make any progress in the z-direction because it immediately hit a wall.

In stage 4, the blue vector is facing the same vector as in stage 2, so we get a repeat. Stages 2 and 3 effectively repeat forever until we reach the maximum iteration count, causing the sphere to get stuck.

The main issue is that the walls are not perpendicular to each other, so if we subtract out each wall's normal one at a time, they'll always undo each other's work. Having both normals act at once lets us handle this case correctly.

Ralith: Good illustration, thanks! I wonder why I've never ran into this issue in other projects... certainly we're more likely here than most to be walking into/over/along weird angles.

Ralith (approved):

Issue #253, opened 2023-01-08
Create invariant identifier for nodes using cryptographic hash function


patowen:
Currently, the only way to uniquely identify a node in a `Graph` is via `NodeId`, which is a counter that increments in the order that nodes get added to the graph. However, this requires the server and client to keep their graph structures in sync to ensure that a given `NodeId` for the server corresponds to the same `NodeId` for the client.

An alternative identifier worth considering is one that only depends on where the node is in the graph and not when it was generated. It is possible to create such an identifier by using a method similar to how "spice" currently works with world generation.

Here is one possible implementation to generate this identifier (which will be called the "node hash" in this description to avoid confusing it with `NodeId`):
- The root node has a hardcoded (or configured) node hash
- To produce the hash of any other node N, choose the node corresponding to a consistently-chosen descender (for instance, picking the descender that comes first in alphabetical order), and call it M. Then, take `(M.node_hash, N.parent_side)` (where `N.parent_side` is the name of the dodeca side connecting M and N) and hash it with a cryptographic hash function like blake2b to produce `N.node_hash`.

If the node hash is used instead of `NodeId` for communication between the server and client, the server would no longer need to care about the order it sends nodes to clients, and clients would be free to expand the graph without needing the server to do so first.

To find a node given its hash, a hashmap of node hash to `NodeId` can be used. If a node is not yet generated, it is impossible to locate it using just its hash alone, so a server transmitting the position of an entity would still need to give clients sufficient context on how to get there from the previous position or from the root.

Ralith:
Bonus: when a reference to a node is only of interest if the node is already known (e.g. lies within the player's view distance), we don't actually need any information about where that node lies in the graph, since either we can look that up on the fly based on the hash, or it's guaranteed to be irrelevant. This will make it very simple and efficient to e.g. transmit and persist entity positions and edited voxel data.

PR #268, opened 2023-03-11
Write save files

Ralith:
Need #253 before we can save entity data, and therefore actually load any information, but this gets the basic plumbing set up.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -4,9 +4,10 @@ use std::{
 };
 
 use client::{graphics, metrics, net, Config, Sim};
+use save::Save;
 
 use ash::extensions::khr;
-use tracing::error_span;
+use tracing::{error_span, info};
 
 fn main() {
     // Set up logging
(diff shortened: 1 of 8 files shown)


client/src/main.rs
```diff
@@ -27,6 +28,10 @@ fn main() {
         let cert = cert.serialize_der().unwrap();
         let sim_cfg = config.local_simulation.clone();
 
+        let save = dirs.data_local_dir().join("default.save");
+        info!("using save file {}", save.display());
+        let save = Save::open(&save, config.local_simulation.chunk_size).unwrap();
```
patowen, line 33: On my first attempt running this without doing any prior setup, I got the following error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Db(DbError(Io(Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." })))', client\src\main.rs:33:74
```

If I manually create the "data" directory, this error goes away. As far as I can tell, `redb` doesn't have a way to ensure that the directory exists when creating the directory, so to avoid needing additional setup, an mkdirs-equivalent call is likely needed.

patowen (changes requested):
Looks good to me, apart from one edge case.

client/src/main.rs
Ralith: Good catch; added a call to `create_dir_all`.

PR #270, opened 2023-03-11
Path-hashed node identifiers

### LINKED ISSUE #253: Create invariant identifier for nodes using cryptographic hash function

Ralith:
Fixes #253.

It occurs to me that, while blake2b is known to be fast in bulk operation, our branching byte-at-a-time use case is unusual and might not be well studied/optimized for. Could be worth benchmarking and experimenting with other hash functions (maybe including less cryptographic ones) but this is probably not the worst place to start.

### DIFF
--- a/common/src/graph.rs
+++ b/common/src/graph.rs
@@ -4,6 +4,7 @@ use std::convert::TryFrom;
 use std::fmt;
 use std::num::NonZeroU32;
 
+use blake3::Hasher;
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -24,7 +25,7 @@ pub struct Graph<N> {
 impl<N> Graph<N> {
     pub fn new() -> Self {
         Self {
-            nodes: vec![Node::new(None, 0)],
+            nodes: vec![Node::new(None, 0, 0)],
             fresh: vec![NodeId::ROOT],
         }
     }
@@ -189,9 +190,25 @@ impl<N> Graph<N> {
         // Always create shorter nodes first so that self.nodes always puts parent nodes before their child nodes, enabling
         // graceful synchronization of the graph
         let shorter_neighbors = self.populate_shorter_neighbors_of_child(parent, side);
+
+        // Select the side along the canonical path from the origin to this node. Future work: make
+        // this the parent to reduce the number of concepts.
+        let (path_side, predecessor) = shorter_neighbors
+            .clone()
+            .chain(Some((side, parent)))
+            .min_by_key(|&(side, _)| side)
+            .unwrap();
+        let mut hasher = Hasher::new();
+        hasher.update(&self.nodes[predecessor.idx()].hash.to_le_bytes());
+        hasher.update(&[path_side as u8]);
+        let mut xof = hasher.finalize_xof();
+        let mut hash = [0; 16];
+        xof.fill(&mut hash);
+        let hash = u128::from_le_bytes(hash);
+
         let id = NodeId::from_idx(self.nodes.len());
         let length = self.nodes[parent.idx()].length + 1;
-        self.nodes.push(Node::new(Some(side), length));
+        self.nodes.push(Node::new(Some(side), length, hash));
         self.link_neighbors(id, parent, side);
         for (side, neighbor) in shorter_neighbors {
             self.link_neighbors(id, neighbor, side);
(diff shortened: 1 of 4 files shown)


patowen:
Oh, good point. It's unclear to me whether any non-cryptographic hash function would work, since this could make it trivial to break public servers or require us to gracefully handle hash collisions, neither of which seem ideal. I'm at least hopeful that the runtime of the hash function won't be the bottleneck, although it could affect the initial load time.

Ralith:
I guess computing a hash should be far cheaper than running terrain gen, for example, so probably it's not a big deal.

common/src/node.rs
```diff
@@ -52,6 +52,8 @@ pub struct Node {
     /// We can only populate chunks which lie within a cube of populated nodes, so nodes on the edge
     /// of the graph always have some `None` chunks.
     pub chunks: Chunks<Chunk>,
+    /// Entities whose origins lie within this node
```
patowen, line 55: I wonder whether we should be less specific and say something like "Entities currently associated with this node" here. On one hand, all our current entities should follow this rule, up to potentially some small buffer around each node in case floating point precision makes this ambiguous. On the other hand, this isn't a requirement at all, and we may still want to even tie entities without a `Position` component to the graph.

The main restriction is that the entity won't necessarily be loaded until the node is loaded, so for best results in the standard case, we do want this comment to be true.

common/src/worldgen.rs
```diff
@@ -695,6 +695,7 @@ mod test {
                     state
                 },
                 chunks: Chunks::default(),
+                entities: Vec::new(),
```
patowen, line 698: Is this redundant with the `graph_entities.rs` module? If not, we should document the purpose of both versions of the node -> entity mapping.

server/src/sim.rs
```diff
@@ -156,6 +163,17 @@ impl Sim {
                 input,
                 self.cfg.step_interval.as_secs_f32(),
             );
+            if prev_node != position.node {
```
patowen, line 166: Does the client need this data? Currently, I believe the client's version of `DualGraph` will always have an empty `entities` vec, although we already have something like this with the server's version of `DualGraph` always having `None` for the surface extraction.

patowen (changes requested):
Thanks for working on this!

common/src/worldgen.rs
Ralith: Forgot that existed. Swapped in.

server/src/sim.rs
Ralith: Clients already infer node -> entity mappings from position updates:

https://github.com/Ralith/hypermine/blob/45eaf09f93110751e46cfd9c26efe174c7adf7fb/client/src/sim.rs#L177-L180

common/src/node.rs
Ralith: Moot due to switch to `GraphEntities`.

common/src/graph.rs
```diff
@@ -274,15 +293,21 @@ struct Node<N> {
...
     length: u32,
     neighbors: [Option<NodeId>; SIDE_COUNT],
+    hash_state: Blake2b128,
+    hash: u128,
 }
 
 impl<N> Node<N> {
-    fn new(parent_side: Option<Side>, length: u32) -> Self {
+    fn new(parent_side: Option<Side>, length: u32, hash_state: Blake2b128) -> Self {
+        let hash = hash_state.clone().finalize();
```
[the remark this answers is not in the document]
Ralith: > Is it possible to reverse this operation, turning a hash into a `hash_state`?

Not through this library, and probably not at all, though I'm not an expert on the underlying theory.

> alternative implementation

I was wondering about this myself. Surprisingly, benchmarking indicates this is actually *faster*; maybe due to better memory locality with smaller nodes? Anyway, switched over. I'm certainly happy to reduce the necessary state and keep the door open to future cleverness.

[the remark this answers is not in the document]
Ralith: Better yet, with the stateless construction blake3 is another 30% faster thank blake2b!

server/src/sim.rs
```diff
@@ -30,6 +30,7 @@ pub struct Sim {
     graph: DualGraph,
     spawns: Vec<Entity>,
     despawns: Vec<EntityId>,
+    graph_entities: GraphEntities,
```
patowen, line 33: Oh, interesting! Previously, only the client kept track of this. No wonder it was so well-hidden.

patowen (approved):

PR #271, opened 2023-03-11
Isolate entity and voxel data in saves

Ralith:
Allows e.g. entity movements to be persisted without rewriting up to 67.5KiB (for a full node of 12-voxel chunks, uncompressed).

We could take this a step further and store each chunk separately, but that might reduce read performance, and voxel edits are expected to be infrequent.

### DIFF
--- a/save/src/lib.rs
+++ b/save/src/lib.rs
@@ -41,7 +41,8 @@ impl Save {
                     meta.insert(&[][..], &*compressed)?;
                     drop(meta);
 
-                    tx.open_table(NODE_TABLE)?;
+                    tx.open_table(VOXEL_NODE_TABLE)?;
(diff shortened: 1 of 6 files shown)


save/src/lib.rs
```diff
@@ -41,7 +41,7 @@ impl Save {
                     meta.insert(&[][..], &*compressed)?;
                     drop(meta);
 
-                    tx.open_table(NODE_TABLE)?;
+                    tx.open_table(VOXEL_NODE_TABLE)?;
```
patowen, line 44: Do we also want to create the `ENTITY_NODE_TABLE` here?

patowen (changes requested):
Apart from one potential missing `tx.open_table`, looks good to me!

PR #273, opened 2023-03-12
Intelligently pick the starting vertex for graph_collision

patowen:
This fixes an issue where we get inconclusive collision checking if the player moves into a node where the "A" chunk isn't loaded. This becomes an issue with smaller render distances.

### DIFF
--- a/common/src/graph_collision.rs
+++ b/common/src/graph_collision.rs
@@ -31,12 +31,24 @@ pub fn sphere_cast(
     // This `hit` variable gets updated over time before being returned.
     let mut hit: Option<GraphCastHit> = None;
 
+    // Pick the vertex closest to position.local as the vertex of the chunk to use to start collision checking
+    let start_vertex = Vertex::iter()
+        .map(|v| {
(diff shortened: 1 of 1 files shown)


common/src/graph_collision.rs
```diff
@@ -31,12 +31,24 @@ pub fn sphere_cast(
     // This `hit` variable gets updated over time before being returned.
     let mut hit: Option<GraphCastHit> = None;
 
+    // Pick the vertex closest to position.local as the vertex of the chunk to use to start collision checking
+    let start_vertex = Vertex::iter()
+        .map(|v| {
+            (
+                v,
+                (v.node_to_dual().cast::<f32>() * position.local * math::origin()).w,
```
patowen, line 39: I wanted to use `min_by_key` with this, but unless we use a crate that wraps `partial_cmp().unwrap()` into a type, I believe this is the best we can do.

Ralith: Reasonable.

PR #274, opened 2023-03-12
Dependency cleanup

Ralith:


client/src/lib.rs
```diff
@@ -10,7 +10,6 @@ macro_rules! cstr {
     }};
 }
 
-extern crate nalgebra as na;
```
patowen, line 13: We may run into trouble trying to use any of `nalgebra`'s macros, since last time I checked, procedural macros aren't hygienic enough to handle the crate being renamed in `cargo.toml`. That being said, we haven't used these macros yet, so it should be fine to switch back to this.

We should just be aware of this, as it could be a point of confusion in the future if we're wondering why nalgebra's procedural macros don't work.

Ralith: Oh, was that why this was like that? I thought it was just leftover from old rust.

patowen (approved):

client/src/lib.rs
patowen: Yeah, it was this commit: https://github.com/Ralith/hypermine/pull/230/commits/021e9cd93a6d1b4590d8997e06e7c42044373468

patowen (approved):

PR #276, opened 2023-03-13
Save exact character positions

Ralith:
Started off using `bincode`, discovered it did some silly things. Should switch network serialization to postcard as well.


patowen (approved):
Looks good to me! There's a reasonable chance we missed a subtle bug somewhere, but if so, we should be able to find/fix it when implementing loading.

PR #277, opened 2023-03-14
Update the character controller to allow walking and jumping

### LINKED ISSUE #60: Orient camera perpendicular to gravity
Clients should automatically send inputs that will orient their character consistently upright. We shouldn't force this on the server side, as that needlessly reduces flexibility.

### LINKED ISSUE #127: Walking

patowen:
Resolves #127, resolves #60

# Player movement
The rough high-level algorithm is as follows:
- Check if the character is on the ground
- Initiate a jump if appropriate
- Accelerate the player based on which keys are held down
- Apply gravity, air resistance, and speed cap
- Use the computed velocity to move the player, resolving collisions as necessary

# Camera orientation
The logic that keeps the player oriented vertically relative to the world is handled client-side.
- If the player is almost facing vertically, yaw and potentially pitch the view to reach alignment.
- Otherwise, roll the view to reach alignment.

There are several subtleties needed to do the following:
- Keep the camera alignment consistent during player movement
- Interpret mouse-based pitch and yaw controls
- Convert an arbitrary orientation to an aligned orientation (algorithm described above)
- Ensure that the player does not get lifted up from the ground by slanted walls


patowen:
Besides the things I've inevitably missed, there are a few changes that I know are needed before this PR is merged, but limits in my own level of experience and sense of good code structure mean that I will need feedback before I make these changes:
- A lot of `character_controller`'s methods have way too many arguments. (Need feedback on which abstractions make sense to combine arguments. Should there be a character state struct? Should functions take fewer mutable arguments in favor of return values? Should there be an overarching struct with a bunch of `&mut self` functions?)
- Client's `sim.rs` has too much functionality. I believe it would make sense to have client-side-only character controller logic in a separate file (Need feedback on whether you agree and how much of this refactor should be in a separate PR)
- There are multiple modules within `character_controller`, which should probably be separate files. (Need feedback on whether you agree with the idea of having `character_controller` be a folder within `common` with the `character_controller` module and its two submodules)

The character controller is definitely not perfect. The most glaring issue is that walking into an obtuse wall corner will result in some shaking. However, this is expected behavior, and fixing this I believe should be outside the scope of this PR. In addition, the necessary input lag when jumping means that some coyote time would likely be a good idea, but I also consider that out of scope.

I believe the most important aspects of the character controller are as follows:
- It should feel relatively fun and intuitive
- The code should be readable and malleable enough that improvements can be made by someone other than me.

Most other aspects would likely have a lower priority than adding other essential features to this game, like block manipulation and saving/loading.

patowen:
To better playtest the character controller, I created a temporary branch on my fork with block placing/breaking: https://github.com/patowen/hypermine/tree/collision-wip-block-placing-temp.

As of the writing of this comment, the character controller code between this branch and that branch are identical, and the only difference is that the `collision-wip-block-placing-temp` branch has hastily-put-together block manipulation functionality borrowed from the collision demo and adapted to partially work with netcode (but not multiplayer).

Ralith:
> If the player is almost facing vertically, [...] Otherwise, [...]

Could we simplify this to always rotating around whatever axis produces the minimum rotation?

>  I believe it would make sense to have client-side-only character controller logic in a separate file.

Strong agree. In an ideal world where we fully understand how to factor everything, I'd expect sim.rs to contain only the most trivial logic, instead mainly dispatching to other modules.

> Need feedback on whether you agree with the idea of having character_controller be a folder within common with the character_controller module and its two submodules

That seems fine.

> I believe the most important aspects of the character controller are as follows: [...]

SGTM.

patowen:
> Could we simplify this to always rotating around whatever axis produces the minimum rotation?

I didn't go this route because the math is more complicated, and I don't think it's any better for the player. One case to consider is the fact that a rotation about the look-axis (roll) is less jarring than a rotation about a perpendicular axis, since the object the player is looking at stays fixed. A roll is almost always ideal, and the main exception is when the camera is close enough to vertical that it's no longer obvious which way is right-way up, since then, the correction starts to feel random. That's when yaw and/or pitch becomes nicer.

For a concrete example, if the player is facing forward but is upside-down, I believe the right move is to turn it 180 degrees to make the player right-way-up, but one can reach a valid orientation by having it face straight up or straight down instead, which is only a 90 degree rotation.

All other suggestions seem good to me, and I should get a chance to address them soon.

client/src/local_character_controller.rs
```diff
@@ -0,0 +1,290 @@
...
+            * na::UnitQuaternion::from_axis_angle(&na::Vector3::x_axis(), delta_pitch)
+            * na::UnitQuaternion::from_axis_angle(&na::Vector3::z_axis(), delta_roll);
+    }
+
+    /// Rotates the camera's view with standard first-person walking simulator mouse controls. This function
+    /// is designed to be flexible enough to work with any starting orientation, but it works best when the
+    /// camera is level, not rolled to the left or right.
+    pub fn look_level(&mut self, delta_yaw: f32, delta_pitch: f32) {
+        // Get orientation-relative up
+        let up = self.orientation.conjugate() * self.up;
```
Ralith, line 67: Is `conjugate()` the same as `inverse()` here? Can we use that for better readability, if so?

client/src/local_character_controller.rs
```diff
@@ -0,0 +1,290 @@
...
+
+        if up.z.abs() < 0.9 {
+            // If facing not too vertically, roll the camera to make it level.
+            let delta_roll = -up.x.atan2(up.y);
+            self.orientation *=
+                na::UnitQuaternion::from_axis_angle(&na::Vector3::z_axis(), delta_roll);
+        } else if up.y > 0.0 {
+            // Otherwise, if not upside-down, pan the camera to make it level.
```
Ralith, line 113: s/pan/yaw/?

common/src/math.rs
```diff
@@ -152,6 +152,20 @@ pub fn project_to_plane<N: RealField + Copy>(
         * ((distance - subject.dot(normal)) / projection_direction.dot(normal));
 }
 
+/// Returns the UnitQuaternion that rotates the `from` vector to the `to` vector, or `None` if
+/// `from` and `to` face opposite directions such that their sum has norm less than `epsilon`.
+/// This version is more numerically stable than nalgebra's equivalent function.
```
Ralith, line 157: Should we push this upstream?

common/src/sim_config.rs
```diff
@@ -28,8 +28,22 @@ pub struct SimConfigRaw {
...
+    /// Character acceleration while in the air in m/s^2
+    pub air_acceleration: Option<f32>,
+    /// Acceleration of gravity in m/s^2
+    pub gravity_acceleration: Option<f32>,
+    /// Linear air resistance in (1/e)-lives per second
+    pub air_resistance: Option<f32>,
+    /// How fast the player jumps off the ground in m/s
+    pub jump_speed: Option<f32>,
+    /// How far away the player needs to be from the ground in meters to be considered in the air
+    pub ground_distance_tolerance: Option<f32>,
```
Ralith, line 46: Could we `#[derive(Serialize, Deserialize)]` for `CharacterConfig` and just use that here? Might play nicely with `#[serde(default)]`, too.

common/src/sim_config.rs
```diff
@@ -28,8 +28,22 @@ pub struct SimConfigRaw {
...
+    pub speed_cap: Option<f32>,
+    /// Maximum ground slope (0=horizontal, 1=45 degrees)
+    pub max_ground_slope: Option<f32>,
     /// Character acceleration while on the ground in m/s^2
     pub ground_acceleration: Option<f32>,
+    /// Character acceleration while in the air in m/s^2
+    pub air_acceleration: Option<f32>,
+    /// Acceleration of gravity in m/s^2
+    pub gravity_acceleration: Option<f32>,
+    /// Linear air resistance in (1/e)-lives per second
```
Ralith, line 41: lives?

common/src/character_controller/mod.rs
```diff
@@ -53,68 +56,190 @@ fn run_standard_character_step(
...
+    // Handle actual movement
+    apply_velocity(
+        ctx,
+        average_velocity,
+        position,
+        velocity,
+        &mut ground_normal,
+    );
+
+    *on_ground = ground_normal.is_some();
```
Ralith, line 111: It's a bit difficult to follow the evolution of `ground_normal`. I wonder if we could factor most of these `&mut` parameters out into fields on a common struct that these are all methods on?

common/src/character_controller/vector_bounds.rs
```diff
@@ -0,0 +1,298 @@
...
+        );
+    }
+
+    /// Checks whether `subject` satisfies the constraint given by `self`. Note that `check_vector` will
+    /// return `true` after a vector is constrained by `constrain_vector` with the same error margin, even
+    /// if it's perturbed slightly. However, that property only holds if the error margin is not too small.
+    fn check_vector(&self, subject: &na::Vector3<f32>, error_margin: f32) -> bool {
+        // An additional margin of error is needed when the bound is checked to ensure that an
+        // applied bound always passes the check.
+        let error_margin_factor_for_check = self.error_margin_factor - 0.5;
```
Ralith, line 170: Why 0.5 specifically? That seems like a significant figure.

common/src/character_controller/vector_bounds.rs
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/// Represents a single constraint for a vector. `VectorBound`s alone conceptually contain
+/// enough information to apply to a vector, but practically, one other piece of information
+/// is needed: `error_margin`, which exists in `VectorBoundGroup`.
+#[derive(Clone)]
+pub struct VectorBound {
+    normal: na::UnitVector3<f32>,
+    projection_direction: na::UnitVector3<f32>,
+    error_margin_factor: f32, // Margin of error when the bound is applied
```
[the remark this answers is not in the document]
patowen: Not in this commit. I probably should have waited until the "Add logic to prevent slanted walls from lifting the player up" commit before adding this multiplier feature. After that commit, the possible values are 1.0 and -1.0.

The logic in that commit is probably one of the most difficult-to-follow pieces of logic in the entire PR. I've made quite an effort to make it more follow-able, but it's definitely not perfect.

common/src/math.rs
patowen: I've thought about that. It's probably worth at least creating an issue for it, although I'll probably want to find more evidence that it's better. The current implementation calls into `scaled_rotation_between_axis`, which I don't think has an equivalent optimization, so this optimization would add a little bit of extra complexity to nalgebra.

common/src/sim_config.rs
patowen: It took me a while to remember what I meant by this, so I probably want to reword it. It is like "half-lives", but "1/e"-lives instead.

I'll reword this, as it's really just the factor applied to the velocity to get the acceleration caused by friction.

common/src/character_controller/mod.rs
patowen: There do seem to be two competing programming principles that I haven't fully worked out:
- Bundling common things together into a struct to reduce parameter count and reduce the amount of code that needs to be read and understood
- Explicitly passing in parameters so that reading the code makes it more apparent exactly what each section's dependencies are.

For this PR, I ended up erring on the side of the second bullet point for all the mutable parameters, but I'm definitely not convinced that's the right move.

These parameters can definitely be factored out into a struct (player-state?). One potential pitfall is that `ground_normal` has to be an `Option`, so if there's a struct for ground-specific movement, it needs to either handle the `None` case (even though flow control would never reach there) or unwrap it. Honestly, I think that's fine, but we do lose some of the correctness benefits.

common/src/character_controller/vector_bounds.rs
patowen: This definitely deserves at least a comment. There's a pretty concrete explanation for when `error_margin_factor` is 1.0:
- If we used 0.0, the error margin would be the same when constraining and checking, which means that the check might fail when it should succeed.
- If we used 1.0, the error margin when checking would be 0, which means that the code might falsely believe everything is okay, even when the vector is not facing far enough away from a wall. This could cause the character to get stuck, as a collision could be detected if machine precision causes collision checking to think the character is going towards that wall.

In short, the error margin factor can't be too close to the constraining error margin factor, nor can it be too close to 0. Therefore, I picked the value halfway in-between.

It might make more sense to use 2 and 1 instead of 1 and 0.5 (for error margin factor and the check delta).

client/src/local_character_controller.rs
patowen: Yeah, it's the same. I forgot `inverse()` was an option.

common/src/character_controller/vector_bounds.rs
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    /// Constrains `vector` with `new_bound` while keeping the existing constraints satisfied.
+    /// All projection transformations applied to `vector` are also applied
+    /// to `tagalong` to allow two vectors to be transformed consistently with each other.
+    pub fn apply_and_add_bound(
+        &mut self,
+        new_bound: VectorBound,
+        vector: &mut na::Vector3<f32>,
+        tagalong: Option<&mut na::Vector3<f32>>,
```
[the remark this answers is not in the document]
patowen: Possibly. The issue is that we have to extract the values out at the end if they're fields, and the names `vector` and `tagalong` aren't self-explanatory, causing a potential readability pitfall, and, we'd need an `unwrap` when extracting `tagalong` at the end.

Both of these could potentially be addressed with traits, but that might hurt readability even more.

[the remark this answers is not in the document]
Ralith: A trait isn't justified here.

The names aren't any more confusing as fields than as arguments, IMO. Still, they could be improved regardless. Could we name them for what they're actually used for in practice? It's not like we have wildly diverse callsites.

common/src/character_controller/mod.rs
Ralith: Yeah, it can be awkward. We could have multiple structs (they're temporaries anyway) but it can get a bit boilerplatey. Not necessarily any action required here.

patowen: Sounds good. In that case, I'll stick with the current implementation, but if you have any concrete suggestions here, I'm happy to try them out.

common/src/character_controller/vector_bounds.rs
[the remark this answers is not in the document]
patowen: Currently, there are two usages: `apply_velocity` uses it to update a velocity, and `get_ground_normal` uses it to update a displacement. I believe using a specific field name might prove difficult.

However, switching around the order of some multiplications etc. should be able to allow these two usages to be made more similar to each other, which is not something I had considered when I previously attempted to make them variables.

Another difficulty I had in my last attempt is that I had other methods that manipulated these vectors, causing borrow-check issues under certain situations. A lot has changed since then, so I'm reasonably optimistic about your suggestion.

I'm always happy to reduce the parameter count.

common/src/sim_config.rs
patowen: I think that's a good idea, but I won't have time to get to this for a while. I plan to update this PR implementing your other suggestions. I chose not to prioritize this because I don't think it will affect how understandable the PR is to review.

patowen:
I am getting some "WARN Unsatisfied existing bound is parallel to new bound. Is the character squeezed between two walls?" messages (a warning I added for an edge case that should ideally be rare). I'm not sure whether my latest changes on July 4th caused them or whether it was the case before then, but when I have more time, I should be able to find the root cause.

common/src/character_controller/vector_bounds.rs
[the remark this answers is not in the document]
patowen: I've replaced this with a boolean, since that makes the logic much easier to follow in my opinion. Thanks for calling this out!

common/src/sim_config.rs
patowen: I finally have time to work on this again, so I made the suggested change and moved `CharacterConfig` into the `sim_config` module.

common/src/math.rs
patowen: I filed https://github.com/dimforge/nalgebra/issues/1262 so that this change can hopefully be pushed upstream one way or another. This issue also has a concrete example of the kinds of improvements we see.

Without this new implementation, I was getting noticeable jitter in the player camera orientation.

common/src/node.rs
```diff
@@ -31,6 +32,16 @@ impl DualGraph {
...
     }
+
+    /// Returns the up-direction relative to the given position, or `None` if the
+    /// position is in an unpopulated node.
+    pub fn get_relative_up(&self, position: &Position) -> Option<na::UnitVector3<f32>> {
+        self.get(position.node).as_ref().map(|n| {
+            na::UnitVector3::new_normalize(
+                (math::mtranspose(&position.local) * n.state.up_direction()).xyz(),
+            )
+        })
```
Ralith, line 43: You could use `?` to avoid some rightward drift here.

client/src/graphics/window.rs
```diff
@@ -222,6 +224,12 @@ impl Window {
                         VirtualKeyCode::F => {
                             down = state == ElementState::Pressed;
                         }
+                        VirtualKeyCode::Space => {
+                            if !jump {
+                                self.sim.set_jump_pressed_true();
```
Ralith, line 229: Is it intended that this be called when the key is released?

common/src/sim_config.rs
Ralith: Original question still stands -- could we have a `pub character: CharacterConfig` field here rather than effectively writing the struct twice?

patowen: Oh, I had misread your suggestion. I did create `CharacterConfig`, but I didn't create `CharacterConfigRaw`.

I don't think I can make this into `pub character: CharacterConfig` because of the unit-conversion (multiplication by `meters_to_absolute`). There might be a serde feature I'm unaware of, though.

client/src/graphics/window.rs
patowen: It is not. Good catch.

Jumping is set up so that you keep jumping while holding space, and you jump at least once when you press space regardless of framerate. This logic ends up hiding the bug you noticed.

common/src/sim_config.rs
Ralith: Oh, I missed that they were different units. All the workarounds I can immediately think of are similarly annoying to the status quo, unfortunately. No specific need to add a `...Raw` variant for its own sake, though maybe the uniformity would be nice.

patowen: I went ahead and added a `CharacterConfigRaw` for uniformity, as it should hopefully make custom config files easier to read as well.

patowen: > All the workarounds I can immediately think of are similarly annoying to the status quo, unfortunately.

I'm not too surprised. If serde had a fancy feature that allowed the unit conversion to be done, that feature would require either backtracking or storing intermediate values somewhere, which likely falls outside its design constraints.

Thinking aloud because it definitely _feels_ like there should be a better way.

One could imagine some fancy newtype-based system to store the meter-based config values originally but then replace those data slots with the absolute-unit-based values (without the code repetition we currently have), but I don't know how to do that in Rust, and I don't know whether it's possible without using custom macros.

If there is something like a `#[derive(Into)]`, that converts corresponding fields, the newtype idea could potentially work with the help of generics, but even then, we would still need to have a version that takes in some kind of context so that we can pass in `meters_to_absolute`. Even if it is possible, it seems overengineered (especially considering the rule of 3, and the fact that `SimConfig` may be altered in-game in the future).

I agree that it's probably not worth pursuing now. Perhaps if/when this repetition becomes unwieldy as additional configuration options are added, it would be worth revisiting.

Ralith: I don't usually take it very seriusly, but the frunk crate's labeled generics might be handy here to at least automate the conversion... but of comical overkill though.

common/src/sim_config.rs
```diff
@@ -24,14 +24,9 @@ pub struct SimConfigRaw {
...
-    pub no_clip_movement_speed: Option<f32>,
-    /// Character maximumum movement speed while on the ground in m/s
-    pub max_ground_speed: Option<f32>,
-    /// Character acceleration while on the ground in m/s^2
-    pub ground_acceleration: Option<f32>,
-    /// Radius of the character in meters
-    pub character_radius: Option<f32>,
+    /// Static configuration information relevant to character physics
+    #[serde(default)]
+    pub character_config: CharacterConfigRaw,
```
Ralith, line 29: nit: `_config` is redundant; the user already knows they're editing a config.

common/src/sim_config.rs
```diff
@@ -42,10 +37,7 @@ pub struct SimConfig {
     pub view_distance: f32,
     pub input_queue_size: Duration,
     pub chunk_size: u8,
-    pub no_clip_movement_speed: f32,
-    pub max_ground_speed: f32,
-    pub ground_acceleration: f32,
-    pub character_radius: f32,
+    pub character_config: CharacterConfig,
```
Ralith, line 40: Similarly here.

Ralith (approved):

common/src/sim_config.rs
patowen: Interesting, I haven't heard of the frunk crate before, so I might check it out sometime.

PR #308, opened 2023-09-09
Fix server crash on client disconnect due to incomplete entity removal

patowen:
There is an `unwrap` error when setting `q` in `snapshot_node`, since it uses `graph_entities` as the source of truth of entities that exist, but `graph_entities` was never updated on entity removal.


Ralith (approved):
Nice catch!

PR #309, opened 2023-09-11
Replace NodeId implementation with global hash

patowen:
Having `NodeId` be the global hash that is guaranteed to be consistent across the server and its clients allows simpler reasoning about code. The main downsides are
- It's 4 times larger, being a `u128` instead of a `u32` internally, which could potentially impact performance.
- The graph has to use a map instead of an array for its internal data representation, also potentially impacting performance.

The advantages are
- There is only one `NodeId`, removing ambiguity in discussions and in variable names etc.
- There is no risk of introducing bugs later on by accidentally assuming that a local `NodeId` means the same on the server and client.

I don't expect `NodeId`'s size to be the source of any bottlenecks, as each node is a large enough part of the data structure, that the handling of individual voxels and/or the matrix math for each node is likely to take much more processing time. Because of this, I think it's likely that this change is worthwhile, but I'm not 100% convinced.

The most complicated change is the rewriting of the `tree()` traversal function in `Graph`, since the original design was not compatible with the hashmap-based graph implementation, so it was replaced with a breadth-first search. It is only used for synchronizing the graph between the server and client. Ideally, we wouldn't need to do that anymore, but trying to naively remove this functionality would introduce a subtle bug because the server could move the player or another entity to a node the client doesn't know about, and the client does not and cannot properly handle that.

### DIFF
--- a/common/src/graph.rs
+++ b/common/src/graph.rs
@@ -1,10 +1,9 @@
 #![allow(clippy::len_without_is_empty)]
 
-use std::convert::TryFrom;
-use std::fmt;
-use std::num::NonZeroU32;
+use std::collections::VecDeque;
 
 use blake3::Hasher;
+use fxhash::{FxHashMap, FxHashSet};
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -16,7 +15,7 @@ use crate::{
 /// Graph of the right dodecahedral tiling of H^3
 #[derive(Debug, Clone)]
 pub struct Graph<N> {
-    nodes: Vec<Node<N>>,
+    nodes: FxHashMap<NodeId, Node<N>>,
     /// This field stores implicitly added nodes to ensure that they're initialized in the correct
     /// order
     fresh: Vec<NodeId>,
@@ -24,8 +23,10 @@ pub struct Graph<N> {
 
 impl<N> Graph<N> {
     pub fn new() -> Self {
+        let mut nodes = FxHashMap::default();
+        nodes.insert(NodeId::ROOT, Node::new(None, 0));
         Self {
-            nodes: vec![Node::new(None, 0, 0)],
+            nodes,
             fresh: vec![NodeId::ROOT],
         }
     }
@@ -37,7 +38,7 @@ impl<N> Graph<N> {
 
     #[inline]
     pub fn contains(&self, node: NodeId) -> bool {
-        node.0.get() as usize <= self.nodes.len()
+        self.nodes.contains_key(&node)
     }
 
     /// Nodes created since the last call to `clear_fresh`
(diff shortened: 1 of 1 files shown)


Ralith:
> There is only one NodeId, removing ambiguity in discussions and in variable names etc.

I think this is compelling. We can always reintroduce local IDs if this shows up in profiling in the future, and otherwise we'll benefit from a simpler model.

> There is no risk of introducing bugs later on by accidentally assuming that a local NodeId means the same on the server and client.

I don't think this is a significant risk; we should just never transmit NodeIds over the wire, which shouldn't be too hard to keep track of.

> Ideally, we wouldn't need to do that anymore, but trying to naively remove this functionality would introduce a subtle bug because the server could move the player or another entity to a node the client doesn't know about, and the client does not and cannot properly handle that.

We should treat such behavior from the server as a critical bug. That said, it's good to make changes like this incremental, so it makes sense to get rid of graph sync in a second pass.

common/src/graph.rs
```diff
@@ -16,16 +15,18 @@ use crate::{
 /// Graph of the right dodecahedral tiling of H^3
 #[derive(Debug, Clone)]
 pub struct Graph<N> {
-    nodes: Vec<Node<N>>,
+    nodes: FxHashMap<NodeId, Node<N>>,
```
Ralith, line 18: Because `NodeId` is already uniformly distributed, we could use an identity hasher to make access even cheaper. Probably won't matter, but it'd be cute.

common/src/graph.rs
```diff
@@ -315,27 +290,58 @@ impl<N> Node<N> {
...
+    }
 }
 
 impl<N> Iterator for TreeIter<'_, N> {
     type Item = (Side, NodeId);
 
     fn next(&mut self) -> Option<Self::Item> {
-        let (node, rest) = self.remaining.split_first()?;
-        self.remaining = rest;
-        self.id = NodeId::from_idx(self.id.idx() + 1);
```
Ralith, line 329: Was `self.id` dead all along? Geez.

Ralith (approved):

common/src/graph.rs
patowen: Oh yeah, I had considered that but didn't know how to set up such a hasher (or what it was called). It looks like it would need a new crate or a manual implementation, so it probably makes sense to leave this as-is for the sake of code simplicity, even though it does seem like a bit of a waste.

common/src/graph.rs
patowen: Oh, so it is. I didn't realize that while making this PR (since the implementation had to change quite a bit), but it does look like the only read of `self.id` is used to set `self.id`.

patowen:
> > There is only one NodeId, removing ambiguity in discussions and in variable names etc.
> 
> I think this is compelling. We can always reintroduce local IDs if this shows up in profiling in the future, and otherwise we'll benefit from a simpler model.
> 
> > There is no risk of introducing bugs later on by accidentally assuming that a local NodeId means the same on the server and client.
> 
> I don't think this is a significant risk; we should just never transmit NodeIds over the wire, which shouldn't be too hard to keep track of.
> 
> > Ideally, we wouldn't need to do that anymore, but trying to naively remove this functionality would introduce a subtle bug because the server could move the player or another entity to a node the client doesn't know about, and the client does not and cannot properly handle that.
> 
> We should treat such behavior from the server as a critical bug. That said, it's good to make changes like this incremental, so it makes sense to get rid of graph sync in a second pass.

All this makes sense to me. I believe the main feature we would need to get rid of this `TreeIter` requirement would be for the server to tell the player where they are with the full node path and to decide when to provide that info (so that clients know which graph nodes they _must_ generate).

Thanks for merging!

PR #314, opened 2023-10-15
Delay client Sim creation until server parameters are received

patowen:
This allows the client's `Sim` to store a `SimConfig` without wrapping it in an `Option`. The main downside is that it slightly complicates `window.rs` and `draw.rs`, but such complication is likely temporary and would go away in a future refactor (such as if a "connecting" screen is created, diverting control flow earlier pre-ServerHello).

### DIFF
--- a/client/src/sim.rs
+++ b/client/src/sim.rs
@@ -7,7 +7,7 @@ use tracing::{debug, error, trace};
 use crate::{net, prediction::PredictedMotion, Net};
 use common::{
     character_controller,
-    graph::{Graph, NodeId},
+    graph::NodeId,
     node::{populate_fresh_nodes, DualGraph},
     proto::{self, Character, CharacterInput, CharacterState, Command, Component, Position},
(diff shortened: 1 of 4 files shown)


client/src/sim.rs
```diff
@@ -85,71 +85,58 @@ impl Sim {
...
-            }
-            Hello(msg) => {
-                self.params = Some(Parameters {
-                    character_id: msg.character,
-                    cfg: msg.sim_config,
-                });
-                // Populate the root node
-                populate_fresh_nodes(&mut self.graph);
+            ConnectionLost(_) | Hello(_) => {
+                panic!("Should not get here, case handled elsewhere");
```
Ralith, line 139: style nit: `unreachable!(...)`

Ralith (approved):
Nice structural improvement. I concur that this will dovetail gracefully with an eventual main menu/connection GUI.

PR #315, opened 2023-10-15
Replace generic Graph with concrete implemetation and include ChunkLayout

patowen:
Currently, outside of tests, the only implementation of `Graph` is `DualGraph`. This used to not be true, as the server used a `Graph<()>` instead, but to allow server-side collision checking, this has changed. In the spirit of only using generics when necessary, this PR gets rid of the `Graph`/`DualGraph` distinction.

As part of this change, this PR also attaches `ChunkLayout` data to `Graph`, as it arguably makes sense for this data to live together, but it was harder to do something like this cleanly when `Graph` was generic. This can enable additional helper methods, such as one that takes a `ChunkId` and a set of coordinates and alters corresponding voxel, taking care of adjacent chunk margins and surface invalidation (without needing to repeatedly pass in the dimension as a parameter).

This PR doesn't refactor most code to use the `ChunkLayout` in `Graph` instead of the `chunk_size` in `SimConfig`, as there isn't a clear benefit. However, to ensure there is at least one usage of this new functionality, this PR alters the collision checking code to use the `ChunkLayout` in `Graph` instead of taking it in as a parameter.

### DIFF
--- a/client/src/prediction.rs
+++ b/client/src/prediction.rs
@@ -2,7 +2,7 @@ use std::collections::VecDeque;
 
 use common::{
     character_controller,
-    node::DualGraph,
+    graph::Graph,
     proto::{CharacterInput, Position},
     SimConfig,
 };
@@ -35,7 +35,7 @@ impl PredictedMotion {
 
     /// Update for input about to be sent to the server, returning the generation it should be
     /// tagged with
-    pub fn push(&mut self, cfg: &SimConfig, graph: &DualGraph, input: &CharacterInput) -> u16 {
+    pub fn push(&mut self, cfg: &SimConfig, graph: &Graph, input: &CharacterInput) -> u16 {
         character_controller::run_character_step(
             cfg,
             graph,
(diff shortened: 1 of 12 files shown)


Ralith:
My one concern here would be whether making the complexity of `Node` intrinsic to `Graph` complicates testing, but it looks like it hasn't, so that seems fine. Fewer types to think about is nice.

patowen:
> My one concern here would be whether making the complexity of `Node` intrinsic to `Graph` complicates testing, but it looks like it hasn't, so that seems fine. Fewer types to think about is nice.

This all makes sense to me. I believe the fact that `Graph` stores `Option<Node>` types instead of `Node` types means that many tests don't run into this issue at all.

Eventually, we may want to add back an abstraction so that, for instance, server code doesn't need to store `surface` as part of the `Chunk` structure, but when we decide to do that refactor, it seems likely that the abstraction would be done differently.

Ralith (approved):

PR #320, opened 2023-10-28
Update quinn and rustls

patowen:
Update rustls from 0.20.6 to 0.21.7.
Update quinn from 0.8.3 to 0.10.2.

Most changes are simple compatibility fixes, mostly involving turning previously-quinn-provided streams into loops with task spawns.

One small but potentially important change is that the `ClientEvent::Lost` message will always be sent when an error occurs, even if the connection was already closed from some other logic and the client was already cleaned up. To be robust to this kind of redundancy and potential race conditions, the `on_client_event` method is set up to skip messages received from clients that were already cleaned up.

### DIFF
--- a/client/src/net.rs
+++ b/client/src/net.rs
@@ -1,7 +1,6 @@
 use std::{sync::Arc, thread};
 
 use anyhow::{anyhow, Error, Result};
-use futures_util::{StreamExt, TryStreamExt};
 use tokio::sync::mpsc;
 
 use common::{codec, proto};
@@ -63,16 +62,12 @@ async fn inner(
     endpoint: quinn::Endpoint,
 ) -> Result<()> {
     let server = cfg.server.unwrap();
-    let quinn::NewConnection {
-        connection,
-        mut uni_streams,
-        ..
-    } = endpoint.connect(server, "localhost").unwrap().await?;
+    let connection = endpoint.connect(server, "localhost").unwrap().await?;
 
     // Open the first stream for our hello message
     let clienthello_stream = connection.open_uni().await?;
     // Start sending commands asynchronously
-    tokio::spawn(handle_outgoing(outgoing, connection));
+    tokio::spawn(handle_outgoing(outgoing, connection.clone()));
     // Actually send the hello message
     codec::send_whole(
         clienthello_stream,
@@ -82,9 +77,9 @@ async fn inner(
     )
     .await?;
 
-    let mut ordered = uni_streams.next().await.unwrap()?;
+    let mut ordered = connection.accept_uni().await?;
     // Handle unordered messages
-    tokio::spawn(handle_unordered(incoming.clone(), uni_streams));
+    tokio::spawn(handle_unordered(incoming.clone(), connection));
 
     // Receive the server's hello message
     let hello = codec::recv::<proto::ServerHello>(&mut ordered)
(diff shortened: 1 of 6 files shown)


client/src/net.rs
```diff
@@ -118,21 +113,31 @@ async fn handle_outgoing(
...
-        // `msgs` and here.
-        let _ = incoming.send(Message::StateDelta(msg));
+    loop {
+        let stream = connection.accept_uni().await;
+        let stream = match stream {
+            Err(quinn::ConnectionError::ApplicationClosed { .. }) => {
+                return Ok(());
+            }
+            Err(e) => {
+                return Err(e.into());
```
Ralith, line 125: It doesn't look like this return value goes anywhere. Might as well modify the function to not return anything, and consider sending an error on `incoming` here. Alternatively, handle errors at the callsite with the `tokio::spawn(async move { if let Err(e) = ... { ... });` pattern.

Ralith:
> The current naive tokio::spawn-based handling of drive_recv allows client messages to appear for a given client after ClientEvent::Lost

One strategy to avoid this might be to only generate `Lost` messages after all tasks associated with a client have ended. Alternatively, we could use some shared state (perhaps `quinn::Connection::close_reason()`) to judge that we should drop messages rather than sending them on.

> certain convenience methods like buffer_unordered are no longer available. This has resulted in an increased number of tokio::spawn calls, and I haven't thought them out well.

This is the correct refactoring. Concurrency limits are built-in to quinn now.

patowen:
> One strategy to avoid this might be to only generate `Lost` messages after all tasks associated with a client have ended. Alternatively, we could use some shared state (perhaps `quinn::Connection::close_reason()`) to judge that we should drop messages rather than sending them on.

I like the idea of delaying the `ClientEvent::Lost` message, as the more data the server receives from the client, the more graceful their exit can be. Now that I know this approach I'm taking doesn't seem too far off, I'll give this another shot.

> This is the correct refactoring. Concurrency limits are built-in to quinn now.

Oh, nice! Good to know there are fewer constants we have to configure.

patowen:
Thinking about the Lost-messages problem more, the client can be removed for any reason, and I hope I can find a solution to this problem that doesn't require being too careful in the future. Fortunately, there's a feature of `slotmap` that I was unaware of:

> The keys returned by `slotmap` are versioned. This means that once a key is removed, it stays removed, even if the physical storage inside the slotmap is reused for new elements. The key is a permanently unique<sup>*</sup> reference to the inserted value.

Therefore, I believe the main solution here is simply to have `on_client_event` ignore messages for non-existing client IDs. The asterisk I believe is just due to numerical overflow, but that shouldn't be an issue here.

There's still the question of how to gracefully handle client exits, but that feature doesn't exist in Hypermine yet in the master branch (clients just quit, and servers say "dropping slow client"), so I don't need to solve that in this PR, since its purpose is to just update quinn.

Anyway, I should be able to update this PR with the help you gave, hopefully enough to be a non-draft.

server/src/lib.rs
```diff
@@ -186,25 +209,19 @@ impl Server {
...
-            }
-        };
         let id = self.clients.insert(Client::new(connection.clone()));
         info!(id = ?id.0, address = %connection.remote_address(), "connection established");
         tokio::spawn(async move {
-            if let Err(e) = drive_recv(id, uni_streams, &mut send).await {
+            if let Err(e) = drive_recv(id, connection, &mut send).await {
+                // drive_recv returns an error when any connection-terminating issue occurs, so we
+                // send a Lost message to ensure the client is cleaned up. Note that this message may
+                // be redundant, as drive_recv also sends such messages.
```
Ralith, line 221: > drive_recv also sends such messages

Does it? Where?

patowen: Whoops. It used to in an earlier WIP revision, but I switched to just closing the connection instead. I'll update this comment.

client/src/net.rs
```diff
@@ -116,23 +111,29 @@ async fn handle_outgoing(
...
-        let _ = incoming.send(Message::StateDelta(msg));
+async fn handle_unordered(incoming: mpsc::UnboundedSender<Message>, connection: quinn::Connection) {
+    loop {
+        let stream = match connection.accept_uni().await {
+            Err(_) => {
+                // accept_uni should only fail if the connection is closed, which is already handled elsewhere.
+                return;
+            }
+            Ok(stream) => stream,
+        };
```
Ralith, line 122: This would be much more concise with let-else.

server/src/lib.rs
```diff
@@ -214,26 +231,35 @@ const MAX_CLIENT_MSG_SIZE: usize = 1 << 16;
...
+        // We spawn a separate task to allow messages to be processed in a different order from when they were
+        // initiated.
+        let connection = connection.clone();
+        tokio::spawn(async move {
+            match codec::recv_whole::<proto::Command>(MAX_CLIENT_MSG_SIZE, stream).await {
+                Err(e) => {
+                    // This error can occur if the client sends a badly-formatted command. In this case,
+                    // we want to drop the client. We close the connection, which will cause `drive_recv` to
+                    // return eventually.
+                    tracing::error!("Error when parsing unordered stream from server: {e}");
```
Ralith, line 254: Should this read "from client"?

Ralith (approved):

server/src/lib.rs
patowen: It should. Thanks! I copied this from the version I wrote in `net.rs`.

PR #325, opened 2023-11-18
Update winit to 0.29.3

patowen:
winit now defaults to raw-window-handle 0.6, but ash-window is not yet compatible with this version of raw-window-handle. Fortunately, the `rwh_05` feature allows the use of the older version of raw-window-handle.

A few notable changes:
- winit did an overhaul of keyboard input, so the use of `virtual_keycode` had to be replaced with `physical_key` or `logical_key`. I guessed that `physical_key` was better.
- `Event::MainEventsCleared` was removed from winit, so I needed to switch to `AboutToWait`. However, I used https://github.com/rust-windowing/winit/blob/7bed5eecfdcbde16e5619fd137f0229e8e7e8ed4/examples/request_redraw.rs as an example, which had me make the additional change of putting all the frame logic in `WindowEvent::RedrawRequested` and using `self.window.request_redraw` in `Event::AboutToWait`. I'm not certain this is what we want for Hypermine, but I didn't notice any issues on my end when testing it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,11 +9,15 @@ quinn = "0.10.2"
 toml = { version = "0.8.0", default-features = false, features = ["parse"] }
 
 [profile.dev]
+lto = true
 opt-level = 1
 debug-assertions = true
 
 [profile.dev.package."*"]
 opt-level = 2
+
+[profile.release]
+lto = true
 
 [profile.release.build-override]
 opt-level = 0
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -14,9 +14,9 @@ server = { path = "../server" }
 tracing = "0.1.10"
 ash = { version = "0.37.1", features = ["loaded"] }
 lahar = { git = "https://github.com/Ralith/lahar", rev = "88abd75e41d04c3a4199d95f581cb135f5962844" }
-winit = "0.28.1"
+winit = { version = "0.29.3", features = ["rwh_05"] }
 ash-window = "0.12.0"
-raw-window-handle = "0.5.0"
+raw-window-handle = { version = "0.5.0", features = ["std"] }
 directories = "5.0.1"
 vk-shader-macros = "0.2.5"
 nalgebra = { workspace = true }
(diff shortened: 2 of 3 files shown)


patowen:
Weird, Linux linking failed. These seem like the most relevant error logs:
`error: linking with `cc` failed: exit status: 1`
`undefined reference to `memfd_create'`
```
  = note: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)
```
I wonder whether this is a bug in a third party library or a mistake in how I set up my dependencies. I haven't tested this PR on my Linux partition, so I'm not sure whether it usually works, and there's just a GitHub-related quirk. This may take a while to debug.

patowen:
I'm going to turn link-time-optimization to see if this fixes anything. I don't really like this solution, though, as it makes compiling slower. If this does work, though, perhaps there's a way to turn on link-time-optimization specifically just for `package-linux`.

Ralith:
Looks like `nix` is gratuitously broken, yeah; that's annoying. No harm in sticking with old winit for the time being, I think.

patowen:
That probably makes the most sense. Besides the main issue at hand, winit seems pretty simple to update, so I'll go ahead and close this PR for now.

patowen:
It's been over three months since I closed this PR. The underlying bug _might_ now be fixed, even though the linked issue has not been touched.

Bevy has been upgraded to 0.13, and that version of Bevy now uses winit 0.29, and given how popular that library is, I would expect that winit 0.29 should be ready now.

I'm reopening this PR so that CI runs again. Hopefully, it will work this time.

patowen:
Huh, even though force-pushes are allowed while a PR is open, re-opening after a force push is not. I'm not sure why that restriction exists. It seems rather arbitrary. I'll open up a new PR.

PR #331, opened 2023-12-17
Save and load all modified chunks

patowen:
Whenever a block is updated, ensure that an updated version of the relevant node is written to the save file. When a world is initialized on the server-side, all modified chunks are loaded from the save file and stored in memory, to be retrieved whenever the relevant node is to be added to the graph.

This is a temporary implementation, as it does not properly scale for larger worlds, where we may not want to store all modified chunks in RAM.

Note that while this implements loading save data, it only loads voxels. Entity locations, such as the player's location, are ignored.

### DIFF
--- a/save/src/lib.rs
+++ b/save/src/lib.rs
@@ -139,6 +139,16 @@ impl Reader<'_> {
             .map_err(GetError::DecompressionFailed)?;
         Ok(Some(Character::decode(&*self.accum)?))
     }
+
+    /// Temporary function to load all voxel-related save data at once.
+    /// TODO: Replace this implementation with a streaming implementation
+    /// that does not require loading everything at once
+    pub fn get_all_voxel_node_ids(&self) -> Result<Vec<u128>, GetError> {
+        self.voxel_nodes
+            .iter()?
+            .map(|n| Ok(n.map_err(GetError::from)?.0.value()))
+            .collect()
+    }
 }
 
 fn decompress(
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -208,7 +208,7 @@ impl Sim {
 
             spawns
                 .modified_chunks
-                .push((chunk_id, voxels.to_serializable(self.cfg.chunk_size)));
+                .push((chunk_id, voxels.serialize(self.cfg.chunk_size)));
         }
         spawns
     }
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -1,6 +1,9 @@
 use std::sync::Arc;
 
-use common::proto::BlockUpdate;
+use anyhow::Context;
+use common::dodeca::Vertex;
+use common::node::VoxelData;
+use common::proto::{BlockUpdate, SerializedVoxelData};
 use common::{node::ChunkId, GraphEntities};
 use fxhash::{FxHashMap, FxHashSet};
 use hecs::Entity;
@@ -32,29 +35,40 @@ pub struct Sim {
     entity_ids: FxHashMap<EntityId, Entity>,
     world: hecs::World,
     graph: Graph,
+    /// Voxel data that has been fetched from a savefile but not yet introduced to the graph
+    preloaded_voxel_data: FxHashMap<ChunkId, VoxelData>,
     spawns: Vec<Entity>,
     despawns: Vec<EntityId>,
     graph_entities: GraphEntities,
+    /// All nodes that have entity-related information yet to be saved
     dirty_nodes: FxHashSet<NodeId>,
+    /// All nodes that have voxel-related information yet to be saved
+    dirty_voxel_nodes: FxHashSet<NodeId>,
+    /// All chunks that have ever had any block updates applied to them and can no longer be regenerated with worldgen
     modified_chunks: FxHashSet<ChunkId>,
 }
 
 impl Sim {
-    pub fn new(cfg: Arc<SimConfig>) -> Self {
+    pub fn new(cfg: Arc<SimConfig>, save: &save::Save) -> Self {
         let mut result = Self {
             rng: SmallRng::from_entropy(),
             step: 0,
             entity_ids: FxHashMap::default(),
             world: hecs::World::new(),
             graph: Graph::new(cfg.chunk_size),
+            preloaded_voxel_data: FxHashMap::default(),
             spawns: Vec::new(),
             despawns: Vec::new(),
             graph_entities: GraphEntities::new(),
             dirty_nodes: FxHashSet::default(),
+            dirty_voxel_nodes: FxHashSet::default(),
             modified_chunks: FxHashSet::default(),
             cfg,
         };
 
+        result
+            .load_all_voxels(save)
+            .expect("save file must be of a valid format");
         ensure_nearby(
             &mut result.graph,
             &Position::origin(),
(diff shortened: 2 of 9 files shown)


server/src/sim.rs
```diff
@@ -257,16 +322,28 @@ impl Sim {
...
+        for fresh_node in fresh_nodes.iter().copied() {
+            for vertex in Vertex::iter() {
+                let chunk = ChunkId::new(fresh_node, vertex);
+                if let Some(voxel_data) = self.pending_modified_chunks.remove(&chunk) {
+                    modified_chunks.push((chunk, voxel_data.to_serializable(self.cfg.chunk_size)));
+                    self.modified_chunks.insert(chunk);
+                    self.graph.populate_chunk(chunk, voxel_data, true)
+                }
+            }
+        }
```
Ralith, line 336: There's too many different things named "modified_chunks" or a variation of it here. We need more distinctive names and clear documentation of what each one represents.

patowen: Good point. I definitely agree, especially since both client's `sim` and server's `sim` have a field called `pending_modified_chunks`, but one uses a `Vec<BlockUpdate>` while the other uses a `VoxelData`.

It's also weird that this is the name for both lists of chunks that were modified and maps that say what the modifications were, since those two things are conceptually quite different.

patowen: I think I found reasonable names for all these variables.
- The two versions of `pending_modified_chunks` are now `preloaded_block_updates` and `preloaded_voxel_data`
- The three versions of `modified_chunks` are now `voxel_data`, `modified_chunks`, and `fresh_voxel_data`

server/src/sim.rs
```diff
@@ -127,6 +166,30 @@ impl Sim {
...
+            }
+            let mut serialized_voxels = Vec::new();
+            let Chunk::Populated { ref voxels, .. } = node_data.chunks[vertex] else {
+                panic!("Unknown chunk listed as modified");
+            };
+            postcard_helpers::serialize(
+                &voxels.to_serializable(self.cfg.chunk_size),
+                &mut serialized_voxels,
+            )
+            .unwrap();
```
Ralith, line 184: This encoding is incorrect. The .proto says:

```
    // Dense 3D array of 16-bit material tags for all voxels in this chunk
    bytes voxels = 2;
```

The postcard output for "struct of `Vec` of `Material`" is not consistent with that. We should construct a `Vec<u8>` for use in `Chunk` directly from `&[Material]` by something like `materials.flat_map(|x| x.to_le_bytes()).collect()`, which happens to be very concise anyway.

save/src/lib.rs
```diff
@@ -139,6 +139,16 @@ impl Reader<'_> {
             .map_err(GetError::DecompressionFailed)?;
         Ok(Some(Character::decode(&*self.accum)?))
     }
+
+    /// Temporary function to load all voxel-related save data at once.
+    /// TODO: Replace this implementation with a streaming implementation
+    /// that does not require loading everything at once
+    pub fn get_all_voxel_node_ids(&mut self) -> Result<Vec<u128>, GetError> {
```
Ralith, line 146: nit:
```suggestion
    pub fn get_all_voxel_node_ids(&self) -> Result<Vec<u128>, GetError> {
```

server/src/sim.rs
```diff
@@ -85,16 +97,43 @@ impl Sim {
...
 
+    fn load_all_voxels(&mut self, save: &save::Save) -> anyhow::Result<()> {
+        let read_guard = save.read()?;
+        let mut read = read_guard.get()?;
+        for node_hash in read.get_all_voxel_node_ids()? {
+            let Some(voxel_node) = read.get_voxel_node(node_hash)? else {
+                continue;
+            };
+            for chunk in voxel_node.chunks.iter() {
+                let voxels: SerializableVoxelData = postcard::from_bytes(&chunk.voxels)?;
```
Ralith, line 123: Incorrect encoding.

server/src/sim.rs
```diff
@@ -31,29 +34,38 @@ pub struct Sim {
     entity_ids: FxHashMap<EntityId, Entity>,
     world: hecs::World,
     graph: Graph,
+    /// Voxel data that has been fetched from a savefile but not yet introduced to the graph
+    preloaded_voxel_data: FxHashMap<ChunkId, VoxelData>,
     spawns: Vec<Entity>,
     despawns: Vec<EntityId>,
     graph_entities: GraphEntities,
     dirty_nodes: FxHashSet<NodeId>,
+    dirty_voxel_nodes: FxHashSet<NodeId>,
```
Ralith, line 43: This field should be documented.

Ralith (changes requested):

server/src/sim.rs
patowen: Thanks! I forgot that the intended output was already documented.

patowen: One issue I'm noticing is that `x` is of type `Material`, not `u16`, so the manual serialization also requires an `as u16`. This is fine for serialization, but for deserialization, we need to turn the `u16` back into a `Material` enum, which Rust doesn't support on its own. I can think of a few options:
- Import a new library to do this.
- Remove the `Material` enum entirely in favor of something more flexible (store a `u16` ourselves instead of using `repr`)
- Create a huge `match` statement to do the conversion (which I'm currently doing in my draft code)
- Use `unsafe` to do the conversion
- Use serde somehow to do the conversion.

I'm currently in favor of the second option, although I don't know whether to block this PR behind such a refactor.

save/src/lib.rs
patowen: There are definitely legitimate use cases for using a `&mut self` in situations like this, which I hadn't thought about, so I had expected Clippy to catch things like this. Good catch!

server/src/sim.rs
Ralith: A big `match`, probably in a `try_from`, is the idiomatic way to do this. LLVM is known to optimize those out reliably. You can use a simple macro to generate them if it gets unwieldly.

patowen: A crate like `int-enum` or `enumn` could prove useful here. I'll shy away from using a macro for now, but I'll use a unit test to enforce consistency on what would otherwise be a DRY-violation.

Ralith: `enumn` looks pleasant to me, but I don't mind handwritten either. DRY can be taken too far.

> I expect we'll want to switch away from an enum for other reasons (for instance, a dynamically-generated materials list with procedural textures and/or mod scripting support)

Oh, that's actually a really good point. Using a (newtyped) u16 would let us make materials totally data-driven, which is compelling for ease of maintenance and tinkering if nothing else. No need to pursue that in this PR though.

patowen: > DRY can be taken too far.

This is definitely something I've had a change of heart on, as I used to consider DRY to be much more important. I think repeated logic is entirely fine as long as there's something enforcing consistency between the two versions, as avoiding inconsistencies between the two (that would cause complexity if intentional and bugs if not) is one of the biggest reasons for DRY as far as I understand.

patowen: I decided to fix this by having the resulting `Vec<u8>` be present in `SerializableVoxelData`, which I consequently renamed to `SerializedVoxelData`. The main reason for this is that the only purpose of `SerializableVoxelData` is to be able to serialize it, and having different parts of the serialization in different parts of the code seems silly if both parts are done manually.

I decided to put this change in a separate commit, so this PR now has two commits.

server/src/sim.rs
Ralith: Oh, I guess that was pretty obvious in retrospect. Oh well, no harm!

Ralith (approved):

server/src/sim.rs
patowen: I still agree with your original suggestion, since one could imagine encountering this code before knowing about how voxel data and entity data is separated in save files and being confused.

patowen:
Thanks for reviewing/merging!

PR #340, opened 2024-01-09
Allow for number keys to select type of block to be placed

jhuang97:
The ten materials that can be selected can be edited by changing the field material_palette in client/src/sim.rs.

The change in common/src/character_controller/mod.rs was necessary for the server not to crash when running on my machine.

As of now, the number keys map to the following materials:
1: Wood Planks
2: Grass
3: Dirt
4: Sand
5: Snow
6: White Brick
7: Grey Brick
8: Basalt
9: Water
0: Lava

### DIFF
--- a/client/src/graphics/window.rs
+++ b/client/src/graphics/window.rs
@@ -256,6 +256,16 @@ impl Window {
                                 sim.toggle_no_clip();
                             }
                         }
+                        VirtualKeyCode::Key1 | VirtualKeyCode::Key2 | VirtualKeyCode::Key3
+                        | VirtualKeyCode::Key4 | VirtualKeyCode::Key5 | VirtualKeyCode::Key6
+                        | VirtualKeyCode::Key7 | VirtualKeyCode::Key8 | VirtualKeyCode::Key9
+                        | VirtualKeyCode::Key0 => {
+                            if state == ElementState::Pressed {
+                                if let Some(sim) = self.sim.as_mut() {
+                                    sim.select_material(number_key_to_index(key));
+                                }
+                            }
+                        }
                         VirtualKeyCode::Escape => {
                             let _ = self.window.set_cursor_grab(CursorGrabMode::None);
                             self.window.set_cursor_visible(true);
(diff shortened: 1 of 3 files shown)


common/src/character_controller/mod.rs
```diff
@@ -34,7 +34,7 @@ pub fn run_character_step(
             graph,
             radius: sim_config.character.character_radius,
         },
-        up: graph.get_relative_up(position).unwrap(),
```
Ralith, line 37: This comment isn't meaningful. If there's a bug here, please open a separate PR (or issue) and keep this PR focused.

client/src/graphics/window.rs
```diff
@@ -256,6 +256,16 @@ impl Window {
...
+                        VirtualKeyCode::Key1 | VirtualKeyCode::Key2 | VirtualKeyCode::Key3
+                        | VirtualKeyCode::Key4 | VirtualKeyCode::Key5 | VirtualKeyCode::Key6
+                        | VirtualKeyCode::Key7 | VirtualKeyCode::Key8 | VirtualKeyCode::Key9
+                        | VirtualKeyCode::Key0 => {
+                            if state == ElementState::Pressed {
+                                if let Some(sim) = self.sim.as_mut() {
+                                    sim.select_material(number_key_to_index(key));
+                                }
+                            }
+                        }
```
Ralith, line 268: nit: We could avoid listing every key twice by moving this logic into the catch-all `_ => {}` case below, modifying `number_key_to_index` to return `Option`, and calling it there.

client/src/sim.rs
```diff
@@ -58,6 +58,10 @@ pub struct Sim {
     place_block_pressed: bool,
     /// Whether the break-block button has been pressed since the last step
     break_block_pressed: bool,
+
+    selected_material: Material,
+    material_palette: [Material; 10],
```
Ralith, line 63: nit: this might as well be a `const`, since we're not likely to mutate it any time soon.

client/src/sim.rs
```diff
@@ -138,6 +146,14 @@ impl Sim {
         self.place_block_pressed = true;
     }
 
+    pub fn select_material(&mut self, idx: usize) {
+        self.selected_material = if idx < self.material_palette.len() {
+            self.material_palette[idx]
+        } else {
+            self.material_palette[0]
+        };
```
Ralith, line 154: This could be more concise:
```suggestion
        self.selected_material = self.material_palette.get(idx).unwrap_or(self.material_palette[0]);
```

Ralith (approved):
Thanks, this looks like a nice straightforward improvement!

Ralith (changes requested):

jhuang97:
thanks for all the comments! I made changes for all of them and also ran cargo fmt locally.

Ralith (approved):
Thanks!

PR #345, opened 2024-01-21
Move camera to the top of the player's collider

Ralith:
~This moves the viewpoint before applying orientation, so looking down is equivalent to face-planting, which isn't ideal.~ Adjusting the viewpoint by `Graph::get_relative_up` after rotation so the camera rotates in-place makes more sense, but suffers from sharp discontinuities when crossing between nodes, presumably due to relying on the discrete `NodeState::up_direction`. We should come up with a properly smooth notion of "up".

### DIFF
--- a/client/src/sim.rs
+++ b/client/src/sim.rs
@@ -437,7 +437,15 @@ impl Sim {
     }
 
     pub fn view(&self) -> Position {
-        self.local_character_controller.oriented_position()
+        let mut pos = self.local_character_controller.position();
+        let up = self.graph.get_relative_up(&pos).unwrap();
+        pos.local *= common::math::translate_along(
+            &(up.as_ref() * (self.cfg.character.character_radius - 1e-3)),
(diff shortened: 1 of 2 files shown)


client/src/sim.rs
```diff
@@ -437,7 +437,15 @@ impl Sim {
     }
 
     pub fn view(&self) -> Position {
-        self.local_character_controller.oriented_position()
+        let mut pos = self.local_character_controller.position();
```
patowen, line 440: I believe you should be able to simplify things by using `oriented_position()` here, which would remove the need to use `* self.local_character_controller.orientation().to_homogeneous()`.

Because `oriented_position()` is used as the argument to `get_relative_up`, the coordinate system should be correct, making up for the change in the order of multiplication.

Ralith: Right you are. https://github.com/Ralith/hypermine/pull/346.

PR #347, opened 2024-01-25
Use `c_char` instead of `i8` where necessary

wackbyte:
Aarch64 Linux has [`c_char` as `u8` rather than `i8`](https://doc.rust-lang.org/1.75.0/src/core/ffi/mod.rs.html#99) as the code expects. This fixes the build there.


Ralith (approved):
Thanks!

PR #349, opened 2024-01-27
Add a cursor to make placing/breaking blocks easier

patowen:
This PR hijacks the fog shader to also add a cursor in the center of the viewport to allow players to target where they place and break blocks.

This introduces an `if` statement to the fragment shader, and branching in shaders is generally bad for performance in my understanding, but given its simplicity, I believe that the shader compiler should be smart enough to optimize this in a way that acts as though such branching basically does not exist.

### DIFF
--- a/client/shaders/fog.frag
+++ b/client/shaders/fog.frag
@@ -16,6 +16,12 @@ void main() {
     float view_length = length(view_pos);
     // Convert to true hyperbolic distance, taking care to respect atanh's domain
     float dist = view_length >= 1.0 ? INFINITY : atanh(view_length);
-    // Exponential^k fog
-    fog = vec4(0.5, 0.65, 0.9, exp(-pow(dist * fog_density, 5)));
+    if (dot(scaled_view_pos.xy, scaled_view_pos.xy) < 0.0001) {
+        // Temporary code to add a cursor in the center of the window for placing/breaking blocks
+        // TODO: Replace with a UI element when UI exists
+        fog = vec4(0.0, 0.0, 0.0, 0.0);
+    } else {
+        // Exponential^k fog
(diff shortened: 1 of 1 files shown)


Ralith:
> branching in shaders is generally bad for performance in my understanding

This hasn't really been true for a while. So long as register allocation costs aren't wildly unbalanced between branches, you can be pretty free with them.

client/shaders/fog.frag
```diff
@@ -16,6 +16,12 @@ void main() {
...
     float dist = view_length >= 1.0 ? INFINITY : atanh(view_length);
-    // Exponential^k fog
-    fog = vec4(0.5, 0.65, 0.9, exp(-pow(dist * fog_density, 5)));
+    if (dot(scaled_view_pos.xy, scaled_view_pos.xy) < 0.0001) {
+        // Temporary code to add a cursor in the center of the window for placing/breaking blocks
+        // TODO: Replace with a UI element when UI exists
+        fog = vec4(0.0, 0.0, 0.0, 0.0);
+    } else {
+        // Exponential^k fog
+        fog = vec4(0.5, 0.65, 0.9, dist < 0.02 ? 0.5 : exp(-pow(dist * fog_density, 5)));
```
Ralith, line 25: This seems to have changed the fog formula, which is unrelated to introducing a cursor. What's the intention behind that?

patowen: Good catch! That was unintentional, caused by me assuming that the cursor was the only change I made to fog in the collision demo.

I believe its purpose was to highlight parts of the terrain closer than 0.02 units away from the camera as a way to verify that collision detection is correct.

Ralith (approved):

PR #351, opened 2024-01-30
Added mozilla overlay and mold linker to shell.nix

### LINKED ISSUE #348: shell.nix uses `rustChannels` which is not available on nix by default
https://github.com/Ralith/hypermine/blob/088de1f9a92b72ef1d182a36a6cecfcdf0c4d9a4/shell.nix#L5

this results in errors such as this:
image

juh9870:
Fixes #348

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -1,7 +1,7 @@
 let
   moz_overlay = import (builtins.fetchTarball
     "https://github.com/mozilla/nixpkgs-mozilla/archive/9b11a87c0cc54e308fa83aac5b4ee1816d5418a2.tar.gz");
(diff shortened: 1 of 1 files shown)


shell.nix
```diff
@@ -1,8 +1,19 @@
-with import <nixpkgs> { };
+let
+  moz_overlay = import (builtins.fetchTarball
+    "https://github.com/mozilla/nixpkgs-mozilla/archive/9b11a87c0cc54e308fa83aac5b4ee1816d5418a2.tar.gz");
+  nixpkgs = import <nixpkgs> { overlays = [ moz_overlay ]; }; # 23.11
```
Ralith, line 4: This comment isn't very clear, and I'm not sure it would add much even if clarified.

Ralith (approved):

PR #355, opened 2024-02-19
Use a simulation rate of 30 steps per second instead of 10

patowen:
A faster tick rate reduces the lag between mouse clicks and the modification of terrain, which would be a good quality of life improvement, especially while prediction is not implemented for terrain modification.


Ralith (approved):
Could still go higher; even 30Hz is pretty low for a real-time game, I think!

patowen:
Sounds good to me. I bumped it up to 30. Going up to 60 might be nice, but it does increase required bandwidth etc. so I'll only bump it up further if you suggest such a change explicitly.

Ralith:
I don't foresee bandwidth being a problem, but this seems like a happy medium.

PR #359, opened 2024-02-24
Stop dependabot from updating packages that likely cannot be updated

patowen:
This PR stops dependabot from checking for updates of `raw-window-handle`, `rustls`, and `rustls-pemfile` because updates to these crates have to be done in tandem with upgrades to other crates (namely `ash-window` and `quinn`).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,13 +6,8 @@ updates:
     interval: daily
   open-pull-requests-limit: 10
   ignore:
-  - dependency-name: whoami
-    versions:
-    - 1.1.0
-    - 1.1.1
+  # Ignore raw-window-handle because it's tied to ash-window
+  - dependency-name: raw-window-handle
+  # Ignore rustls and rustls-pemfile because they're tied to quinn
   - dependency-name: rustls
-    versions:
-    - 0.19.0
-  - dependency-name: directories
-    versions:
-    - 3.0.1
+  - dependency-name: rustls-pemfile


patowen:
@Ralith, I don't know if there's a good way to ensure that this PR will work as intended. I have a few concerns, and the answer might simply be to merge this PR and see what happens, but I'd be interested in your thoughts here:

- If one of these crates has a CVE-related security update, will Dependabot let us know? Since `cargo.lock` isn't checked in, documentation suggests that this isn't the case, but I decided to try to at least leave the option open by only blocking major version updates.
- Is Dependabot aware that the format `0.major.minor` is often used, or will it treat 0.11 to 0.12 as a minor version upgrade?

Also, how much should we worry about security (at least with automatic detection)? If one of our dependencies misuses `unsafe` in a way that potentially allows save files or netcode to be exploited, that could make Hypermine dangerous when sharing save files or playing on whatever the equivalent of public servers are for niche things like Hypermine. However, at the same time, the attack surface seems pretty low, and a vulnerability that would make Hypermine dangerous is maybe something we would hear about anyway regardless of whether Dependabot would let us know.

If we don't need to worry about security here, it might be nicer just to ignore these dependencies instead of filtering to major versions, since minor and patch versions are usually automatic anyway.

Ralith:
Per docs, I don't think minor version updates exist unless we commit Cargo.lock, and if they did they'd probably be spammier than we have the resources to support.

patowen:
That's a fair point. I'll go ahead and just include a list of dependencies to ignore instead of trying to do anything fancier.

Ralith (approved):

Issue #102, opened 2020-05-26
Improve numerical stability of ground plane distance estimation


Ralith:
At sufficiently high elevations with respect to the ground plane, glitchy floating islands appear. This is likely due to exponential growth of the magnitude of the normal vector encoding the distance to the ground plane. We could mitigate the representation problem by throwing additional precision at the problem in the form of an explicit exponent: rather than the vector `v`, store the pair `(v/(2^c), c)`. Further investigation (@MagmaMcFry?) is necessary to determine the details of computing distance/direction of the plane from this representation without running into the same issue again.

MagmaMcFry:
Suppose you have a pair `(x,c)`, where x is a vec4 and c is a *nonnegative* number, representing the unit normal x*2^c of a plane H, and we want to compute the distance d between H and a point p. We know that sinh(d) = mip(x*2^c, p) = m*2^c, where m := mip(x,p).

Case 0: m is zero. Then d = 0.

Case 1: m is positive. Now log2(sinh(d)) = c+log2(m) =: k.
Compute k first. If k is relatively small, then you can just compute d = asinh(2^k) directly. If k is relatively large (at least 26), then up to double precision, sinh(d) = exp(d)/2, so log2(sinh(d)) = log2(exp(d)/2) = d*log2(e) - 1 = k, so you can compute d = (k+1)*ln(2).

Case 2: m is negative. In this case, negate m, apply the computation in case 1, then negate d.

The only possibility for numerical instability is when p is very far away from the current frame of reference (and then only when p appears to be very close to H in the local Klein model). We won't run into this case as long as p is close to the current node.

MagmaMcFry:
As a unit test for this system, I suggest starting with the extended normal `(x=(1,0,0,0), c=0)` and repeatedly boosting it away from the origin using a boost by distance 10 in negative X direction. After the `n`th boost, the distance between the normal and the point `p = (0,0,0,1)` should be `10*n`. Repeat this 200 times to make sure that it works even if unextended normals would overflow double-precision floats.

patowen:
Currently, I believe the main source of this numerical precision issue is in `Plane`'s [multiplication implementation](https://github.com/Ralith/hypermine/blob/6d061accf30d6e8c4de623aa3ff017f5368651d0/common/src/plane.rs#L49-L56). It runs a `lorentz_normalize`, which is numerically unstable when the plane is far from the origin.

[...]

PR #360, opened 2024-02-25
Default to f32 instead of f64, removing unnecessary casts

patowen:
# Purpose

A common issue I have encountered when trying to implement various features was the fact that much of Hypermine's code relies on `f32`, but certain common functions, especially in `dodeca.cs`, only returned `f64`, requiring a lot of redundant casting.

# Changes

This PR changes dodeca to return `f32` types by default, with alternative methods with `_f64` appended to their name to allow them to return `f64` instead.

To take advantage of these changes, some functions, such as the ones in `traversal` were altered to take in an `f32` instead of an `f64`, ensuring that all math is done with `f32`. As a side note, this highlighted an oversight in `ensure_nearby` and `nearby_nodes`, as a depth-search search was used instead of the superior breadth-first search, causing unnecessary numerical instability (which resulted in NaNs after the switch to `f32`). This PR also fixes that oversight.

This PR still uses `f64` for `Plane` to avoid making #102 worse, and it uses `f64` when interacting with `na::RealField` because otherwise, casting types to `na::RealField` does not work. (See https://github.com/dimforge/simba/issues/54)

# Moving forward

This PR is incremental in nature, as it does not yet fully solve the `f32` vs `f64` disconnect. To do that, I can imagine two approaches, depending on what we want Hypermine's engine to be capable of, although there are almost certainly other ways:

Option 1 (Move entirely to `f32`): Stop accommodating `f64` entirely and use `f32` for everything, simplifying the code base. Part of this would involve editing `math.rs` to use `f32` instead of `na::RealField + Copy`, which could simplify things greatly. The main downside would be that some ad-hoc implementations would run into numerical issues more quickly. For instance, the issue related to #102 would be worsened until we fix it.

Option 2 (Embrace `f64` as an first-class option): Use traits more consistently to allow our core functions in `dodeca` and `math` to work with `f32` or `f64`. It would likely be a good idea for us to define our own trait implemented by `f32` and `f64` and stop using `na::RealField` (at least for things that depend on `dodeca`), since that would give us more control. For instance, it would allow us to write more generic `dodeca` methods that return `f32` or `f64` depending on context.


common/src/traversal.rs
```diff
@@ -42,25 +42,25 @@ pub fn ensure_nearby(graph: &mut Graph, start: &Position, distance: f64) {
...
 ) -> Vec<(NodeId, na::Matrix4<f32>)> {
     struct PendingNode {
         id: NodeId,
-        transform: na::Matrix4<f64>,
+        transform: na::Matrix4<f32>,
     }
 
     let mut result = Vec::new();
-    let mut pending = Vec::<PendingNode>::new();
+    let mut pending = VecDeque::<PendingNode>::new();
```
Ralith, line 53: Is this change relevant?

Ralith (approved):

common/src/traversal.rs
patowen: It is. When switching `traversal` to use `f32` instead of `f64`, I found out that it was using a depth-first search instead of a breadth-first search. This line of code is part of the change to switch it to a breadth-first search.

In this context, depth-first searches are more numerically unstable, since long chains of nodes cause numerical errors to propagate, and that became highly apparent when trying to use `f32` here (resulting in NaNs).

Ralith: If the specific strategy is important, consider explaining why with inline comments. In an ideal world this would also be a separate commit, preceding the changeover to f32.

patowen: That's a fair point. I'll be an idealist here and split the commits.

Ralith: I'd like to see a little more detail in the comment as to *why* breadth-first is more stable. In a tree, this wouldn't be the case, because there's exactly one path from the root to each node anyway. I'm guessing the difference is that this graph contains cycles, so depth first can take a (much) longer than minimal path to a node?

patowen: Got it. I'll flesh out the comment a bit. Your guess is correct.

patowen: I've updated the wording to the following:
> We do a breadth-first instead of a depth-first traversal here to ensure that we take the minimal path to each node. This greatly helps prevent error from accumulating due to hundreds of transformations being composed.

This comment appears in two functions: `ensure_nearby` and `nearby_nodes`, so I've biased towards shorter comments. However, I'm wondering if I should just mention this as a more general principle at the top of the file.

Please let me know if you have any concerns with this wording and if I should flesh it out further.

Ralith: LGTM, thanks! Any sort of mention that a minimal path isn't always taken otherwise gets the point across, IMO.

patowen: Awesome! I'll go ahead and turn on auto-merge (rebase).

PR #372, opened 2024-03-29
Update zstd-safe to 7.1

### LINKED ISSUE #370: Build is currently broken (unresolved import `zstd_sys::ZSTD_cParameter::ZSTD_c_experimentalParam6`)
Building Hypermine from scratch results in the following error:

```
error[E0432]: unresolved import `zstd_sys::ZSTD_cParameter::ZSTD_c_experimentalParam6`
   --> C:\Users\Patrick\.cargo\registry\src\index.crates.io-6f17d22bba15001f\zstd-safe-6.0.6\src\lib.rs:609:13
    |
609 |             ZSTD_c_experimentalParam6 as ZSTD_c_targetCBlockSize,
    |             -------------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |             |
    |             no `ZSTD_c_experimentalParam6` in `ZSTD_cParameter`
    |             help: a similar name exists in the module: `ZSTD_c_experimentalParam1`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `zstd-safe` (lib) due to 1 previous error
```

See https://github.com/gyscos/zstd-rs/issues/270 and https://github.com/gyscos/zstd-rs/issues/268 for the likely root cause.

This comes from the dependency we import in save/Cargo.toml:
```toml
zstd = { package = "zstd-safe", version = "6.0.4", features = ["std", "experimental"] }
```

### Current workaround

Add the following after the `zstd` dependency in save/cargo.toml:
```toml
zstd-sys = "=2.0.9"
```

Ralith:
Fixes #370.

### DIFF
--- a/save/src/lib.rs
+++ b/save/src/lib.rs
@@ -158,12 +158,12 @@ fn decompress(
 ) -> Result<(), &'static str> {
     dctx.init().map_err(zstd::get_error_name)?;
     let mut input = zstd::InBuffer::around(compressed);
-    let mut out = zstd::OutBuffer::around(out);
     let out_size = zstd::DCtx::out_size();
     loop {
-        if out.dst.len() + out_size > out.dst.capacity() {
-            out.dst.reserve(out_size);
+        if out.len() + out_size > out.capacity() {
+            out.reserve(out_size);
         }
+        let mut out = zstd::OutBuffer::around_pos(out, out.len());
         let n = dctx
             .decompress_stream(&mut out, &mut input)
             .map_err(zstd::get_error_name)?;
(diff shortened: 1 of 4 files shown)


save/src/lib.rs
```diff
@@ -158,12 +158,12 @@ fn decompress(
...
     let mut input = zstd::InBuffer::around(compressed);
-    let mut out = zstd::OutBuffer::around(out);
     let out_size = zstd::DCtx::out_size();
     loop {
-        if out.dst.len() + out_size > out.dst.capacity() {
-            out.dst.reserve(out_size);
+        if out.len() + out_size > out.capacity() {
+            out.reserve(out_size);
         }
+        let mut out = zstd::OutBuffer::around_pos(out, out.len());
```
patowen, line 166: Ah, that was simpler than I expected. Somehow, I thought there would be a problem with making `OutBuffer` short-lived, but this worked perfectly fine.

Ralith: Ensuring we append rather than overwrite is the tricky bit, but luckily there was this API.

PR #379, opened 2024-04-21
Implement accurate margins between adjacent chunks

### LINKED ISSUE #41: Populate chunk margins
Chunk data sent to the GPU is wrapped in a one-voxel margin that should be, but is not currently, populated with the data from neighboring chunks. If supplied, this data would produce correct AO and hidden face detection for voxels on the face of a chunk, improving both visuals and performance.

There are two exceptions:
- Chunks which are otherwise void do not benefit from having margins set, as they have no faces to hide or occlude. Setting a margin might also interfere with efficient sparse representation of such chunks.
- Chunk faces which face graph nodes that don't yet exist, i.e. outside of the graph, cannot be populated, as their chunks are sensitive to parameters that have not yet been computed. Fortunately, we can make up whatever values we like for them: a camera within the graph cannot possibly observe any potentially exposed voxel faces along such a chunk face, and ambient occlusion errors will be difficult to see, especially in the presence of distance fog.

### LINKED ISSUE #330: Surface extraction can produce extra surfaces with a risk of z-fighting
If a voxel in the chunk's margin is solid but is next to a void voxel in the chunk's interior, the surface extraction logic will produce a surface on a face of the chunk that can be viewed from the inside. This is redundant with the adjacent chunk, causing the same voxel face to be rendered twice, which results in overdraw and could result in z-fighting (for instance, if the textures are oriented differently).

For potential surfaces that span a margin voxel and an interior voxel, the surface should only be rendered ~if the interior voxel is solid and the margin voxel is void~ if the margin voxel is solid and the interior voxel is void, not vice versa. (See comment thread below on why this edit was made)

Fixing this would also slightly reduce the worst-case number of surfaces when the dimension is even to `3*d^3` (but still `3*(d^3+d^2)` when odd). For simplicity of reasoning, though, it may not be worth taking advantage of this.

This issue won't be apparent until #41 is resolved,

patowen:
Fixes #41 (Although ambient occlusion won't be 100% accurate due to ignored edge ~and corner~ margins)
Fixes #330 (Necessary to avoid noticeable z-fighting after this change)

This PR ensures that the chunk margins that affect which surfaces are extracted are accurate, ensuring that no surfaces between two solid voxels are rendered, which should result in better draw performance and a clearer view when no-clipping beneath the terrain.

After a chunk is ready to be added to the world (which could be from terrain generation, loading saves, or downloading chunks from a server), the margins of that chunk and its adjacent chunks are modified to be consistent with each other. Then, whenever a block update occurs, margins in adjacent chunks are updated to be consistent with that block update.

To help facilitate this feature, a few additional chunk-coordinate-related math functions have been added to make it easier to reason about how adjacent chunks interact with each other, especially given that different vertices have (necessarily) inconsistent orientations.

A small change has been made to the "extract.comp" shader to prevent the wrong chunk from rendering a particular surface.

### DIFF
--- a/common/src/dodeca.rs
+++ b/common/src/dodeca.rs
@@ -1,7 +1,9 @@
 //! Tools for processing the geometry of a right dodecahedron
+
+use data::*;
 use serde::{Deserialize, Serialize};
 
-use crate::dodeca::data::*;
+use crate::voxel_math::ChunkAxisPermutation;
 
 /// Sides of a right dodecahedron
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
@@ -112,16 +114,43 @@ impl Vertex {
         sides_to_vertex()[a as usize][b as usize][c as usize]
     }
 
-    /// Sides incident to this vertex, in canonical order
+    /// Sides incident to this vertex, in canonical order.
+    ///
+    /// This canonical order determines the X, Y, and Z axes of the chunk
+    /// corresponding to the vertex.
     #[inline]
     pub fn canonical_sides(self) -> [Side; 3] {
         vertex_sides()[self as usize]
     }
 
-    /// Vertices adjacent to this vertex, opposite the sides in canonical order
+    /// Vertices adjacent to this vertex in canonical order.
+    ///
+    /// The canonical order of adjacent vertices is based on the canonical order
+    /// of sides incident to the vertex, as each of the three adjacent vertices
+    /// corresponds to one of the three sides. As for which side, when two
+    /// vertices are adjacent, they share two out of three sides of the
+    /// dodecahedron. The side they do _not_ share is the side they correspond
+    /// to.
+    ///
+    /// Put another way, anything leaving a chunk in the negative-X direction
+    /// will end up crossing `canonical_sides()[0]`, while anything leaving a
+    /// chunk in the positive-X direction will end up arriving at
+    /// `adjacent_vertices()[0]`.
     #[inline]
     pub fn adjacent_vertices(self) -> [Vertex; 3] {
         adjacent_vertices()[self as usize]
+    }
+
+    /// Chunk axes permutations for vertices adjacent to this vertex in
+    /// canonical order.
+    ///
+    /// The chunks of two adjacent vertices meet at a plane. When swiching
+    /// reference frames from one vertex to another, it is necessary to reflect
+    /// about this plane and then apply the permutation returned by this
+    /// function.
+    #[inline]
+    pub fn chunk_axis_permutations(self) -> &'static [ChunkAxisPermutation; 3] {
+        &chunk_axis_permutations()[self as usize]
     }
 
     /// For each vertex of the cube dual to this dodecahedral vertex, provides an iterator of at
@@ -227,10 +256,12 @@ pub const BOUNDING_SPHERE_RADIUS_F64: f64 = 1.2264568712514068;
 pub const BOUNDING_SPHERE_RADIUS: f32 = BOUNDING_SPHERE_RADIUS_F64 as f32;
 
 mod data {
+    use std::array;
     use std::sync::OnceLock;
 
     use crate::dodeca::{Side, Vertex, SIDE_COUNT, VERTEX_COUNT};
     use crate::math;
+    use crate::voxel_math::ChunkAxisPermutation;
 
     /// Whether two sides share an edge
     pub fn adjacent() -> &'static [[bool; SIDE_COUNT]; SIDE_COUNT] {
@@ -331,6 +362,39 @@ mod data {
                 }
             }
             result
+        })
+    }
+
+    // Which transformations have to be done after a reflection to switch reference frames from one vertex
+    // to one of its adjacent vertices (ordered similarly to ADJACENT_VERTICES)
+    pub fn chunk_axis_permutations() -> &'static [[ChunkAxisPermutation; 3]; VERTEX_COUNT] {
+        static LOCK: OnceLock<[[ChunkAxisPermutation; 3]; VERTEX_COUNT]> = OnceLock::new();
+        LOCK.get_or_init(|| {
+            array::from_fn(|vertex| {
+                array::from_fn(|result_index| {
+                    let mut test_sides = vertex_sides()[vertex];
+                    // Keep modifying the result_index'th element of test_sides until its three elements are all
+                    // adjacent to a single vertex (determined using `Vertex::from_sides`).
+                    for side in Side::iter() {
+                        if side == vertex_sides()[vertex][result_index] {
+                            continue;
+                        }
+                        test_sides[result_index] = side;
+                        let Some(adjacent_vertex) =
+                            Vertex::from_sides(test_sides[0], test_sides[1], test_sides[2])
+                        else {
+                            continue;
+                        };
+                        // Compare the natural permutation of sides after a reflection from `vertex` to `adjacent_vertex`
+                        // to the canonical permutation of the sides for `adjacent_vertex`.
+                        return ChunkAxisPermutation::from_permutation(
+                            test_sides,
+                            adjacent_vertex.canonical_sides(),
+                        );
+                    }
+                    panic!("No suitable vertex found");
+                })
+            })
         })
     }
 
@@ -482,6 +546,57 @@ mod tests {
             assert_eq!(v, Vertex::from_sides(c, a, b).unwrap());
             assert_eq!(v, Vertex::from_sides(c, b, a).unwrap());
         }
+    }
+
+    #[test]
+    fn adjacent_chunk_axis_permutations() {
+        // Assumptions for this test to be valid. If any assertions in this section fail, the test itself
+        // needs to be modified
+        assert_eq!(Vertex::A.canonical_sides(), [Side::A, Side::B, Side::C]);
+        assert_eq!(Vertex::B.canonical_sides(), [Side::A, Side::B, Side::E]);
+
+        assert_eq!(Vertex::F.canonical_sides(), [Side::B, Side::C, Side::F]);
+        assert_eq!(Vertex::J.canonical_sides(), [Side::C, Side::F, Side::H]);
+
+        // Test cases
+
+        // Variables with name vertex_?_canonical_sides_reflected refer to the canonical sides
+        // of a particular vertex after a reflection that moves it to another vertex.
+        // For instance, vertex_a_canonical_sides_reflected is similar to Vertex::A.canonical_sides(),
+        // but one of the sides is changed to match Vertex B, but the order of the other two sides is left alone.
+        let vertex_a_canonical_sides_reflected = [Side::A, Side::B, Side::E];
+        let vertex_b_canonical_sides_reflected = [Side::A, Side::B, Side::C];
+        assert_eq!(
+            Vertex::A.chunk_axis_permutations()[2],
+            ChunkAxisPermutation::from_permutation(
+                vertex_a_canonical_sides_reflected,
+                Vertex::B.canonical_sides()
+            )
+        );
+        assert_eq!(
+            Vertex::B.chunk_axis_permutations()[2],
+            ChunkAxisPermutation::from_permutation(
+                vertex_b_canonical_sides_reflected,
+                Vertex::A.canonical_sides()
+            )
+        );
+
+        let vertex_f_canonical_sides_reflected = [Side::H, Side::C, Side::F];
+        let vertex_j_canonical_sides_reflected = [Side::C, Side::F, Side::B];
+        assert_eq!(
+            Vertex::F.chunk_axis_permutations()[0],
+            ChunkAxisPermutation::from_permutation(
+                vertex_f_canonical_sides_reflected,
+                Vertex::J.canonical_sides()
+            )
+        );
+        assert_eq!(
+            Vertex::J.chunk_axis_permutations()[2],
+            ChunkAxisPermutation::from_permutation(
+                vertex_j_canonical_sides_reflected,
+                Vertex::F.canonical_sides()
+            )
+        );
     }
 
     #[test]
--- a/common/src/margins.rs
+++ b/common/src/margins.rs
@@ -1,8 +1,10 @@
 use crate::{
     dodeca::Vertex,
+    graph::Graph,
     math,
-    node::VoxelData,
+    node::{Chunk, ChunkId, VoxelData},
     voxel_math::{ChunkAxisPermutation, ChunkDirection, CoordAxis, CoordSign, Coords},
+    world::Material,
 };
 
 /// Updates the margins of both `voxels` and `neighbor_voxels` at the side they meet at.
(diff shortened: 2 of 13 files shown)


patowen:
I don't believe the CI failure is related to this PR. I'm willing to look into it a bit more closely, but the PR should still be ready to review.

patowen:
I ran some very crude benchmarks, and it seems like any performance costs for margin computation is likely insignificant enough that we don't need to put much effort into optimizing it for now. I added a few extra temporary metrics to add an additional crude benefit to measure the performance of the code in `margins.rs`:

```diff
--- a/client/src/graphics/voxels/mod.rs
+++ b/client/src/graphics/voxels/mod.rs
@@ -99,9 +99,14 @@ impl Voxels {
         for chunk in frame.drawn.drain(..) {
             self.states.peek_mut(chunk).refcount -= 1;
         }
+        let populate_chunks_started = Instant::now();
+        let mut num_chunks_populate = 0;
         while let Some(chunk) = self.worldgen.poll() {
+            num_chunks_populate += 1;
+            let populate_chunk_started = Instant::now();
             let chunk_id = ChunkId::new(chunk.node, chunk.chunk);
             sim.graph.populate_chunk(chunk_id, chunk.voxels, false);
+            metrics::histogram!("populate_chunk").record(populate_chunk_started.elapsed());

             // Now that the block is populated, we can apply any pending block updates the server
             // provided that the client couldn't apply.
@@ -112,6 +117,9 @@ impl Voxels {
                 }
             }
         }
+        if num_chunks_populate != 0 {
+            metrics::histogram!("populate_chunks").record(populate_chunks_started.elapsed());
+        }

         // Determine what to load/render
         let view = sim.view();
```

With a view distance of 90, where I load up the game with `cargo run --bin client --release` and run to the left a little bit on the bridge, I get the following on my computer.

On `master` branch:
```
key=populate_chunk percentile_25=300ns percentile_50=400ns percentile_75=500ns max=125.055µs
key=populate_chunks percentile_25=3.001µs percentile_50=4.503µs percentile_75=11.607µs max=285.439µs
key=frame.cpu percentile_25=7.069695ms percentile_50=7.553023ms percentile_75=8.146943ms max=17.154047ms
key=frame.cpu.voxels.draw percentile_25=26.111µs percentile_50=33.215µs percentile_75=41.407µs max=137.727µs
key=frame.cpu.voxels.graph_traversal percentile_25=1.458175ms percentile_50=1.560575ms percentile_75=1.758207ms max=3.268607ms
key=frame.cpu.voxels.node_scan percentile_25=2.646015ms percentile_50=2.887679ms percentile_75=3.219455ms max=11.255807ms
key=frame.gpu.draw percentile_25=1.196031ms percentile_50=1.407999ms percentile_75=1.759231ms max=3.491839ms
key=frame.gpu.after_draw percentile_25=43.711µs percentile_50=49.535µs percentile_75=92.479µs max=878.079µs
```

On `margins` branch:
```
key=populate_chunk percentile_25=300ns percentile_50=400ns percentile_75=600ns max=81.727µs
key=populate_chunks percentile_25=3.401µs percentile_50=6.403µs percentile_75=29.903µs max=942.079µs
key=frame.cpu percentile_25=7.229439ms percentile_50=7.770111ms percentile_75=8.372223ms max=14.811135ms
key=frame.cpu.voxels.draw percentile_25=26.015µs percentile_50=32.703µs percentile_75=40.703µs max=1.399807ms
key=frame.cpu.voxels.graph_traversal percentile_25=1.446911ms percentile_50=1.553407ms percentile_75=1.715199ms max=2.990079ms
key=frame.cpu.voxels.node_scan percentile_25=2.811903ms percentile_50=3.110911ms percentile_75=3.510271ms max=9.019391ms
key=frame.gpu.draw percentile_25=188.543µs percentile_50=612.863µs percentile_75=738.303µs max=1.102847ms
key=frame.gpu.after_draw percentile_25=45.567µs percentile_50=57.695µs percentile_75=118.463µs max=1.138687ms
```

To make this easier to read, here's the same thing but with numbers reduced to 2 significant figures and `master` and `margins` next to each other for easier comparison:

```
branch  metric                 [  25%,   50%,   75%,   max]

master  populate_chunk         [300ns, 400ns, 500ns, 130µs]
margins populate_chunk         [300ns, 400ns, 600ns,  82µs]

master  populate_chunks        [3.0µs, 4.5µs,  12µs, 290µs]
margins populate_chunks        [3.4µs, 6.4µs,  30µs, 940µs]

master  cpu                    [7.1ms, 7.6ms, 8.1ms,  17ms]
margins cpu                    [7.2ms, 7.8ms, 8.4ms,  15ms]

master  voxels.draw            [ 26µs,  33µs, 41µs,  140µs]
margins voxels.draw            [ 26µs,  33µs,  41µs, 1.4ms]

master  voxels.graph_traversal [1.5ms, 1.6ms, 1.8ms, 3.3ms]
margins voxels.graph_traversal [1.4ms, 1.6ms, 1.7ms, 3.0ms]

master  voxels.node_scan       [2.6ms, 2.9ms, 3.2ms,  11ms]
margins voxels.node_scan       [2.8ms, 3.1ms, 3.8ms, 9.0ms]

master  gpu.draw               [1.2ms, 1.4ms, 1.8ms, 3.5ms]
margins gpu.draw               [190µs, 610µs, 740µs, 1.1ms]

master  gpu.after_draw         [ 44µs,  50µs,  92µs, 880µs]
margins gpu.after_draw         [ 46µs,  58µs, 120µs, 1.1ms]
```

I'm not sure how much faith can be put in these numbers, as the benchmarks I ran were very crude, likely also depending on difficult-to-reproduce factors like route taken, speed, etc.

As margins are things that ought to have been done, I think this level of benchmarking is probably fine for this PR, but when we start actually working on optimizing the game, it would definitely be worth reworking how we measure performance.

patowen:
I looked into the CI failure, and I believe the root cause is https://github.com/rust-lang/jobserver-rs/issues/87, I think that hopefully, once https://github.com/rust-lang/jobserver-rs/pull/88 is deployed (which should be in version 0.1.31), this PR will be able to pass the CI checks. I believe the breakage was introduced in 0.1.29 (and continued onto 0.1.30 after 0.1.29 was yanked), and 0.1.29 was released on April 11th 2024, 10 days ago.

common/src/voxel_math.rs
```diff
@@ -56,9 +58,25 @@ pub enum CoordSign {
...
+
+impl std::ops::Mul for CoordSign {
+    type Output = CoordSign;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        if self == rhs {
+            CoordSign::Plus
+        } else {
+            CoordSign::Minus
+        }
```
Ralith, line 73: style nit: this would be more concise as a `match self == rhs`

common/src/voxel_math.rs
```diff
@@ -89,3 +107,154 @@ impl IndexMut<CoordAxis> for Coords {
...
+        [
+            Self::PLUS_X,
+            Self::PLUS_Y,
+            Self::PLUS_Z,
+            Self::MINUS_X,
+            Self::MINUS_Y,
+            Self::MINUS_Z,
+        ]
+        .iter()
+        .copied()
```
Ralith, line 154: style nit: as of a semi-recent Rust update, we can use `into_iter()` directly on an array

common/src/voxel_math.rs
```diff
@@ -89,3 +107,154 @@ impl IndexMut<CoordAxis> for Coords {
...
+    fn mul_assign(&mut self, rhs: Self) {
+        *self = *self * rhs;
+    }
+}
+
+/// Represents one of the 6 possible orientations a chunk can be viewed from, including reflections, that preserve the origin.
+/// In other words, any instance of this struct represents a permutation of the axes.
+/// This is analogous to a 3x3 rotation matrix with a restricted domain.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SimpleChunkOrientation {
```
Ralith, line 164: This name is kind of opaque. When do we want to use it vs. the non-simple kind?

common/src/margins.rs
```diff
@@ -0,0 +1,213 @@
...
+/// It is assumed that `voxels` corresponds to a chunk that lies at `vertex` and that
+/// `neighbor_voxels` is at direction `direction` from `voxels`.
+pub fn fix_margins(
+    dimension: u8,
+    vertex: Vertex,
+    voxels: &mut VoxelData,
+    direction: ChunkDirection,
+    neighbor_voxels: &mut VoxelData,
+) {
+    let neighbor_orientation = match direction.sign {
```
Ralith, line 20: Would it make sense to bundle up this entire `match` as a helper method/getter, either on `Vertex` or at least freestanding in this module?

common/src/margins.rs
```diff
@@ -0,0 +1,213 @@
...
+        CoordSign::Minus => 0,
+    };
+    let edge_coord = match direction.sign {
+        CoordSign::Plus => dimension,
+        CoordSign::Minus => 1,
+    };
+    let voxel_data = voxels.data_mut(dimension);
+    let neighbor_voxel_data = neighbor_voxels.data_mut(dimension);
+    for j in 0..dimension {
+        for i in 0..dimension {
```
Ralith, line 36: Can we factor out some common subexpressions below?

common/src/margins.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub fn to_index(self, chunk_size: u8) -> usize {
+        let chunk_size_with_margin = chunk_size as usize + 2;
+        (self.0[0] as usize)
+            + (self.0[1] as usize) * chunk_size_with_margin
+            + (self.0[2] as usize) * chunk_size_with_margin.pow(2)
+    }
+}
+
+impl From<Coords> for CoordsWithMargins {
+    fn from(value: Coords) -> Self {
```
Ralith, line 163: nit: mark trivial constructors/getters here and below `#[inline]`

common/src/node.rs
```diff
@@ -87,34 +87,43 @@ impl Graph {
...
-            // its margin cleared.
-            for chunk_direction in ChunkDirection::iter() {
-                if let Some(chunk_id) =
-                    self.get_chunk_neighbor(chunk, chunk_direction.axis, chunk_direction.sign)
+    /// Populates a chunk with the given voxel data and ensures that margins are correctly fixed up if necessary.
+    pub fn populate_chunk(&mut self, chunk: ChunkId, mut voxels: VoxelData, modified: bool) {
+        let dimension = self.layout().dimension;
+        // Fix up margins for the chunk we're inserting along with any neighboring chunks
+        for chunk_direction in ChunkDirection::iter() {
+            if let Some(Chunk::Populated {
```
Ralith, line 95: style nit: use let-else to avoid rightward drift

common/src/node.rs
```diff
@@ -87,34 +87,43 @@ impl Graph {
...
+                surface: neighbor_surface,
+                old_surface: neighbor_old_surface,
+            }) = self
+                .get_chunk_neighbor(chunk, chunk_direction.axis, chunk_direction.sign)
+                .map(|chunk_id| &mut self[chunk_id])
+            {
+                // We need to fix up margins between the current chunk and the neighboring chunk if and only if
+                // there's a potential surface between them. This can occur if either is modified or if neither
+                // is designated as solid. Note that if one is designated as solid, that means that it's deep enough
+                // in the terrain or up in the air that there will be no surface between them.
```
Ralith, line 107: > if one is designated as solid, that means that it's deep enough in the terrain or up in the air that there will be no surface between them.

Is this necessarily true? What if we have a chunk whose interior is solid void, adjacent to a chunk (solid or not) that contains solid voxels incident to the common face?

common/src/margins.rs
```diff
@@ -94,6 +96,55 @@ pub fn initialize_margins(dimension: u8, voxels: &mut VoxelData) {
...
+    let edge_coord = match direction.sign {
+        CoordSign::Plus => dimension,
+        CoordSign::Minus => 1,
+    };
+    if coords[direction.axis] != edge_coord {
+        // There is nothing to do if we're not on an edge voxel.
+        return;
+    }
+    let Some(Chunk::Populated {
+        modified: _neighbor_modified,
```
Ralith, line 120: If we don't need the field, just omit it (add a trailing `, ..`).

Ralith (changes requested):

common/src/dodeca.rs
```diff
@@ -313,6 +321,33 @@ lazy_static! {
         result
     };
 
+    static ref ADJACENT_CHUNK_ORIENTATIONS: [[SimpleChunkOrientation; 3]; VERTEX_COUNT] = {
+        let mut result = [[SimpleChunkOrientation::identity(); 3]; VERTEX_COUNT];
```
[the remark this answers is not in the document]
patowen: Good catch. I noticed that when revising my documentation for that method. I definitely down for unburying things from macros.

common/src/dodeca.rs
```diff
@@ -313,6 +321,33 @@ lazy_static! {
         result
     };
 
+    static ref ADJACENT_CHUNK_ORIENTATIONS: [[SimpleChunkOrientation; 3]; VERTEX_COUNT] = {
```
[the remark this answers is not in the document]
patowen: That seems like it would be a good idea. I just need to think of a way to test it that isn't just as error prone as the original function.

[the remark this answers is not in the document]
Ralith: Even hard-coding a few manually verified cases would be better than nothing.

common/src/voxel_math.rs
patowen: I'm rather unsure about the name as well. It could be called `ChunkAxisPermutation` or `OriginPreservingChunkOrientation`.

Now that I got rid of `ChunkOrientation`, this could take its place, but I'm not convinced `ChunkOrientation` will be gone forever, so that seems risky.

If I don't think of anything better, I'll probably rename this to `ChunkAxisPermutation`.

common/src/node.rs
patowen: That's a good question. My understanding is that `VoxelData::Solid` corresponds to chunks that are solid, including their margins. I vaguely remember that being explicitly stated somewhere in the past, but I don't remember where.

However, I also vaguely remember that this extra restriction only applied to non-void solid chunks because the purpose of that restriction was to be able to say that solid chunks don't need to be rendered at all.

I would propose that we make this apply to void chunks as well, since that would allow us to make this assumption.

common/src/margins.rs
Ralith: Yep, that's a common refactor as the amount of logic handling an enum grows.

common/src/dodeca.rs
```diff
@@ -313,6 +321,33 @@ lazy_static! {
...
+        let mut result = [[SimpleChunkOrientation::identity(); 3]; VERTEX_COUNT];
+    for vertex in 0..VERTEX_COUNT {
+        for result_index in 0..3 {
```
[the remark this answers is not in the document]
patowen: Whoa, I had no idea such functionality existed. It looks like it got added after I started work on Hypermine, although I really thought there was no way to initialize an array like that with a one-liner. Thanks for letting me know about this feature!

common/src/voxel_math.rs
patowen: That does look nicer. I wonder if it's worth defaulting to that in situations where I would use the ternary operator in other programming languages. I think there might be some circumstances where that's less readable, so I'll at least keep in mind that it's an option.

common/src/margins.rs
patowen: Good question. I'm also considering bundling it into the return value of `get_chunk_neighbor`.

common/src/node.rs
Ralith: > However, I also vaguely remember that this extra restriction only applied to non-void solid chunks because the purpose of that restriction was to be able to say that solid chunks don't need to be rendered at all.

Skimming the worldgen logic supports this hypothesis: we generate solid dirt chunks or solid void chunks when distant from the surface, and otherwise start with a solid void chunk and possibly fill in some voxels. Hard to say for sure how exact "distant from the surface" is, though.

> I would propose that we make this apply to void chunks as well, since that would allow us to make this assumption.

Meaning we convert solid void chunks to dense chunks if their margins are touched? Sure, that seems reasonable.

common/src/dodeca.rs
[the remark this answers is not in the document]
Ralith: Rust gets cool new stuff with some regularity. blog.rust-lang.org is good reading!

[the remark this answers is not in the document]
patowen: Oh yeah, I have been checking up on it pretty regularly, but that particular feature just flew under the radar.

common/src/node.rs
patowen: In that case, I believe this PR is in a good state. Unless the logic was done with a razor-thin margin, worldgen already ensures that solid void chunks don't have any voxels directly adjacent to them since the call to `const ELEVATION_MARGIN: f64 = 0.7;` was empirically determined.

However, I'll see if I can find the discussion I vaguely recall about the meaning of `VoxelData::Solid`. I'll leave this comment thread unresolved until you confirm that you're happy with the current state of things.

common/src/dodeca.rs
```diff
@@ -124,6 +126,14 @@ impl Vertex {
         adjacent_vertices()[self as usize]
     }
 
+    /// Chunk permutations adjacent to this vertex, opposite the sides in canonical order.
+    /// These permutations convert coordinates relative to a reflected version of the current
+    /// chunk to coordinates relative to the adjacent chunk in its canonical orientation.
```
Ralith, line 131: This is pretty opaque. Why are the coordinates initially reflected wrt. the local chunk, and in which way? Is a "chunk permutation" different from an axis permutation?

common/src/dodeca.rs
```diff
@@ -334,6 +346,39 @@ mod data {
         })
     }
 
+    // Which transformations have to be done after a reflection to switch reference frames from one vertex
+    // to one of its adjacent vertices (ordered similarly to ADJACENT_VERTICES)
```
Ralith, line 350: Is "adjacent vertices" the right term here? What about the chunk associated with the same vertex in an adjacent dodecahedron?

Ralith (approved):

common/src/dodeca.rs
patowen: I believe it should be. They're vertices in a single dodecahedron that share an edge. Since the term used is "adjacent vertices" and not "vertices of adjacent chunks", I don't think it's ambiguous.

common/src/dodeca.rs
patowen: I'll need to think a bit more about the phrasing of this. I'll also need to avoid the phrase "chunk permutation" because that makes no sense. I think I got that phrase by find/replacing orientation to permutation.

There is only one way to reflect a chunk and reach another chunk in the same dodeca, and that is to reflect it about the surface that they share. I'll probably want to mention/describe that surface in these docs, as it's pretty confusing.

As for "opposite the sides in canonical order", that also feels somewhat vague. I'll revisit all of this documentation before merging to make it less opaque.

patowen: @Ralith, I believe I figured out a clearer way to word this documentation, but please double-check that it makes sense. I'll hold off on merging until then.

Ralith (approved):

PR #380, opened 2024-04-21
Update rcgen to 0.13.1

patowen:
This upgrade is based on the docs at https://github.com/rustls/rcgen/blob/main/rcgen/docs/0.12-to-0.13.md.


patowen:
This PR "works", but it results in the installation of two versions of `ring`, one as a dependency of `rcgen`, and the other as a dependency of `quinn-proto`, so it's possible we might want to wait on this one as well. That being said, it doesn't seem to cause any actual compatibility issues, so this PR may be in a good state to merge anyway.

Ralith (approved):
New Quinn release is coming Soon™, so probably fine to get this out of the way.

Issue #381, opened 2024-04-22
Modernize Rust idioms


Ralith:
Tracking minor cosmetic improvements enabled by recent changes to Rust:

- [x] Replace `.iter().copied()` on arrays with `.into_iter()`
- [x] Replace `lazy_static` with const logic where possible, and otherwise [`OnceLock`](std::sync::OnceLock) (or eventually `LazyLock`, once stabilized)
- [ ] Replace default-then-linearly-overwrite array construction patterns with `array::from_fn`

inthar-raven:
What's the right way to make a lazy `OnceLock`? `Lazy<OnceLock<_>>` or `OnceLock<Lazy<_>>`?

Ralith:
I don't know what the `Lazy` you're referring to is.

inthar-raven:
`once_cell::sync::Lazy`
I tried to use a `OnceCell` for the static values and was suggested to wrap the whole thing in `Lazy::new(||)`. Since I don't think `OnceLock` can be initialized statically, I think `Lazy<OnceLock<_>>` is the right option. Though now every time I access the static values I have to uset `.get().unwrap()` first...

inthar-raven:
Here's a bigger problem with `Lazy<OnceLock<_>>`...
![image](https://github.com/Ralith/hypermine/assets/36112167/44b9e232-201d-47f7-bf78-f2b09afaed89)

Ralith:
We shouldn't introduce a dependency on `once_cell`. `OnceLock` is part of std. [The documentation](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) illustrates how to use it in a static.

patowen:
I think using const logic would still be a bit painful due to being unable to use iterators (so we should probably avoid it for now to avoid potential burnout). `OnceLock` seems promising, though.

Ralith:
I think there's at least a handful of low hanging fruit that don't use iterators. Definitely no need to force the issue, though. A more serious limitation might be the missing transcendental operations on floats.

PR #382, opened 2024-04-22
Replace .iter().copied() on arrays with .into_iter()

inthar-raven:
ref #381

### DIFF
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -283,9 +283,9 @@ impl Sim {
         populate_fresh_nodes(&mut self.graph);
 
         let mut fresh_voxel_data = vec![];
-        for fresh_node in (&fresh_nodes).into_iter() {
+        for fresh_node in fresh_nodes.iter().cloned() {
             for vertex in Vertex::iter() {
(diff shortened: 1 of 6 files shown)


server/src/sim.rs
```diff
@@ -283,9 +283,9 @@ impl Sim {
         populate_fresh_nodes(&mut self.graph);
 
         let mut fresh_voxel_data = vec![];
-        for fresh_node in fresh_nodes.iter().copied() {
+        for fresh_node in (&fresh_nodes).into_iter() {
```
Ralith, line 286: `fresh_nodes` is a `Vec`, not an array, so it's not `Copy`, so this change doesn't improve clarity or concision.

Ralith (changes requested):
We'll also want to handle the various cases of `.iter().cloned()`, which seems to actually be the more common, functionally equivalent pattern.

inthar-raven:
`.into_iter()` consumes the iterator, and that caused some problems.

Ralith (approved):
Thanks!

PR #383, opened 2024-04-24
Replace lazy_static! with OnceLock

inthar-raven:
(ref #381)

### DIFF
--- a/common/src/cursor.rs
+++ b/common/src/cursor.rs
@@ -1,6 +1,7 @@
 use std::sync::OnceLock;
 
-use crate::dodeca::{Side, Vertex, SIDE_COUNT};
+use crate::dodeca::SIDE_COUNT;
+use crate::dodeca::{Side, Vertex};
 use crate::graph::{Graph, NodeId};
 use crate::node::ChunkId;
 
--- a/common/src/cursor.rs
+++ b/common/src/cursor.rs
@@ -1,8 +1,8 @@
+use std::sync::OnceLock;
+
 use crate::dodeca::{Side, Vertex, SIDE_COUNT};
 use crate::graph::{Graph, NodeId};
 use crate::node::ChunkId;
-
-use lazy_static::lazy_static;
 
 /// Navigates the cubic dual of a graph
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
@@ -28,9 +28,9 @@ impl Cursor {
         // in both the dodecahedron sharing the face unique to the new vertex and that sharing the
         // face that the new vertex isn't incident to.
         let (a, b, c) = (self.a, self.b, self.c);
-        let a_prime = NEIGHBORS[a as usize][b as usize][c as usize].unwrap();
-        let b_prime = NEIGHBORS[b as usize][a as usize][c as usize].unwrap();
-        let c_prime = NEIGHBORS[c as usize][b as usize][a as usize].unwrap();
+        let a_prime = neighbors()[a as usize][b as usize][c as usize].unwrap();
+        let b_prime = neighbors()[b as usize][a as usize][c as usize].unwrap();
+        let c_prime = neighbors()[c as usize][b as usize][a as usize].unwrap();
         use Dir::*;
         let (sides, neighbor) = match dir {
             Left => ((a, b, c_prime), c),
@@ -103,9 +103,10 @@ impl std::ops::Neg for Dir {
     }
 }
 
-lazy_static! {
-    /// Maps every (A, B, C) sharing a vertex to A', the side that shares edges with B and C but not A
-    static ref NEIGHBORS: [[[Option<Side>; SIDE_COUNT]; SIDE_COUNT]; SIDE_COUNT] = {
+/// Maps every (A, B, C) sharing a vertex to A', the side that shares edges with B and C but not A
+fn neighbors() -> &'static [[[Option<Side>; SIDE_COUNT]; SIDE_COUNT]; SIDE_COUNT] {
+    static LOCK: OnceLock<[[[Option<Side>; SIDE_COUNT]; SIDE_COUNT]; SIDE_COUNT]> = OnceLock::new();
+    LOCK.get_or_init(|| {
         let mut result = [[[None; SIDE_COUNT]; SIDE_COUNT]; SIDE_COUNT];
         for a in Side::iter() {
             for b in Side::iter() {
@@ -114,19 +115,20 @@ lazy_static! {
                         if s == a || s == b || s == c {
                             continue;
                         }
-                        let (opposite, shared) = match (s.adjacent_to(a), s.adjacent_to(b), s.adjacent_to(c)) {
-                            (false, true, true) => (a, (b, c)),
-                            (true, false, true) => (b, (a, c)),
-                            (true, true, false) => (c, (a, b)),
-                            _ => continue,
-                        };
+                        let (opposite, shared) =
+                            match (s.adjacent_to(a), s.adjacent_to(b), s.adjacent_to(c)) {
+                                (false, true, true) => (a, (b, c)),
+                                (true, false, true) => (b, (a, c)),
+                                (true, true, false) => (c, (a, b)),
+                                _ => continue,
+                            };
                         result[opposite as usize][shared.0 as usize][shared.1 as usize] = Some(s);
                     }
                 }
             }
         }
         result
-    };
+    })
 }
 
 #[cfg(test)]
@@ -139,8 +141,8 @@ mod tests {
         for v in Vertex::iter() {
             let [a, b, c] = v.canonical_sides();
             assert_eq!(
-                NEIGHBORS[a as usize][b as usize][c as usize],
-                NEIGHBORS[a as usize][c as usize][b as usize]
+                neighbors()[a as usize][b as usize][c as usize],
+                neighbors()[a as usize][c as usize][b as usize]
             );
         }
     }
--- a/common/src/dodeca.rs
+++ b/common/src/dodeca.rs
@@ -1,9 +1,7 @@
 //! Tools for processing the geometry of a right dodecahedron
-use std::sync::OnceLock;
-
 use serde::{Deserialize, Serialize};
 
-use crate::math;
+use crate::dodeca::data::*;
 
 /// Sides of a right dodecahedron
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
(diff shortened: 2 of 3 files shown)


inthar-raven:
What's considered a good naming convention for functions that return `&'static` values?

common/src/cursor.rs
```diff
@@ -103,9 +102,10 @@ impl std::ops::Neg for Dir {
     }
 }
 
-lazy_static! {
-    /// Maps every (A, B, C) sharing a vertex to A', the side that shares edges with B and C but not A
-    static ref NEIGHBORS: [[[Option<Side>; SIDE_COUNT]; SIDE_COUNT]; SIDE_COUNT] = {
+/// Maps every (A, B, C) sharing a vertex to A', the side that shares edges with B and C but not A
+fn neighbors_static() -> &'static [[[Option<Side>; SIDE_COUNT]; SIDE_COUNT]; SIDE_COUNT] {
```
Ralith, line 106: `_static` is an unhelpful suffix. Anything returned by a free function must necessarily have a static lifetime. Did you actually encounter any collisions when leaving it off?

common/src/cursor.rs
```diff
@@ -1,8 +1,7 @@
 use crate::dodeca::{Side, Vertex, SIDE_COUNT};
 use crate::graph::{Graph, NodeId};
 use crate::node::ChunkId;
-
-use lazy_static::lazy_static;
+use std::sync::OnceLock;
```
Ralith, line 4: style nit: `std` use statements should go in separate block above others.

common/src/dodeca.rs
```diff
@@ -361,25 +402,51 @@ lazy_static! {
...
+    static LOCK: OnceLock<[na::Matrix4<f32>; VERTEX_COUNT]> = OnceLock::new();
+    LOCK.get_or_init(|| dual_to_node_f64_static().map(|n| n.cast()))
+}
+
+fn node_to_dual_f32_static() -> &'static [na::Matrix4<f32>; VERTEX_COUNT] {
+    static LOCK: OnceLock<[na::Matrix4<f32>; VERTEX_COUNT]> = OnceLock::new();
+    LOCK.get_or_init(|| node_to_dual_f64_static().map(|n| n.cast()))
+}
+
+fn dual_to_chunk_factor_f32_static() -> &'static f32 {
```
Ralith, line 442: Don't return (or pass) floats by reference. The reference is bigger then the bare float would be.

common/src/dodeca.rs
```diff
@@ -255,97 +260,133 @@ lazy_static! {
...
-        let mip_origin_normal = math::mip(&math::origin(), &SIDE_NORMALS_F64[0]); // This value is the same for every side
+/// Transform that converts from cube-centric coordinates to dodeca-centric coordinates
+fn dual_to_node_f64_static() -> &'static [na::Matrix4<f64>; VERTEX_COUNT] {
+    static LOCK: OnceLock<[na::Matrix4<f64>; VERTEX_COUNT]> = OnceLock::new();
+    LOCK.get_or_init(|| {
+        let mip_origin_normal = math::mip(&math::origin(), &side_normals_f64_static()[0]); // This value is the same for every side
         let mut result = [na::zero(); VERTEX_COUNT];
-        for i in 0..VERTEX_COUNT {
-            let [a, b, c] = VERTEX_SIDES[i];
+        for (i, item) in result.iter_mut().enumerate().take(VERTEX_COUNT) {
```
Ralith, line 343: `take` is redundant here: `result.iter_mut()` already has a length of exactly `VERTEX_COUNT`.

Also, please keep unrelated refactoring in a separate commit. This makes it much easier to ensure that each change has the intended effect when reviewing.

Ralith (changes requested):
Don't forget to remove the `lazy_static` dependency from Cargo.toml.

common/src/cursor.rs
inthar-raven: I did:

```
error[E0599]: no method named `iter` found for unit type `()` in the current scope
   --> common/src/dodeca.rs:460:38
    |
460 |         let triples = vertex_sides().iter().collect::<HashSet<_>>();
    |                                      ^^^^ method not found in `()`

error[E0277]: `()` is not an iterator
   --> common/src/dodeca.rs:462:24
    |
462 |         for &triple in vertex_sides() {
    |                        ^^^^^^^^^^^^^^ `()` is not an iterator
    |
    = help: the trait `Iterator` is not implemented for `()`, which is required by `(): IntoIterator`
    = note: required for `()` to implement `IntoIterator`
```

Ralith: Rename the test function to `vertex_sides_consistent()` or something.

patowen:
I'm wondering if this can be refactored even further. Now that we don't have to put everything in one big `lazy_static` block, we might not need the helper functions at all, and instead, all the static `OnceLock`s could live in each of the public functions instead.

patowen:
Regarding my previous comment, Initial experimentation on my end suggests that it is indeed possible, although it would be trickier due to not as easily being able to directly convert one array to another. An example implementation for `Side::normal_64` could be as simple as
```rust
/// Outward normal vector of this side
#[inline]
pub fn normal_f64(self) -> &'static na::Vector4<f64> {
    static SIDE_NORMALS_64: OnceLock<[na::Vector4<f64>; SIDE_COUNT]> = OnceLock::new();
    &SIDE_NORMALS_64.get_or_init(|| {
        let phi = libm::sqrt(1.25) + 0.5; // golden ratio
        let f = math::lorentz_normalize(&na::Vector4::new(1.0, phi, 0.0, libm::sqrt(phi)));

        let mut result: [na::Vector4<f64>; SIDE_COUNT] = [na::zero(); SIDE_COUNT];
        let mut i = 0;
        for (x, y, z, w) in [
            (f.x, f.y, f.z, f.w),
            (-f.x, f.y, -f.z, f.w),
            (f.x, -f.y, -f.z, f.w),
            (-f.x, -f.y, f.z, f.w),
        ] {
            for (x, y, z, w) in [(x, y, z, w), (y, z, x, w), (z, x, y, w)] {
                result[i] = na::Vector4::new(x, y, z, w);
                i += 1;
            }
        }
        result
    })[self as usize]
}
```

However, the implementation for `Side::normal` is not as easy, as it would need to be something like
```rust
/// Outward normal vector of this side
#[inline]
pub fn normal(self) -> &'static na::Vector4<f32> {
    static SIDE_NORMALS_F32: OnceLock<[na::Vector4<f32>; SIDE_COUNT]> = OnceLock::new();
    &SIDE_NORMALS_F32.get_or_init(|| {
        Side::iter()
            .map(|side| side.normal_f64().cast())
            .collect::<Vec<_>>()
            .try_into()
            .unwrap()
    })[self as usize]
}
```

Similar changes would have to be made to any field that depends on another field (which is most of them). @Ralith may have ideas on whether this extra complexity is worth this consolidation, but my current thought is that it probably isn't (although it could be in the future if paired with some other refactors that would make these implementations easier).

I think my suggestion for now to avoid making this PR too involved would be to disregard my previous suggestion in removing the helper functions and instead just dump all the things that used to be `lazy_static` into a private module within `dodeca.rs`.

Ralith:
It's rarely a bad idea to keep things simple and narrowly scoped.

common/src/dodeca.rs
```diff
@@ -171,234 +169,305 @@ impl Vertex {
...
-        CHUNK_TO_NODE_PARITY[self as usize]
+        chunk_to_node_parity()[self as usize]
     }
 }
 
-pub const VERTEX_COUNT: usize = 20;
-pub const SIDE_COUNT: usize = 12;
-pub const BOUNDING_SPHERE_RADIUS_F64: f64 = 1.2264568712514068;
-pub const BOUNDING_SPHERE_RADIUS: f32 = BOUNDING_SPHERE_RADIUS_F64 as f32;
+pub mod gridding {
```
Ralith, line 224: The suggestion was to make this a *private* module. Not something other modules have access to.

common/src/dodeca.rs
```diff
@@ -171,234 +169,305 @@ impl Vertex {
...
+    use std::sync::OnceLock;
+
+    use crate::dodeca::{Side, Vertex};
+    use crate::math;
+
+    pub const VERTEX_COUNT: usize = 20;
+    pub const SIDE_COUNT: usize = 12;
+    pub const BOUNDING_SPHERE_RADIUS_F64: f64 = 1.2264568712514068;
+    pub const BOUNDING_SPHERE_RADIUS: f32 = BOUNDING_SPHERE_RADIUS_F64 as f32;
```
Ralith, line 233: There is no reason to move these out of `dodeca`, and doing so causes a lot of unnecessary churn downstream.

Ralith (changes requested):

patowen:
Looks like the CI failure in `CI / Check protos` is probably resolved. I'll try rerunning that job.

common/src/dodeca.rs
inthar-raven: A number of other modules use these constants. I'll have to move things around if I make everything in the module private.

Ralith: These constants should be left where they were.

patowen: The suggestion is to move these out of the `gridding` module and to make the `gridding` module private. These functions were not originally in the `lazy_static!` block, so they should go directly in the `dodeca` module as they were before.

The only files that need to change in this PR are `dodeca.rs` and `cursor.rs`.

common/src/dodeca.rs
patowen, line 224: I wonder what this module should be called. I expect that `gridding` would be confusing because `dodeca` doesn't care about individual voxels. Some names that pop into my head are `internal_data`, `lazy_static`, `lazy_data`, or `once_locks`, but none of them feel that great.

If I had to decide, I would just call it `internal_data`, but I'm wondering if you or @Ralith have any thoughts here.

Ralith: That'd be fine, or `data`, `tables`, `details`, anything like that.

client/src/graphics/voxels/mod.rs
```diff
@@ -142,7 +142,7 @@ impl Voxels {
         for &(node, ref node_transform) in &nodes {
             let node_to_view = local_to_view * node_transform;
             let origin = node_to_view * math::origin();
-            if !frustum_planes.contain(&origin, dodeca::BOUNDING_SPHERE_RADIUS) {
+            if !frustum_planes.contain(&origin, BOUNDING_SPHERE_RADIUS) {
```
Ralith, line 145: Please leave this qualified as `dodeca::BOUNDING_SPHERE_RADIUS`, as it was originally, here and in the other occurrences. That form is intentional, as it makes it clear what bounding sphere we're talking about.

inthar-raven:
No idea why Clippy didn't catch formatting errors locally...

common/src/dodeca.rs
```diff
@@ -1,9 +1,12 @@
...
-use lazy_static::lazy_static;
 use serde::{Deserialize, Serialize};
 
-use crate::math;
+use crate::dodeca::data::*;
+
+pub const VERTEX_COUNT: usize = 20;
+pub const SIDE_COUNT: usize = 12;
+pub const BOUNDING_SPHERE_RADIUS_F64: f64 = 1.2264568712514068;
+pub const BOUNDING_SPHERE_RADIUS: f32 = BOUNDING_SPHERE_RADIUS_F64 as f32;
```
Ralith, line 9: Try to avoid moving code around if you don't need to. These were originally underneath `impl Vertex`. Moving them forces code reviewers to verify that no unintended changes were made to them, and makes `git blame` output less useful.

patowen:
> No idea why Clippy didn't catch formatting errors locally...

The errors caught by Clippy are different from the formatting issues caught by `rustfmt`. To catch the latter, you have to run `cargo fmt` to have it automatically format your code, or `cargo fmt --check` if you just want to check for formatting errors (although I recommend the former).

Ralith (changes requested):
Thanks, pretty close now! It's delightful to see rustfmt start working for all that formerly macro-wrapped code.

common/src/dodeca.rs
```diff
@@ -171,234 +174,300 @@ impl Vertex {
...
+    pub fn chunk_to_dual_factor_f32() -> f32 {
+        static LOCK: OnceLock<f32> = OnceLock::new();
+        *LOCK.get_or_init(|| chunk_to_dual_factor_f64() as f32)
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::math;
```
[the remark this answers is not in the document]
patowen: I think I see the confusion. The reason this line isn't needed on the `master` branch is because `dodeca.rs` as a whole uses the `math` module. However, everything using that `math` module has been moved into the `data` module, so the `dodeca` module no longer uses the `math` module directly. Therefore, the tests have to import it directly instead rather than relying on `super::*` to obtain the `math` module.

client/src/graphics/voxels/mod.rs
```diff
@@ -142,7 +141,7 @@ impl Voxels {
         for &(node, ref node_transform) in &nodes {
             let node_to_view = local_to_view * node_transform;
             let origin = node_to_view * math::origin();
-            if !frustum_planes.contain(&origin, dodeca::BOUNDING_SPHERE_RADIUS) {
+            if !frustum_planes.contain(&origin, common::dodeca::BOUNDING_SPHERE_RADIUS) {
```
Ralith, line 144: Now you've instead introduced extra, unnecessary qualification. Please restore this to the original form, `dodeca::BOUNDING_SPHERE_RADIUS`. In general, avoid modifying code that is not related to the intent of your changes.

common/src/traversal.rs
```diff
@@ -185,8 +186,7 @@ impl<'a> RayTraverser<'a> {
                     // because the AABB check can have false positives.
                     let ray_node_distance = (next_node_transform * self.ray.position).w.acosh();
                     let ray_length = tanh_distance.atanh();
-                    if ray_node_distance - ray_length - self.radius > dodeca::BOUNDING_SPHERE_RADIUS
-                    {
+                    if ray_node_distance - ray_length - self.radius > BOUNDING_SPHERE_RADIUS {
```
Ralith, line 189: Please keep this in the original form, `dodeca::BOUNDING_SPHERE_RADIUS`.

server/src/sim.rs
```diff
@@ -296,7 +296,7 @@ impl Sim {
 
         // We want to load all chunks that a player can interact with in a single step, so chunk_generation_distance
         // is set up to cover that distance.
-        let chunk_generation_distance = dodeca::BOUNDING_SPHERE_RADIUS
+        let chunk_generation_distance = BOUNDING_SPHERE_RADIUS
```
Ralith, line 299: Please keep this in the original form, `dodeca::BOUNDING_SPHERE_RADIUS`.

Ralith (changes requested):

inthar-raven:
The reason you're seeing these changes might be because I've been using VSCode (which checks for errors on the fly). Does that have any unwanted behavior?

inthar-raven:
(The reason I made these changes was to get the code to be validated)

common/src/traversal.rs
inthar-raven: But:

```rs
error[E0433]: failed to resolve: use of undeclared crate or module `dodeca`
   --> common/src/traversal.rs:188:71
    |
188 | ...radius > dodeca::BOUNDING_SPHERE_RADIUS {
    |             ^^^^^^ use of undeclared crate or module `dodeca`
    |
help: consider importing this module
    |
1   + use crate::dodeca;
```

And it doesn't see the imports for some reason:
```rs
use crate::{
    collision_math::Ray,
    dodeca::{Side, Vertex, BOUNDING_SPHERE_RADIUS},
    graph::{Graph, NodeId},
    math,
    node::ChunkId,
    proto::Position,
};
```

inthar-raven: (I tried this on the command line, not VSCode.)

inthar-raven: I'll look into this in a couple of days.

patowen: You'll need to undo your change to the import as as well. In `traversal.rs`, the original import from `dodeca` was `dodeca::{self, Side, Vertex},`. The `self` is what allowed `dodeca` to resolve.

inthar-raven:
Done. Changes should just be on Cargo.toml, cursor.rs and dodeca.rs now.

Ralith (approved):
Thanks!

PR #386, opened 2024-04-30
Fix inconsistencies with margin reconciliation logic

patowen:
This PR fixes two bugs related to https://github.com/Ralith/hypermine/pull/379.

## Bug 1
Some of the logic assumed that world generation would never create "solid" chunks unless it was sure that the chunk was deep enough underground or high enough up in the air to not touch any surfaces. This assumption allowed us to avoid reconciling margins between a solid chunk and a dense chunk unless the dense chunk was "modified".

However, since this assumption was incorrect (for instance, due to world generation not turning solid "void" chunks into dense chunks unless a voxel is actually generated), this produced gaps in the terrain whenever a dense chunk met a solid void chunk.

## Bug 2
When a voxel is modified, we assumed that only neighboring chunks' margins needs to update, since all other margins should already be correct. However, if we want to be certain, we also need to update the margins of the chunk the modified voxel is in. This is because we only guarantee that margin voxels are the correct material when they would be rendered.

Because of this, it would have been possible to dig down into a solid dirt chunk, dig across a bit, and dig up again, seeing what looks like dirt in a neighboring chunk's voxel no matter what material that voxel actually is.

## Solution
The fix to bug 1 is to remove the assumption, adding logic to `fix_margins` to handle either chunk being solid.

The fix to bug 2 is to change `update_margin_voxel` into a bidirectional update, updating the margins of both relevant chunks. Since this changes the public interface of the function, I also renamed it to `reconcile_margin_voxels`.

## Bonus cleanup
The only purpose of the `modified` boolean in `Chunk::Populated` was to determine whether assumptions based on world generation would hold when determining whether chunks should be solid or not. Since we no longer make assumptions based on world generation, this field is now obsolete, allowing us to remove it completely.

### DIFF
--- a/common/src/margins.rs
+++ b/common/src/margins.rs
@@ -192,7 +192,6 @@ pub fn reconcile_margin_voxels(
         voxels: neighbor_voxels,
         surface: neighbor_surface,
         old_surface: neighbor_old_surface,
-        ..
     } = &mut graph[neighbor_chunk]
     else {
         unreachable!();
@@ -213,7 +212,6 @@ pub fn reconcile_margin_voxels(
         voxels,
         surface,
         old_surface,
-        ..
     } = &mut graph[chunk]
     else {
         unreachable!();
@@ -383,7 +381,6 @@ mod tests {
         for chunk in [current_chunk, node_neighbor_chunk, vertex_neighbor_chunk] {
             *graph.get_chunk_mut(chunk).unwrap() = Chunk::Populated {
                 voxels: VoxelData::Solid(Material::Void),
-                modified: false,
                 surface: None,
                 old_surface: None,
             };
--- a/common/src/margins.rs
+++ b/common/src/margins.rs
@@ -139,50 +139,87 @@ pub fn initialize_margins(dimension: u8, voxels: &mut VoxelData) {
     }
 }
 
-/// Assuming that the voxel at `chunk` and `coords` is set to `material`, updates the cooresponding
-/// margin in the chunk at direction `direction` from `chunk` if such a margin exists. Unpopulated chunks
-/// are ignored.
-pub fn update_margin_voxel(
+/// Based on the given `coords` and the neighboring voxel at direction
+/// `direction` (if it's in a different chunk), updates both of their respective
+/// margins to match each others' materials.
+pub fn reconcile_margin_voxels(
     graph: &mut Graph,
     chunk: ChunkId,
     coords: Coords,
     direction: ChunkDirection,
-    material: Material,
 ) {
-    let coords: CoordsWithMargins = coords.into();
+    let coords_of_boundary_voxel: CoordsWithMargins = coords.into();
     let dimension = graph.layout().dimension();
-    let boundary_coord = match direction.sign {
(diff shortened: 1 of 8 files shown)


common/src/margins.rs
```diff
@@ -21,6 +21,39 @@ pub fn fix_margins(
...
+        let is_void = voxels.get(0) == Material::Void;
+        if all_voxels_at_edge(
+            dimension,
+            neighbor_voxels,
+            neighbor_axis_permutation * direction,
+            |m| (m == Material::Void) == is_void,
+        ) {
+            return;
+        }
+    }
```
Ralith, line 54: nit: dedup? `for (a, b) in [(&*voxels, &*neighbor_voxels), (&*neighbor_voxels, &*voxels)]`

common/src/node.rs
```diff
@@ -98,31 +98,24 @@ impl Graph {
...
                 .get_chunk_neighbor(chunk, chunk_direction.axis, chunk_direction.sign)
                 .map(|chunk_id| &mut self[chunk_id])
             else {
                 continue;
             };
-            // We need to fix up margins between the current chunk and the neighboring chunk if and only if
-            // there's a potential surface between them. This can occur if either is modified or if neither
-            // is designated as solid. Note that if one is designated as solid, that means that it's deep enough
-            // in the terrain or up in the air that there will be no surface between them.
-            if (!voxels.is_solid() && !neighbor_voxels.is_solid()) || modified || *neighbor_modified
```
Ralith, line 115: nit: delete "if necessary" from the comment on this function

common/src/margins.rs
```diff
@@ -138,50 +138,86 @@ pub fn initialize_margins(dimension: u8, voxels: &mut VoxelData) {
...
+    let coords_of_edge_voxel: CoordsWithMargins = coords.into();
     let dimension = graph.layout().dimension();
-    let edge_coord = match direction.sign {
-        CoordSign::Plus => dimension,
-        CoordSign::Minus => 1,
-    };
-    if coords[direction.axis] != edge_coord {
-        // There is nothing to do if we're not on an edge voxel.
+
+    // There is nothing to do if we're not on an edge voxel.
```
Ralith, line 153: Face, not edge, right?

common/src/margins.rs
patowen: Potentially, but for readability reasons, I'll probably avoid deduping here. The issue is that I also would need to distinguish between `direction` and `neighbor_axis_permutation * direction`.

common/src/margins.rs
patowen: Fair. I think my use of `edge_coord` instead of something like `boundary_coord` was a mistake. I'll fix this up in multiple places.

I'm happy to say "face voxel" or "boundary voxel" here. The former is more specific, while the latter would be more consistent with `boundary_coord`. (I don't want to say `face_coord` because that seems like a less suitable name for a single coordinate value).

patowen:
I believe the extra changes I made to this PR are noncontroversial enough, so I'll go ahead and merge when CI passes.

PR #393, opened 2024-05-05
Don't package PRs

Ralith:
Packaging has little value as CI, and is slow.


patowen:
It did catch https://github.com/Ralith/hypermine/pull/325#issuecomment-1817375348, but we could have caught it in the `master` branch instead and reverted if necessary, so the convenience of having fewer builds to wait for seem worth it to me as well.

PR #391, opened 2024-05-05
Add an initial UI to Hypermine

patowen:
This is currently a draft, as there are two issues:
- Fixed: ~Since Hypermine has two subpasses (the second one for fog), to ensure the UI gets drawn on top of everything and is unaffected by fog, the UI needs to be in the second subpass or possibly a third one. Currently, Yakui doesn't seem to support rendering anywhere except the first subpass. (It creates a `vk::GraphicsPipelineCreateInfo` struct and keeps the `subpass` field in that struct as its default.)~
- Fixed: ~I get the following validation error on the second call to `draw::draw` (but strangely, no frames before or after):
`Validation Error: [ VUID-vkFreeMemory-memory-00677 ] | MessageID = 0x485c8ea2 | vkFreeMemory():  can't be called on VkDeviceMemory 0xeeeba00000000288[] that is currently in use by VkBuffer 0x8f6f470000000287[]. The Vulkan spec states: All submitted commands that refer to memory (via images or buffers) must have completed execution (https://vulkan.lunarg.com/doc/view/1.3.280.0/windows/1.3-extensions/vkspec.html#VUID-vkFreeMemory-memory-00677) id=VUID-vkFreeMemory-memory-00677 number=1214025378 queue_labels= cmd_labels= objects=`. The error seems to trigger when I call `transfers_finished`. I haven't fully looked into the root cause yet, so it could just be a simple mistake on my part, as I've mainly been blindly following the `yakui-vulkan` example.~

Note that long-term, I'm not sure whether we can keep putting fog in a separate subpass, since that likely wouldn't work properly if there are any transparent objects in the scene.

(Fixed) ~If I ignore the fog and just render the UI after everything else in the first subpass, the letters are replaced with their bounding boxes filled in with the sky color. If I turn the fog off, everything works.~

I would like to keep this PR as a draft until we have the following (which I may file as a separate PR, closing this one):
- No Vulkan validation errors
- A central cursor
- Some indicator (likely textual) of the block you have selected
- A UI toggle for screenshot purposes
- Good code quality

PR #395, opened 2024-05-06
Add a basic GUI to Hypermine powered by yakui

patowen:
This PR adds a GUI with the following features:
- A central cursor
- Some indicator (likely textual) of the block you have selected
- The ability to turn it on and off (for screenshot purposes) with F1

image

### DIFF
--- a/client/src/graphics/base.rs
+++ b/client/src/graphics/base.rs
@@ -139,7 +139,11 @@ impl Base {
                             .queue_create_infos(&[vk::DeviceQueueCreateInfo::default()
                                 .queue_family_index(queue_family_index)
                                 .queue_priorities(&[1.0])])
-                            .enabled_extension_names(&device_exts),
+                            .enabled_extension_names(&device_exts)
+                            .push_next(
+                                &mut vk::PhysicalDeviceVulkan12Features::default()
+                                    .descriptor_binding_partially_bound(true),
+                            ),
                         None,
                     )
                     .unwrap(),
--- a/client/src/graphics/draw.rs
+++ b/client/src/graphics/draw.rs
@@ -48,6 +48,9 @@ pub struct Draw {
     /// Reusable storage for barriers that prevent races between buffer upload and read
     buffer_barriers: Vec<vk::BufferMemoryBarrier<'static>>,
 
+    /// Yakui Vulkan context
+    yakui_vulkan: yakui_vulkan::YakuiVulkan,
+
     /// Miscellany
     character_model: Asset<GltfScene>,
 }
@@ -181,6 +184,17 @@ impl Draw {
 
             gfx.save_pipeline_cache();
 
+            let mut yakui_vulkan_options = yakui_vulkan::Options::default();
+            yakui_vulkan_options.render_pass = gfx.render_pass;
+            yakui_vulkan_options.subpass = 1;
+            let mut yakui_vulkan = yakui_vulkan::YakuiVulkan::new(
+                &yakui_vulkan::VulkanContext::new(device, gfx.queue, gfx.memory_properties),
+                yakui_vulkan_options,
+            );
+            for _ in 0..PIPELINE_DEPTH {
+                yakui_vulkan.transfers_submitted();
+            }
+
             let character_model = loader.load(
                 "character model",
                 super::GlbFile {
@@ -207,6 +221,8 @@ impl Draw {
 
                 buffer_barriers: Vec::new(),
                 image_barriers: Vec::new(),
+
+                yakui_vulkan,
 
                 character_model,
             }
@@ -235,6 +251,12 @@ impl Draw {
         let device = &*self.gfx.device;
         let state = &mut self.states[self.next_state];
         device.wait_for_fences(&[state.fence], true, !0).unwrap();
+        self.yakui_vulkan
+            .transfers_finished(&yakui_vulkan::VulkanContext::new(
+                device,
+                self.gfx.queue,
+                self.gfx.memory_properties,
+            ));
         state.in_flight = false;
     }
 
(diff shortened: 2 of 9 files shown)


patowen:
Draft status: This version currently crashes intermittently when switching materials with a Vulkan validation error. Based on my own testing, I believe that this can be resolved when https://github.com/SecondHalfGames/yakui/pull/155 is merged, and this PR points to the new branch. See comment thread of https://github.com/Ralith/hypermine/pull/391 for details.

Otherwise, I think it's ready for review.

client/src/graphics/draw.rs
```diff
@@ -253,9 +271,11 @@ impl Draw {
     ///
     /// Submits commands that wait on `image_acquired` before writing to `framebuffer`'s color
     /// attachment.
+    #[allow(clippy::too_many_arguments)] // Every argument is of a different type, making this less of a problem.
     pub unsafe fn draw(
         &mut self,
         mut sim: Option<&mut Sim>,
+        gui_state: &GuiState,
```
Ralith, line 278: Take a `yakui::paint::PaintDom` here, and store the `GuiState` at a higher level. That way the Vulkan rendering code can remain fully isolated from the abstract GUI logic, which will maintain better separation of concerns and probably make incremental compiles happier besides.

client/src/graphics/draw.rs
```diff
@@ -340,6 +379,10 @@ impl Draw {
         );
         timestamp_index += 1;
 
+        self.yakui_vulkan.transfers_finished(&yakui_vulkan_context);
```
Ralith, line 382: I think this morally belongs at the end of `fn wait`.

client/src/graphics/gui.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub fn new() -> Self {
+        GuiState { show_gui: true }
+    }
+
+    pub fn toggle_gui(&mut self) {
+        self.show_gui = !self.show_gui;
+    }
+}
+
+pub fn gui(gui_state: &GuiState, sim: &Sim) {
```
Ralith, line 21: nit: method on `GuiState`?

Ralith (changes requested):

patowen:
While testing this, I encountered a (potentially rare) warning log after I closed the application saying the following:
```WARN sendmsg error: Os { code: 10093, kind: Uncategorized, message: "Either the application has not called WSAStartup, or WSAStartup failed." }, Transmit: { destination: [::1]:60027, src_ip: None, enc: None, len: 202, segment_size: None }```

This seems to be a log from `quinn-udp::log_sendmsg_error`. I'm not sure whether it's important, since that might mean that Quinn tried to send a message after something was already finalized.

My `cargo.lock` has `quinn` at version 0.10.0 with `quinn-udp` at version 0.4.1 and `quinn-proto` at 0.10.5.

File: client/src/graphics/draw.rs (hunk 1 above)
Comment (patowen) [reply client/src/graphics/draw.rs]: I hadn't thought about the incremental compiles aspect, but I'm always for anything that keeps Hypermine compilation speeds fast.

File: client/src/graphics/draw.rs (hunk 2 above)
Comment (patowen) [reply client/src/graphics/draw.rs]: That does make a lot of sense.

**Comment by patowen** [comment]
I had renamed the `gui` module to `gui_state` because it only contained a struct called `GuiState` and its implementation. However, I'm starting to prefer the name `gui` again because it seems like a good place to put future things, such as custom widgets etc.

I expect this code might see a fair bit of churn in future PRs just due to the uncertainty of how all the UI elements will work together.

**Comment by Ralith** [comment]
> EDIT: Now that Quinn has been upgraded, this comment is likely irrelevant now. Thanks for upgrading it!

I don't think we changed anything that would be relevant to that error, FWIW. Sounds like something called [`WSACleanup`](https://learn.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-wsacleanup) unexpectedly, so if we ever see it reproducibly, that might be something to try to catch in a debugger. Out of scope of this PR, though.

**Review by Ralith — CHANGES_REQUESTED** [review]

File: client/src/graphics/draw.rs [hunk 4]
```diff
@@ -50,6 +50,10 @@ pub struct Draw {
 
     /// Miscellany
     character_model: Asset<GltfScene>,
+    /// Yakui Vulkan context
+    pub yakui_vulkan: yakui_vulkan::YakuiVulkan,
+    /// Yakui context
+    pub yak: yakui::Yakui,
```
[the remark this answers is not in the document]
patowen: Good catch. I had assumed clippy would have caught it, but it's public.

client/src/graphics/gui.rs
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    /// Toggles whether the GUI is shown
+    pub fn toggle_gui(&mut self) {
+        self.show_gui = !self.show_gui;
+    }
+
+    /// Prepare the GUI for rendering. This should be called between
+    /// Yakui::start and yakui::finish.
+    pub fn prepare_gui(&self) {
```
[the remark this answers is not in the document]
patowen: I like that idea. I'll go ahead with `run`.

client/src/graphics/base.rs
```diff
@@ -139,7 +139,11 @@ impl Base {
                             .queue_create_infos(&[vk::DeviceQueueCreateInfo::default()
                                 .queue_family_index(queue_family_index)
                                 .queue_priorities(&[1.0])])
-                            .enabled_extension_names(&device_exts),
+                            .enabled_extension_names(&device_exts)
+                            .push_next(
+                                &mut vk::PhysicalDeviceDescriptorIndexingFeatures::default()
```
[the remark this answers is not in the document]
patowen: I chose this because it best matched the documentation:
> This crate requires at least Vulkan 1.2 and a GPU with support for `VkPhysicalDeviceDescriptorIndexingFeatures.descriptorBindingPartiallyBound`

Strangely, based on https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceDescriptorIndexingFeatures.html and https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceVulkan12Features.html, both seem to be provided by Vulkan 1.2, suggesting they were introduced at the same time, so I'm not sure why this redundancy exists.

Ralith (approved):

PR #406, opened 2024-06-22
Blinker POC. Possible framework for constantly-updating entities.

blockmath:
Also divorced the ID generation code from `server::sim::Sim`.

Still todo for a full feature: making a generic system for entities to sync back to the client.

Future: `TickerEntity` should be the component associated with any constantly-updating entity, and the actual ticker-specific `Ticker` struct should be moved to be its own component, with the `ticker` field in `TickerEntity` denoting which `Component` the ticker is held in.


common/src/id_generator.rs
```diff
@@ -0,0 +1,33 @@
+use crate::EntityId;
+use fxhash::FxHashMap;
+use hecs::Entity;
+use rand::rngs::SmallRng;
+use rand::{Rng, SeedableRng};
+
+pub struct IDGenerator {
```
patowen, line 7: Since this struct contains the map from `EntityId` to `Entity`, it has become more than just an ID generator, so it likely deserves a different name, one that might help better define its purpose. Perhaps `EntityIdMap`?

However, I have second guesses about this kind of abstraction. Ideally, we should be able to spawn an entity, assign it an `EntityId`, and handle the `spawns`, `dirty_nodes`, and `graph_entities` fields all in a single method call, but the owner of that method would be `sim`, bringing us back to square one in terms of the borrow checker.

It pains me to suggest this, given my previous suggestion in Discord, but the best move for now might be to revert the creation of `id_generator`. I'm relatively unfamiliar with this part of the codebase and how to write good code for it, which is why my suggestions keep contradicting themselves.

server/src/sim.rs
```diff
@@ -272,6 +273,14 @@ impl Sim {
         let span = error_span!("step", step = self.step);
         let _guard = span.enter();
 
+        {
+            let mut ticker_query = self.world.query::<&mut TickerEntity>();
+            let ticker_iter = ticker_query.iter();
+            for (_entity, ticker_entity) in ticker_iter {
```
patowen, line 282: You can shorten this to
```rust
for (_entity, ticker_entity) in self.world.query::<&mut TickerEntity>().iter() {
```
in which case you'll also be able to get rid of the braces, since the scope will already be limited to the `for` loop.

common/src/ticker.rs
```diff
@@ -0,0 +1,35 @@
+use crate::Step;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TickerEntity {
+    pub last_ticked: Step,
```
patowen, line 6: This parameter is currently unused. If you get rid of it, you can get rid of this wrapper entirely and make `Ticker` the component instead.

common/src/ticker.rs
```diff
@@ -0,0 +1,35 @@
...
+impl Ticker {
+    pub fn new() -> Self {
+        Self { on: false }
+    }
+    pub fn tick(&mut self, _step: Step) {
+        self.on = !self.on;
+
+        // Just for testing
+        if self.on {
+            println!("Ticked ON!");
```
patowen, line 24: Hyprmine uses the `tracing` library for its logging. To apply that here, you can use
```rust
tracing::info!("Ticked ON!");
```

common/src/ticker.rs
```diff
@@ -0,0 +1,35 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct TickerEntity {
+    pub last_ticked: Step,
+    pub ticker: Ticker,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Ticker {
```
patowen, line 11: Due to the idea of using different components for different types of objects (even ones with a similar logic of doing something every frame), I think the right move is to be consistent in using the term "ticker" everywhere or "blinker" everywhere (probably the latter).

common/src/ticker.rs
```diff
@@ -0,0 +1,35 @@
...
+    on: bool,
+}
+
+impl Ticker {
+    pub fn new() -> Self {
+        Self { on: false }
+    }
+    pub fn tick(&mut self, _step: Step) {
```
patowen, line 19: A method on `Ticker` _seems_ like the best place to put logic for entities with the `Ticker` component. However, I worry that this comes from an object-oriented way of thinking. With ECS, components are usually thought of as "plain-old data", so the logic that manipulates the `Ticker` component would want to live outside it.

This logic could live directly in `server::sim`, or it could live in a separate module (much like `common::character_controller`, although that is somewhat special because both the server and the client have to run its logic).

It's likely that the `self.on = !self.on` line of code would live in the same function as the `for` loop that loops through all the tickers.

server/src/sim.rs
```diff
@@ -30,14 +31,15 @@ use crate::postcard_helpers::{self, SaveEntity};
 
 pub struct Sim {
     cfg: Arc<SimConfig>,
-    rng: SmallRng,
     step: Step,
-    entity_ids: FxHashMap<EntityId, Entity>,
+    pub rng: SmallRng,
+    pub entity_ids: FxHashMap<EntityId, Entity>,
```
patowen, line 36: `rng` and `entity_ids` don't need to be `pub` here. I think the order and visibility of these fields got changed as part of the revert I suggested earlier.

server/src/sim.rs
```diff
@@ -51,13 +53,14 @@ pub struct Sim {
...
         let mut result = Self {
-            rng: SmallRng::from_entropy(),
             step: 0,
+            rng: SmallRng::from_entropy(),
             entity_ids: FxHashMap::default(),
             world: hecs::World::new(),
             graph: Graph::new(cfg.chunk_size),
             preloaded_voxel_data: FxHashMap::default(),
             spawns: Vec::new(),
+            pending_ticker_spawns: Vec::new(),
```
patowen, line 63: As `pending_ticker_spawns` is only used in the `step` function, let's make it a local variable.

server/src/sim.rs
```diff
@@ -412,7 +437,7 @@ impl Sim {
         (spawns, delta)
     }
 
-    fn new_id(&mut self) -> EntityId {
+    pub fn new_id(&mut self) -> EntityId {
```
patowen, line 440: This function doesn't need to be `pub`.

blockmath:
Yeah, most of the weird `pub` stuff probably got that way from the revert.

client/src/local_character_controller.rs
```diff
@@ -3,7 +3,7 @@ use common::{math, proto::Position};
 pub struct LocalCharacterController {
     /// The last extrapolated inter-frame view position, used for rendering and gravity-specific
     /// orientation computations
-    position: Position,
+    pub position: Position,
```
patowen, line 6: As an alternative to making this public, you can also use the `oriented_position()` method. That method might have fewer surprises, such as making sure the orientation of the blinker you place matches the direction the player is facing.

patowen (changes requested):
This is looking pretty good! I'm optimistic that this will be able to be merged relatively soon, although, especially since I'm not that confident in this part of the codebase, it will need a review from @Ralith before it's ready to merge.

common/src/ticker.rs
blockmath: Done!

common/src/ticker.rs
blockmath: Good to know. Changed.

common/src/ticker.rs
blockmath: Got it, `Ticker` is now `Blinker`.

common/src/ticker.rs
blockmath: You're right, it's very C++ pilled. Will do.

server/src/sim.rs
blockmath: Right, I don't know why I made that a field.

client/src/local_character_controller.rs
blockmath: Nice. Will do.

blockmath:
All right, the blinker now actually syncs back to the client. I have no idea how I would go about actually rendering it, though...

common/src/ticker.rs
patowen, line 1: Don't forget to rename this file to `blinker.rs`.

I believe the only other place `ticker` is still mentioned is `for (position, ticker) in pending_blinker_spawns` in `server::sim`.

Also, don't forget to run `cargo fmt` (this file and `proto.rs` need updating).

server/src/sim.rs
```diff
@@ -30,8 +31,8 @@ use crate::postcard_helpers::{self, SaveEntity};
 
 pub struct Sim {
     cfg: Arc<SimConfig>,
-    rng: SmallRng,
     step: Step,
+    rng: SmallRng,
```
patowen, line 35: To shrink the size of this diff, `rng` should be before `step` here and in the constructor below.

patowen (changes requested):
All that's left from me are minor nits.

patowen:
> All right, the blinker now actually syncs back to the client. I have no idea how I would go about actually rendering it, though...

Yeah, that's going to be rather tricky. The rendering code doesn't have many examples of things being rendered besides the world grid and other players. It's possible that will need to be a separate PR.

blockmath:
I agree, that's why I'm deciding to finish now rather than waiting to do that.

common/src/blinker.rs
```diff
@@ -0,0 +1,18 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Blinker {
```
patowen, line 4: I don't have strong opinions on this, but there are some options (not mutually exclusive) for what to do with this file:
- Most components live in `proto.rs`. This _can_ live there, too, but it doesn't have to. I have a PR that makes `Material` a component, so this isn't a hard and fast rule.
- Since all fields are public, having a constructor is optional, as it can be constructed manually in `server::sim` instead.

client/src/sim.rs
```diff
@@ -279,6 +290,20 @@ impl Sim {
         }
     }
 
+    fn update_blinker_state(&mut self, id: EntityId, new_blinker: &Blinker) {
+        match self.entity_ids.get(&id) {
+            None => debug!(%id, "blinker state update for unknown entity"),
+            Some(&entity) => match self.world.get::<&mut Blinker>(entity) {
+                Ok(mut blinker) => {
+                    blinker.on = new_blinker.on;
```
patowen, line 298: This can be changed to
```rust
*blinker = new_blinker.clone();
```
to avoid potentially forgetting to update this if adding another field to `Blinker`.

patowen (approved):

server/src/sim.rs
```diff
@@ -350,6 +368,15 @@ impl Sim {
             self.dirty_nodes.insert(position.node);
         }
 
+        for (position, ticker) in pending_blinker_spawns {
```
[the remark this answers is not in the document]
blockmath: Already fixed.

common/src/blinker.rs
blockmath: Noted, those are good patterns.

client/src/sim.rs
blockmath: That's a good idea. Done.

server/src/sim.rs
```diff
@@ -370,6 +386,15 @@ impl Sim {
             self.dirty_nodes.insert(position.node);
         }
 
+        for (position, blinker) in pending_blinker_spawns {
```
patowen, line 389: One of the changes I made in my PR provided a more succinct way of accomplishing this:
```rust
for (position, blinker) in pending_blinker_spawns {
    self.spawn((position, blinker));
}
```

blockmath: Cool, thanks!

patowen (approved):

PR #407, opened 2024-06-23
Store all `Material` types in an array

patowen:
The ability to walk through the list of available materials will likely come in handy in many situations. For me, it came up when I thought about implementing scrolling through different items in the inventory, which is why I made that change now, although I later decided to defer that feature until later.

Rather than throwing away the code I wrote, I believe it's probably best to just put it in its own PR.

Probably the most controversial decision here is calling it `Material::ALL`. I'm open to other suggestions.


common/src/world.rs
```diff
@@ -50,55 +50,56 @@ pub enum Material {
 
 impl Material {
     pub const COUNT: usize = 40;
+
+    pub const ALL: [Self; Self::COUNT] = [
```
Ralith, line 54: nit: `VALUES`?

Ralith (approved):
Seems reasonable to me.

common/src/world.rs
patowen: That's far better than `ALL`. Thanks!

PR #412, opened 2024-06-29
Fix issues with Windows package workflow

patowen:
An update to the shaderc-sys crate from 0.7 to 0.8 added a requirement to be able to view the vk.xml file to gather information about the installed version of Vulkan. To fix the workflow, we need to include that file in the list of files being extracted.

In addition, the cache action was upgraded from v2 to v4 because the old one was broken with errors like the following:
```
C:/VulkanSDK/1.3.231.1/: Can't create '\\\\?\\C:\\VulkanSDK\\1.3.231.1'
```

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -17,7 +17,7 @@ jobs:
           lfs: true
       - name: Cache Vulkan SDK
         id: cache-vulkan
-        uses: actions/cache@v2
+        uses: actions/cache@v4
         with:
           path: ${{ env.VULKAN_SDK }}
           key: vulkan-${{ env.VULKAN_VERSION }}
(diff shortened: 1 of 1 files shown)


patowen:
Since the Windows package workflow doesn't run by default on PRs, there's not a good way to test this fix here until it is merged. However, I have tested it myself, and the results will be temporarily available at <https://github.com/patowen/hypermine/actions>.

.github/workflows/package.yml
```diff
@@ -17,15 +17,15 @@ jobs:
...
+        uses: actions/cache@v4
         with:
           path: ${{ env.VULKAN_SDK }}
           key: vulkan-${{ env.VULKAN_VERSION }}
       - name: Install Vulkan SDK
         if: steps.cache-vulkan.outputs.cache-hit != 'true'
         run: |
           Invoke-WebRequest -Uri "https://sdk.lunarg.com/sdk/download/${{ env.VULKAN_VERSION }}/windows/VulkanSDK-${{ env.VULKAN_VERSION }}-Installer.exe" -OutFile vulkan.exe
-          7z x -aoa vulkan.exe "-o$env:VULKAN_SDK" Lib
+          7z x -aoa vulkan.exe "-o$env:VULKAN_SDK" Lib share/vulkan/registry/vk.xml
```
Ralith, line 28: Should we just extract everything? Might avoid future annoyance.

Ralith (approved):

.github/workflows/package.yml
patowen: I'm hopeful that this is just a one-off. The fewer files we extract, the quicker (and less resource-heavy) the GitHub actions will be. I checked on my own machine, and the extracted version is 809 MB if we include everything but only 202 MB if we stick with this. (It would be less we need to cache as well.)

I have a weak preference towards just extracting what we need, but I'm happy to switch to extracting everything if you would prefer.

PR #413, opened 2024-06-30
Simplify and document expansion functions of Graph

patowen:
This PR accomplishes a few things, all in the spirit of hopefully reducing the mental load of understanding all the algorithms that allow `Graph` to grow:
- The term "parent" no longer depends on the graph's generation order. It now refers to the node along the first descender in enum order.
- Uses of the term "parent" and "child" were removed whenever they didn't refer to the above meaning of "parent". This includes renaming `insert_child` to `insert_neighbor` and renaming `populate_shorter_neighbors_of_child` to `populate_shorter_neighbors_of_subject` (where "subject" is a term that was introduced internally for "the neighbor node we want to insert" to try to improve clarity).
- An unused code path was removed in `ensure_neighbor` to cover the impossible scenario of a not-yet-generated shorter neighbor. Removing this code-path turned `ensure_neighbor` into a one-liner.
- The most conceptually difficult function in the algorithm, `populate_shorter_neighbors_of_subject` is now documented in detail. For simplification purposes, it was also altered to return all shorter neighbors, including the passed in argument (as it previously excluded this argument, resulting in extra logic for `insert_neighbor`).

### DIFF
--- a/common/src/graph.rs
+++ b/common/src/graph.rs
@@ -159,62 +159,42 @@ impl Graph {
     /// Ensures that the neighbour node at a particular side of a particular node exists in the graph,
     /// as well as the nodes from the origin to the neighbour node.
     pub fn ensure_neighbor(&mut self, node: NodeId, side: Side) -> NodeId {
-        let v = &self.nodes[&node];
-        if let Some(x) = v.neighbors[side as usize] {
-            // Neighbor already exists
-            return x;
-        }
-
-        let neighbor_is_further = |parent_side| {
-            (side != parent_side && !side.adjacent_to(parent_side))
-                || !self.is_near_side(v.parent().unwrap(), side)
-        };
-
-        // Create a new neighbor
-        if v.parent_side.map_or(true, neighbor_is_further) {
-            // New neighbor will be further from the origin
-            return self.insert_child(node, side);
-        }
-
-        // Neighbor is closer to the origin; find it, backfilling if necessary
-        let x = self.nodes[&v.parent().unwrap()].neighbors[side as usize].unwrap();
-        let parent_side = v.parent_side.unwrap();
-        let neighbor = self.ensure_neighbor(x, parent_side);
-        self.link_neighbors(node, neighbor, side);
-        neighbor
+        self.nodes[&node].neighbors[side as usize]
+            .unwrap_or_else(|| self.insert_neighbor(node, side))
     }
 
     /// Whether `node`'s neighbor along `side` is closer than it to the origin
-    fn is_near_side(&self, node: NodeId, side: Side) -> bool {
+    fn is_descender(&self, node: NodeId, side: Side) -> bool {
         let v = &self.nodes[&node];
         v.neighbors[side as usize].map_or(false, |x| self.nodes[&x].length < v.length)
     }
 
-    pub fn insert_child(&mut self, parent: NodeId, side: Side) -> NodeId {
-        // Always create shorter nodes first so that self.nodes always puts parent nodes before their child nodes, enabling
-        // graceful synchronization of the graph
-        let shorter_neighbors = self.populate_shorter_neighbors_of_child(parent, side);
+    /// Inserts the neighbor of the given node at the given side into the graph, ensuring that all
+    /// shorter nodes it depends on are created first.
+    pub fn insert_neighbor(&mut self, node: NodeId, side: Side) -> NodeId {
+        // To help improve readability, we use the term "subject" to refer to the not-yet-created neighbor node, since the term.
+        // "neighbor" is already used in many other places.
 
-        // Select the side along the canonical path from the origin to this node. Future work: make
-        // this the parent to reduce the number of concepts.
-        let (path_side, predecessor) = shorter_neighbors
-            .clone()
-            .chain(Some((side, parent)))
-            .min_by_key(|&(side, _)| side)
-            .unwrap();
+        // An assumption made by this function is that `side` is not a descender of `node`. This is a safe assumption to make
+        // because a node cannot be constructed before its shorter neighbors. The call to `populate_shorter_neighbors_of_neighbor`
+        // guarantees this.
+        let shorter_neighbors_of_subject = self.populate_shorter_neighbors_of_subject(node, side);
+
+        // Select the side along the canonical path from the origin to this node. This is guaranteed
+        // to be the first entry of the `shorter_neighbors_of_subject` iterator.
+        let (parent_side, parent) = shorter_neighbors_of_subject.clone().next().unwrap();
         let mut hasher = Hasher::new();
-        hasher.update(&predecessor.0.to_le_bytes());
-        hasher.update(&[path_side as u8]);
+        hasher.update(&parent.0.to_le_bytes());
+        hasher.update(&[parent_side as u8]);
         let mut xof = hasher.finalize_xof();
         let mut hash = [0; 16];
         xof.fill(&mut hash);
         let id = NodeId(u128::from_le_bytes(hash));
 
-        let length = self.nodes[&parent].length + 1;
+        let length = self.nodes[&node].length + 1;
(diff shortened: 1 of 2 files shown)


patowen:
I believe some of the comments I've written don't make sense, although it's hard to tell in what way they don't make sense and how to improve them, so I hope that peer review will help here.

common/src/graph.rs
```diff
@@ -159,62 +159,39 @@ impl Graph {
...
+    /// shorter nodes it depends on are created first.
+    pub fn insert_neighbor(&mut self, node: NodeId, side: Side) -> NodeId {
+        // An assumption made by this function is that `side` is not a descender of `node`. This is a safe assumption to make
+        // because a node cannot be constructed before its shorter neighbors. The call to `populate_shorter_neighbors_of_neighbor`
+        // guarantees this.
+        let shorter_neighbors_of_neighbor = self.populate_shorter_neighbors_of_neighbor(node, side);
+
+        // Select the side along the canonical path from the origin to this node. This is guaranteed
+        // to be the first entry of the `shorter_neighbors_of_neighbor` iterator.
+        let (parent_side, parent) = shorter_neighbors_of_neighbor.clone().next().unwrap();
```
Ralith, line 182:
```suggestion
        let (parent_side, parent) = shorter_neighbors_of_neighbor.next().unwrap();
```

common/src/graph.rs
```diff
@@ -231,27 +208,49 @@ impl Graph {
...
+            } else {
+                // The `candidate_descender_side` is not a descender of the subject, so no action is necessary.
             }
-            let x = self.nodes[&parent].neighbors[neighbor_side as usize].unwrap();
-            let neighbor = self.ensure_neighbor(x, parent_side);
-            neighbors[count] = Some((neighbor_side, neighbor));
-            count += 1;
         }
-        (0..2).filter_map(move |i| neighbors[i])
+        (0..3).filter_map(move |i| shorter_neighbors_of_subject[i])
```
Ralith, line 253: While we're revisiting this, might as well take advantage of the `[T]::into_iter` that's been introduced since it was written, just to be a bit less weird:
```suggestion
        shorter_neighbors_of_subject.into_iter().filter_map(|x| x)
```

common/src/graph.rs
```diff
@@ -231,27 +208,49 @@ impl Graph {
...
+    /// Ensure all shorter neighbors of a not-yet-created neighbor node exist, and return them (including the given node)
+    /// in the form of ordered pairs containing the side they share with this not-yet-created neighbor node, and their node ID.
+    /// These ordered pairs will be sorted by side, based on enum order.
+    fn populate_shorter_neighbors_of_neighbor(
         &mut self,
-        parent: NodeId,
-        parent_side: Side,
+        node: NodeId,
+        side: Side,
     ) -> impl Iterator<Item = (Side, NodeId)> + Clone {
```
Ralith, line 218: Do we still need `+ Clone` on this?

common/src/graph.rs
```diff
@@ -159,62 +159,39 @@ impl Graph {
...
         let mut xof = hasher.finalize_xof();
         let mut hash = [0; 16];
         xof.fill(&mut hash);
         let id = NodeId(u128::from_le_bytes(hash));
 
-        let length = self.nodes[&parent].length + 1;
+        let length = self.nodes[&node].length + 1;
         self.nodes
-            .insert(id, NodeContainer::new(Some(side), length));
-        self.link_neighbors(id, parent, side);
```
Ralith, line 216: Why was this not necessary?

Ralith: Oh, I see.

common/src/graph.rs
```diff
@@ -159,62 +159,39 @@ impl Graph {
...
-            .chain(Some((side, parent)))
-            .min_by_key(|&(side, _)| side)
-            .unwrap();
+    /// Inserts the neighbor of the given node at the given side into the graph, ensuring that all
+    /// shorter nodes it depends on are created first.
+    pub fn insert_neighbor(&mut self, node: NodeId, side: Side) -> NodeId {
+        // An assumption made by this function is that `side` is not a descender of `node`. This is a safe assumption to make
+        // because a node cannot be constructed before its shorter neighbors. The call to `populate_shorter_neighbors_of_neighbor`
+        // guarantees this.
+        let shorter_neighbors_of_neighbor = self.populate_shorter_neighbors_of_neighbor(node, side);
```
Ralith, line 178: I found this variable name confusing. I understood it to mean "shorter neighbors of the neighbor [of the new node]", but in fact it seems to be "shorter neighbors [of the new node]". Maybe standardize on `subject` like below?

Ralith:
Great improvement overall! I don't think I ever actually understood this logic, so this clarity is striking.

common/src/graph.rs
patowen: I'm not surprised it's confusing. It's been a while, so looking at this PR again with fresh eyes should hopefully help me think of better variable names here.

common/src/graph.rs
Ralith: On second thought, do we need to return an iterator at *all*? It seems like we only care about the parent, so this could probably be simplified to just return `(Side, NodeId)` directly.

patowen: We care about all elements of this iterator because of the
```rust
for (side, neighbor) in shorter_neighbors_of_neighbor
```
loop near the bottom of `insert_neighbor`. This is also why the iterator needs to be cloned, because it's first used to find the "parent", and then it is used to perform all the linking of neighbors.

common/src/graph.rs
patowen: Thanks! That does seem nice. Clippy further simplifies `filter_map(|x| x)` to `flatten()`, which seems a bit weird to me, but that's only because I consider the concept of an `Option` being a type of iterator to be a bit weird conceptually. I'll trust the designers of Rust that this is okay.

Ralith: Think of it as a container of 0 or 1 items, just like an array is a container of N items, or a Vec is a container of an arbitrary number of items.

patowen: Oh, yeah. I do think of it that way. It's just not the most natural way for me.

common/src/graph.rs
patowen: Thinking on this again, I agree with you that standardizing on the name `subject` makes sense, as long as that naming decision doesn't leak outside the `graph` module. I'll go ahead with your suggestion.

patowen:
Ah, the "Compare" button isn't as useful this time because it also includes all the changes from `master` that were included. I'll list out the changes I made:

- At the top of `insert_neighbor`, I added the comment
```rust
// To help improve readability, we use the term "subject" to refer to the not-yet-created neighbor node, since the term.
// "neighbor" is already used in many other places.
```
- In `insert_neighbor, I renamed `shorter_neighbors_of_neighbor` to `shorter_neighbors_of_subject`.
- I changed the name of the function `populate_shorter_neighbors_of_neighbor` with `populate_shorter_neighbors_of_subject` and updated the function header comment to explain what "subject" means.
- I removed the comment at the top of the `populate_shorter_neighbors_of_subject` function explaining the use of "subject", as the concept is no longer introduced there.
- I added the following addendum to `populate_shorter_neighbors_of_subject`'s long comment:
```rust
// Note that this shorter neighbor might not yet be in the graph, which is why we call `ensure_neighbor`
// here instead of `neighbor`. This means that nodes in the graph will be recursively created as necessary.
```
- I used `shorter_neighbors_of_subject.into_iter().flatten()` as suggested.

patowen:
I realize that my note about calling `ensure_neighbor` instead of `neighbor` was unnecessarily verbose for what is already a long and dense explanation, so I replaced it with the `(and recursively create if needed)` parenthetical, which accomplishes the same thing in much fewer words.

Ralith (approved):

PR #414, opened 2024-06-30
Load character entities from the save file

patowen:
This PR sets up the initial implementation of loading entity data from a save file by loading player characters. In single-player, this has the effect of having the game remember where you were when you close it. If you change your name, you will spawn in as a new character.

The following are not currently implemented:
* Saving client-authoritative orientation data (the orientation quaternion)
* Only loading entities in nearby chunks (Currently, everything is loaded.)
* Saving the inventory

### DIFF
--- a/common/src/proto.rs
+++ b/common/src/proto.rs
@@ -66,7 +66,7 @@ pub struct Command {
     pub orientation: na::UnitQuaternion<f32>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct CharacterInput {
     /// Relative to the character's current position, excluding orientation
     pub movement: na::Vector3<f32>,
@@ -113,4 +113,13 @@ pub struct Character {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct Inventory {
     pub contents: Vec<EntityId>,
+}
+
+pub mod connection_error_codes {
+    use quinn::VarInt;
+
+    pub const CONNECTION_LOST: VarInt = VarInt::from_u32(0);
+    pub const STREAM_ERROR: VarInt = VarInt::from_u32(1);
+    pub const BAD_CLIENT_COMMAND: VarInt = VarInt::from_u32(2);
+    pub const NAME_CONFLICT: VarInt = VarInt::from_u32(3);
 }
--- a/server/src/lib.rs
+++ b/server/src/lib.rs
@@ -16,7 +16,11 @@ use tokio::sync::mpsc;
 use tokio_stream::wrappers::{IntervalStream, ReceiverStream};
 use tracing::{debug, error, error_span, info, trace};
 
-use common::{codec, proto, SimConfig};
+use common::{
+    codec,
+    proto::{self, connection_error_codes},
+    SimConfig,
+};
 use input_queue::InputQueue;
 use save::Save;
 use sim::Sim;
@@ -138,9 +142,10 @@ impl Server {
         }
         for client_id in overran {
             error!("dropping slow client {:?}", client_id.0);
-            self.clients[client_id]
-                .conn
-                .close(1u32.into(), b"client reading too slowly");
+            self.clients[client_id].conn.close(
+                connection_error_codes::STREAM_ERROR,
+                b"client reading too slowly",
+            );
             self.cleanup_client(client_id);
         }
 
@@ -161,7 +166,14 @@ impl Server {
             ClientEvent::Hello(hello) => {
                 assert!(client.handles.is_none());
                 let snapshot = Arc::new(self.sim.snapshot());
-                let (id, entity) = self.sim.spawn_character(hello);
+                let Some((id, entity)) = self.sim.activate_or_spawn_character(&hello) else {
+                    error!("could not spawn {} due to name conflict", hello.name);
+                    client
+                        .conn
+                        .close(connection_error_codes::NAME_CONFLICT, b"name conflict");
+                    self.cleanup_client(client_id);
+                    return;
+                };
                 let (ordered_send, ordered_recv) = mpsc::channel(32);
                 ordered_send.try_send(snapshot).unwrap();
                 let (unordered_send, unordered_recv) = mpsc::channel(32);
@@ -183,7 +195,9 @@ impl Server {
             }
             ClientEvent::Lost(e) => {
                 error!("lost: {:#}", e);
-                client.conn.close(0u32.into(), b"");
+                client
+                    .conn
+                    .close(connection_error_codes::CONNECTION_LOST, b"");
                 self.cleanup_client(client_id);
             }
             ClientEvent::Command(cmd) => {
@@ -199,7 +213,7 @@ impl Server {
 
     fn cleanup_client(&mut self, client: ClientId) {
         if let Some(ref x) = self.clients[client].handles {
-            self.sim.destroy(x.character);
+            self.sim.deactivate_character(x.character);
         }
         self.clients.remove(client);
     }
@@ -249,7 +263,10 @@ async fn drive_recv(
                     // we want to drop the client. We close the connection, which will cause `drive_recv` to
                     // return eventually.
                     tracing::error!("Error when parsing unordered stream from client: {e}");
-                    connection.close(2u32.into(), b"could not process stream");
+                    connection.close(
+                        connection_error_codes::BAD_CLIENT_COMMAND,
+                        b"could not process stream",
+                    );
                 }
                 Ok(msg) => {
                     let _ = send.send((id, ClientEvent::Command(msg))).await;
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -1,7 +1,8 @@
 use std::sync::Arc;
 
 use anyhow::Context;
-use common::dodeca::Vertex;
+use common::dodeca::{Side, Vertex};
+use common::math::MIsometry;
 use common::node::VoxelData;
 use common::proto::{BlockUpdate, Inventory, SerializedVoxelData};
 use common::world::Material;
@@ -11,6 +12,7 @@ use hecs::{DynamicBundle, Entity, EntityBuilder};
 use rand::rngs::SmallRng;
 use rand::{Rng, SeedableRng};
 use save::ComponentType;
+use serde::{Deserialize, Serialize};
 use tracing::{error, error_span, info, trace};
 
 use common::{
@@ -69,6 +71,9 @@ impl Sim {
             .load_all_voxels(save)
             .expect("save file must be of a valid format");
         result
+            .load_all_entities(save)
+            .expect("save file must be of a valid format");
+        result
     }
 
     pub fn save(&mut self, save: &mut save::Save) -> Result<(), save::DbError> {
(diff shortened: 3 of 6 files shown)


server/src/sim.rs
```diff
@@ -109,6 +115,84 @@ impl Sim {
...
+        &mut self,
+        read: &mut save::Reader,
+        node: NodeId,
+        save_entity: SaveEntity,
+    ) -> anyhow::Result<()> {
+        let entity_id = EntityId::from_bits(u64::from_le_bytes(save_entity.entity));
+        let mut entity_builder = EntityBuilder::new();
+        entity_builder.add(entity_id);
+        for (component_type, component_bytes) in save_entity.components {
+            if component_type == ComponentType::Position as u64 {
```
Ralith, line 143: nit: Use `ComponentType::try_from(component_type as i32)` then `match` on the result, and maybe abstract out a `load_component` helper for readability?

server/src/sim.rs
```diff
@@ -290,7 +389,12 @@ impl Sim {
         let _guard = span.enter();
 
         // Extend graph structure
-        for (_, (position, _)) in self.world.query::<(&mut Position, &mut Character)>().iter() {
+        for (_, (position, character)) in
+            self.world.query::<(&mut Position, &mut Character)>().iter()
+        {
+            if !character.state.active {
```
Ralith, line 395: I'm concerned about having to repeat this frequently, and risk bugs when it's forgotten. What if we instead introduced a new component `struct InactiveCharacter(Character);` and converted back and forth as needed?

server/src/sim.rs
```diff
@@ -85,6 +88,9 @@ impl Sim {
         let mut tx = save.write()?;
         let mut writer = tx.get()?;
         for (_, (pos, ch)) in self.world.query::<(&Position, &Character)>().iter() {
+            if !ch.state.active {
```
Ralith, line 91: Why shouldn't we save inactive characters?

Ralith (changes requested):

server/src/sim.rs
patowen: While it's good to save the entity itself to avoid it disappearing next time its node is saved, we don't need to save the inactive character's path from the origin every frame because the character doesn't move, so we just end up overwriting data with itself. Since paths could get pretty long, this seemed potentially worth optimizing.

It's at least worth a code comment, but I don't feel strongly here, so I am happy to revert this part (although it would result in more code if I do the `InactiveCharacter` wrapping).

server/src/sim.rs
patowen: We don't currently have any machinery for adding and removing components in messages sent from the server to the client. However, the client doesn't have any reason to know the locations of inactive characters, so I could probably get away with just adding the entire entity to the "despawns" list when it gets deactivated. I'll at least try it, as I can always revert if necessary.

patowen: In `snapshot_node`, I need to include the `Character` regardless of whether it's active or not. I don't think there's a good way of turning a `hecs::Ref<InactiveCharacter>` into a `hecs::Ref<Character>`, so it's not clear how I would want to implement `snapshot_node`. Calling another function for
```rust
components.push((ComponentType::Name as u64, ch.name.as_bytes().into()));
```
seems like overkill, but repeating all the code for the two component types seems too easy to get desynchronized.

There are risks with either implementation, since an inactive character might be ignored when it shouldn't be or treated as active when it shouldn't be. The simplicity of the boolean is enticing, and I'd probably need a bit more convincing to switch over to using another component.

server/src/sim.rs
patowen: Actually, thinking about this again, this feels like a premature optimization. I could imagine later wanting to add a feature where an admin can move logged-out players around, and forgetting about this could result in unexpected desynchronization. I'll get rid of it.

server/src/sim.rs
patowen: Thanks! I hadn't realized that `ComponentType::try_from` was already written. I assume this is handled by `::prost::Enumeration`.

server/src/sim.rs
Ralith: > I could probably get away with just adding the entire entity to the "despawns" list when it gets deactivated

That makes sense to me. No need for clients to know anything about inactive characters.

>  I don't think there's a good way of turning a hecs::Ref<InactiveCharacter> into a hecs::Ref<Character>

[`Ref::map`](https://docs.rs/hecs/latest/hecs/struct.Ref.html#method.map). With a little more finagling you could also operate in terms of `&T` instead of `Ref<T>`.

A more ECS-idiomatic solution would be to have a dedicated `Name` component that both disabled and active characters share.

> repeating all the code for the two component types seems too easy to get desynchronized.

Notwithstanding the above solutions, better to repeat `components.push(...)` once than `if !character.state.active { ... }` an unbounded number of times.

patowen: That's a fair point. It's good to know that `Ref::map` exists. I might have to think more about this one, as I would like to try to get a self-consistent mental model that makes sense for this. It's a bit tricky because the server runs some kind of "simulation", but logging in and logging out causes players to temporarily disappear with the ability to reappear at the same spot later on with the same status.

In some games, like Satisfactory, logged out players still show up in the game world in their inactive state, in which case clients would want to know about them. That might be worth considering, in the spirit of allowing experimentation when we don't have a solid plan for gameplay.

I'm also second-guessing my decision to have `active` be a member of `Character`. It's possible we would want to replace a different component instead. For instance, having an `InactiveCharacter` component doesn't immediately resolve potential pitfalls because if we, for instance, added a `MinimapIcon` component (designed to mark things that should appear on the minimap) to the player, we would still need to remember to remove that component when players log out.

Another tricky thing is that, when saving, what we actually want to save is what the state of the simulation would be if everybody logged out, so the more differences there are between logged-in and logged-out players, the more potential room for error there is in the saving logic.

All in all, this is a tricky problem that I haven't really figured out what to do for. One part of me is thinking that logged-in and logged-out players should be treated as similarly as possible, leaving it up to the system logic to filter out inactive entities (possibly with an `Inactive` marker component), while the other part of me is thinking that logged-out players shouldn't be returned from ECS queries unless you are actively looking for them (which might involve hoisting them out of ECS entirely, although the question remains for what to do about the items they're holding).

If anything leans me one way, it would probably be towards the former due to the fact that we might want to do something with a logged-out player, such as having an admin move them or having the model rendered.

I'm definitely open to suggestions, though.

Ralith: I think removing any unwanted components on log-out is the idiomatic approach. That's a single code location to worry about and has highly predictable effects. Relying on every single system to understand and skip "inactive" entities has much more far reaching implications, and removing them entirely would make save/load more of a special case, especially for components (like `Inventory`) that might otherwise share logic with other entities.

patowen: Which components are unwanted? I want `Position` because we need to remember where they were for when we log back in. I want `Character` because preserving the character state would allow things like velocity to be preserved if someone logs out. I want `Inventory` because players keep there items when logged out. The only component that is definitely unneeded is `CharacterInput`, since nobody is inputting anything when logged out.

However, if I just remove the `CharacterInput` component, other systems that don't care about that component would still need to understand and find some way to skip "inactive" entities.

One workaround you had suggested previously is an `InactiveCharacter`, but I'm not sure why it's okay to keep a real `Inventory` component when the `Character` component needs to be replaced.

That being said, I do think removing unneeded components is likely part of the solution. However, I still don't think there's a clean way to ensure that systems will never have to be aware of this "active" concept sometimes.

Since a lot of these decisions feel arbitrary, I think I might be more comfortable if I better understand the end goal. Why is the component called `Character` and not `Player`? Is the plan to reuse it for NPCs? Is the `name` field of the `Character` component only for players? When controlling an NPC's AI, if it interacts with other characters, what components will it care about?

I would like to have a good idea of some of those questions before I feel comfortable splitting the `Character` component. My understanding is that the idiomatic approach to ECS is for components to represent some kind of attribute the entity holds, and I think I'll need to know what those attributes are before I make too many changes. It doesn't help that after a player logs out, they're still conceptually a "character", just in some kind of stasis.

Honestly, requiring systems to be aware of the "active" state seems easier to refactor down the road (after we are actually seeing that repeated logic in enough places) than trying to design the components to make "active" and "inactive" come naturally. If you would prefer, I could still make the `InactiveCharacter` component, but I won't have a good intuition for it, so I wouldn't be able to design around it with any confidence.

Ralith: > Which components are unwanted?

Whichever ones are required by the logic you don't want to run.

> I want Character because preserving the character state would allow things like velocity to be preserved if someone logs out.

You want the information, not the compositional identity. Hence `struct InactiveCharacter(Character);`.

> other systems that don't care about that component would still need to understand and find some way to skip "inactive" entities.

If you remove the components they need, then you don't need to modify the other systems. What systems are you concerned about specifically?

> Why is the component called Character and not Player?

A player is a person. A character is an entity a player might control in a game world.

> Is the plan to reuse it for NPCs?

No specific plan here.

> Is the name field of the Character component only for players?

It seems like that might be convenient, so why not?

> I would like to have a good idea of some of those questions before I feel comfortable splitting the Character component.

There's only two fields; not a whole lot of splitting to agonize over. As a general rule of thumb, data should be gathered into a single component when it is always needed together, and shouldn't be when it isn't. When in doubt, keep things together for simplicity.

> requiring systems to be aware of the "active" state seems easier to refactor down the road

Most decisions at this stage are reasonably easy to refactor. They'll become less so as precedent sets in.

> after we are actually seeing that repeated logic in enough places

How many places is enough? We're already at four, counting the renderer.

patowen: I think the best move here is for me to trust your judgement. If it's easy to refactor whatever the decision is, then I think I should stop agonizing over this and just make `InactiveCharacter`.

It might make sense to still transmit it to the client to make debugging easier by keeping the server and client as in-sync as possible. However, it might also be best to wait until a bug like this occurs that we want to debug. "Hidden information" in general seems like something we could want to support.

patowen: Now that I've added the `InactiveCharacter` struct, I'll mark this as resolved.

server/src/lib.rs
```diff
@@ -161,7 +161,12 @@ impl Server {
             ClientEvent::Hello(hello) => {
                 assert!(client.handles.is_none());
                 let snapshot = Arc::new(self.sim.snapshot());
-                let (id, entity) = self.sim.spawn_character(hello);
+                let Some((id, entity)) = self.sim.activate_or_spawn_character(&hello) else {
+                    error!("could not spawn {} due to name conflict", hello.name);
+                    client.conn.close(3u32.into(), b"name conflict");
```
Ralith, line 166: "3" is an opaque magic number. Introduce a named constant in common::proto instead.

server/src/sim.rs
```diff
@@ -109,6 +112,104 @@ impl Sim {
...
+                let column_slice: [f32; 16] = postcard::from_bytes(&component_bytes)?;
+                entity_builder.add(Position {
+                    node,
+                    local: na::Matrix4::from_column_slice(&column_slice),
+                });
+            }
+            ComponentType::Name => {
+                let name = String::from_utf8(component_bytes)?;
+                // Ensure that every node occupied by a character is generated.
+                if let Some(character) = read.get_character(&name)? {
```
Ralith, line 173: style: reduce rightward drift with let-else:
```rs
let Some(character) = read.get_character(&name)? else { return Ok(()); };
```

server/src/sim.rs
```diff
@@ -109,6 +112,104 @@ impl Sim {
...
+                    let mut current_node = NodeId::ROOT;
+                    for side in character
+                        .path
+                        .into_iter()
+                        .map(|side| Side::from_index(side as usize))
+                    {
+                        current_node = self.graph.ensure_neighbor(current_node, side);
+                    }
+                    if current_node != node {
```
Ralith, line 182: Should we log an error here? Presumably such an entity can never be loaded.

server/src/sim.rs
```diff
@@ -109,6 +112,104 @@ impl Sim {
...
+    ) -> anyhow::Result<()> {
+        match component_type {
+            ComponentType::Position => {
+                let column_slice: [f32; 16] = postcard::from_bytes(&component_bytes)?;
+                entity_builder.add(Position {
+                    node,
+                    local: na::Matrix4::from_column_slice(&column_slice),
+                });
+            }
+            ComponentType::Name => {
```
Ralith, line 170: Should we rename this to `CharacterId` for clarity?

server/src/sim.rs
```diff
@@ -109,6 +112,104 @@ impl Sim {
...
+                // Prepare all relevant components that are needed to support ComponentType::Name
+                entity_builder.add(InactiveCharacter(Character {
+                    name,
+                    state: CharacterState {
+                        velocity: na::Vector3::zeros(),
+                        on_ground: false,
+                        orientation: na::UnitQuaternion::identity(),
+                    },
+                }));
+                entity_builder.add(CharacterInput {
```
Ralith, line 201: Why do inactive characters need input state? Should this be inserted during activation instead?

server/src/sim.rs
```diff
@@ -149,7 +250,11 @@ impl Sim {
                     postcard::to_stdvec(pos.local.as_ref()).unwrap(),
                 ));
             }
-            if let Some(ch) = entity.get::<&Character>() {
+            if let Some(ch) = entity.get::<&Character>().or_else(|| {
+                entity
+                    .get::<&InactiveCharacter>()
+                    .map(|ich| hecs::Ref::map(ich, |ich| &ich.0)) // Extract Ref<Character> from Ref<InactiveCharacter>
+            }) {
```
Ralith, line 257: When you try to get a semantically equivalent field from two different possible component types, that's a strong signal that you should normalize that field into its own freestanding component (perhaps `struct Characterid(String)`) instead. Perhaps in a follow-up commit?

server/src/sim.rs
```diff
@@ -487,7 +641,11 @@ impl Sim {
     }
 }
 
+/// Collect all information about a particular entity for transmission to clients.
 fn dump_entity(world: &hecs::World, entity: Entity) -> Vec<Component> {
+    if world.satisfies::<&InactiveCharacter>(entity).unwrap() {
+        panic!("Inactive characters should not be sent to clients")
+    }
```
Ralith, line 648: style nit: it would be more conventional to write this as an `assert`.

Ralith (changes requested):
I think this already looks much cleaner, thanks!

common/src/proto.rs
```diff
@@ -110,6 +110,9 @@ pub struct Character {
     pub state: CharacterState,
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct InactiveCharacter(pub Character);
```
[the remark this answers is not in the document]
patowen: That might be a good idea. Currently, only `server`'s `sim.rs` uses this struct, so making it a private struct in `sim` might be the right move. Just moving it to the server crate won't guarantee that the client code won't see it, since `client` also has to depend on `server` because of single-player.

server/src/sim.rs
patowen: That might be a good idea, as we might want a different `name` field in the future. I'm not convinced we want to keep the word "character" long-term, as I assume we want a way to better distinguish between player characters and NPCs. For now, though, I'll rename to `CharacterId`.

server/src/sim.rs
patowen: I may have to work on this in a follow-up PR. There are too many controversial open questions related to this. There are several properties that could correspond to players:
- their unique ID
- a friendly name they might give themselves
- whether they are currently logged in or not
- their location in the world
- their `CharacterState` (consisting of physics-related info and client-provided orientation)
- the status of their inventory
- their current `CharacterInput`

Part of the reason I've been uncertain about the use of the `InactiveCharacter` struct is that it is conceptually unclear to me why the presence of `InactiveCharacter` vs `Character` is a good source of truth for whether a player is logged in or not. Why aren't any other structs, like `Inventory`, wrapped? I could imagine an "inventory stealer" entity that takes items from anything that has an inventory and a position, and, with the current implementation, there would be nothing stopping it from taking items from logged-out players.

It almost seems like we want to use `InactivePosition` instead of `InactiveCharacter`, since position is more fundamental, and it will pretty much stop all system logic from interacting with such entities. Also, using `InactivePosition` would seal the fate of having to get a semantically equivalent field from two different possible component types. Also, even `InactivePosition` doesn't cover everything because any system that just lists active players will need to check for the presence of a component in an unnatural way.

We could add an inactive counterpart to a whole bunch of components, but messing with more components than we need to seems like it will be more annoying long-term.

When more features are added in the future, the answers to these questions will hopefully become clearer, but for now, I feel like doing "something that works" is probably sufficient. I'm not convinced that this one example is enough to show that `Character` should be split.

patowen:
My previous update was merging the master branch and resolving merge conflicts. My next update will be handling review comments.

server/src/lib.rs
patowen: I went ahead and replaced the other opaque magic numbers used the same way as well.

server/src/sim.rs
patowen: That would be a good idea. I made the change.

common/src/proto.rs
```diff
@@ -114,3 +114,12 @@ pub struct Character {
...
     pub contents: Vec<EntityId>,
 }
+
+pub mod connection_error_codes {
+    use quinn::VarInt;
+
+    pub const CONNECTION_LOST: VarInt = VarInt::from_u32(0);
+    pub const BAD_CLIENT_COMMAND: VarInt = VarInt::from_u32(2);
+    pub const STREAM_ERROR: VarInt = VarInt::from_u32(1);
+    pub const NAME_CONFLICT: VarInt = VarInt::from_u32(3);
```
Ralith, line 124: nit: numerical order

Ralith (approved):

common/src/proto.rs
patowen: Oops, good catch. I got the names mixed up and fixed them but forgot to reorder them.

patowen:
Thanks for reviewing/merging!

Issue #264, opened 2023-02-26
Use newtype idiom to generate Hyperbolic Space-specific matrix and vector types


patowen:
To help avoid mistakes and make code easier to read and write, we should consider creating types to enforce the various invariants Hyperbolic vectors can have, much like nalgebra's `Point` vs. `Vector` vs. `UnitVector` types.

For example, we could have the following types (names subject to change):
- `HPoint`: A vector that can represent a point in hyperbolic space. The invariant is `x^2 + y^2 + z^2 < w^2` and `w > 0`.
- `HVector`: A vector that can represent a direction in hyperbolic space. The invariant is `x^2 + y^2 + z^2 > w^2`.
- `NormalHPoint`: A Lorentz-normalized `HPoint`. The invariant is `x^2 + y^2 + z^2 - w^2 = -1` and `w > 0`.
- `NormalHVector`: A Lorentz-normalized `HVector`. The invariant is `x^2 + y^2 + z^2 - w^2 = 1`.
- `HIsometry`: A matrix representing an isometry in Hyperbolic space. All columns are Lorentz-orthogonal to each other. The first three columns are `NormalHVector`s, and the last column is a `NormalHPoint`.

Under normal use, it should be pretty easy to enforce these invariants. There can be convenient helper and impl functions for common operations and conversions. For exceptional cases, functions with `unchecked` in their name can be added to opt out of this extra safety.

Ralith:
Let's continue the nalgebra convention and use `HyperPoint` and `HyperVector` rather than overloading "vector" further.

patowen:
I like that idea. I'll edit the description from `HyperWVector` to `HyperPoint` and from `HyperXyzVector` to `HyperVector`. If we for some reason need a single type that encompasses both (which I don't expect will happen), which originally would have been called `HyperVector`, we can just use a term like `GeneralHyperVector` instead.

patowen:
Another option is to call it `HPoint`, `HVector`, `UnitHPoint`, `UnitHVector`, and `HIsometry` to keep things shorter. There's already an `HPoint` in `math.rs`, but it isn't used anywhere, so it would make sense to replace it.

Ralith:
I like that concision. Given the broad scope of these types I think that's justified.

Another thought: for the "unit" types, would it make sense to represent w implicitly, since it's uniquely determined by the other components? Also, to avoid confusion with vectors of unit length, would it make sense to name them differently (perhaps LorentzHPoint)?

[...]

PR #420, opened 2024-07-23
types for hyperbolic space

strawberrycinnabar:
by #264 adds `MIsometry` and `MVector` types


client/src/graphics/draw.rs
```diff
@@ -289,7 +289,7 @@ impl Draw {
         let draw_started = Instant::now();
         let view = sim.as_ref().map_or_else(Position::origin, |sim| sim.view());
         let projection = frustum.projection(1.0e-4);
-        let view_projection = projection.matrix() * math::mtranspose(&view.local);
+        let view_projection = projection.matrix() * <math::MIsometry<_> as Into<na::Matrix4<_>>>::into(view.local.mtranspose());
```
Ralith, line 292: This should be `na::Matrix4::<_>::from(view.local.mtranspose())`, and similarly in the other cases below. Using UFCS to invoke `Into::into` is usually far more verbose than necessary.

Ralith (changes requested):

common/src/dodeca.rs
```diff
@@ -45,33 +46,32 @@ impl Side {
...
     }
 
     /// Reflection across this side
     #[inline]
-    pub fn reflection(self) -> &'static na::Matrix4<f32> {
+    pub fn reflection(self) -> &'static MIsometry<f32> {
         &reflections_f32()[self as usize]
     }
 
-    /// Reflection across this side
```
patowen, line 64: This comment looks like it was accidentally removed.

common/src/dodeca.rs
```diff
@@ -602,17 +600,17 @@ mod tests {
...
     }
 
     #[test]
     fn radius() {
-        let corner = Vertex::A.chunk_to_node_f64() * math::origin();
+        let corner = *Vertex::A.dual_to_node() * MVector::origin();
         assert_abs_diff_eq!(
-            BOUNDING_SPHERE_RADIUS_F64,
-            math::distance(&corner, &math::origin()),
+            BOUNDING_SPHERE_RADIUS,
```
patowen, line 612: If we switch from `f64` to `f32` precision, it may be good to pick a more generous epsilon for the test, like `1e-5` or so. Otherwise, it's possible the test might fail if an otherwise acceptable change is made.

That being said, it might be better to just revert this precision loss by using `dual_to_node_f64` here. I don't believe there's any logic here that only works with `f32` precision.

common/src/math.rs
```diff
@@ -3,97 +3,463 @@
...
+    pub fn renormalize_isometry(self) -> MIsometry<N>
+    {
+        let boost = translate(&MVector::origin(), &MVector(self.0.index((.., 3)).into()).lorentz_normalize());
+        let inverse_boost = boost.mtranspose();
+        let rotation = renormalize_rotation_reflection(
+            &((inverse_boost * self).0).fixed_view::<3, 3>(0, 0).clone_owned(),
+        );
+        MIsometry(boost.0 * rotation.to_homogeneous())
+    }
+    /*pub fn cast<N2: Scalar>(self) -> MIsometry<N2> where N: simba::scalar::SubsetOf<N2>
```
patowen, line 205: Don't forget to delete commented-out code before merging.

common/src/math.rs
```diff
@@ -3,97 +3,463 @@
...
+    /// Minkowski inner product, aka <a, b>_h
+    pub fn mip(self, other: &Self) -> N
+    {
+        self.0.x * other.0.x + self.0.y * other.0.y + self.0.z * other.0.z - self.0.w * other.0.w
+    }
+    pub fn minkowski_outer_product(self, other: &Self) -> na::Matrix4<N>
+    {
+        ((self).0) * na::RowVector4::new(other.0.x, other.0.y, other.0.z, -other.0.w)
+    }
+    pub fn from_gans(gans: &na::Vector3<N>) -> Self
```
patowen, line 234: Is this function called from anywhere? I expect that the answer would be no, in which case it should be safe to remove this function.

common/src/math.rs
```diff
@@ -3,97 +3,463 @@
...
+        MVector(na::Vector4::new(gans.x, gans.y, gans.z, w))
+    }
+    #[inline]
+    pub fn zero() -> Self
+    {
+        Self(na::zero())
+    }
+    pub fn origin() -> Self
+    {
+        Self(na::Vector4::new(na::zero(),na::zero(),na::zero(),na::one()))
```
patowen, line 248: I would recommend simplifying this to
```rust
pub fn origin() -> Self {
    Self::w()
}
```

Personally, I'm not sure if we should have the `origin()` function at all, since it's equivalent to `w()`, but that was already true before this PR, so in the spirit of keeping changes incremental, I would recommend keeping `origin()` around for now, but we might as well make its equivalence to `w()` clearer.

common/src/math.rs
```diff
@@ -3,97 +3,463 @@
...
+        MIsometry(self.0 * rhs)
+    }
+}
+
+impl<N: RealField + Copy> std::ops::AddAssign for MVector<N>
+{
+    #[inline]
+    fn add_assign(&mut self, other: Self)
+    {
+        *self = *self + other;
```
patowen, line 372: I wonder if `self.0 += other.0` would be better here.

strawberrycinnabar: back when i tried that, looks like it just calls add_assign and just recurses, so i guess for brevity i could add a macro or something, though im not familiar with macros

common/src/math.rs
strawberrycinnabar: id be fine with changing the definition to ``w()``, but having ``w()`` be solely used feels a bit like obscuring the meaning of it

client/src/graphics/draw.rs
```diff
@@ -289,7 +289,7 @@ impl Draw {
         let draw_started = Instant::now();
         let view = sim.as_ref().map_or_else(Position::origin, |sim| sim.view());
         let projection = frustum.projection(1.0e-4);
-        let view_projection = projection.matrix() * <math::MIsometry<_> as Into<na::Matrix4<_>>>::into(view.local.mtranspose());
+        let view_projection = projection.matrix() * na::Matrix4::<_>::from(view.local.mtranspose());
```
Ralith, line 292: Is the `::<_>` here actually needed?

common/src/math.rs
Ralith: It won't recurse, because `self.0` is a different type than `self`.

common/src/math.rs
Ralith: The suggestion is to change the definition, not remove it. I think we all agree that `origin()` is hugely valuable for readability.

common/src/math.rs
strawberrycinnabar: oh misread, that makes sense

common/src/dodeca.rs
strawberrycinnabar: i just used ``f64`` there

common/src/math.rs
```diff
@@ -3,97 +3,390 @@
...
+    /// Point or plane reflection around point or normal `p`
+    pub fn reflect(self) -> MIsometry<N> {
+        MIsometry(
+            na::Matrix4::<N>::identity()
+                - self.minkowski_outer_product(&self) * na::convert::<_, N>(2.0) / self.mip(&self),
+        )
+    }
+    /// Minkowski inner product, aka <a, b>_h
+    pub fn mip(self, other: &Self) -> N {
+        self.0.x * other.0.x + self.0.y * other.0.y + self.0.z * other.0.z - self.0.w * other.0.w
```
patowen, line 195: Now that you've implemented `Deref`, it should be possible to simplify this to
```rust
self.0.x * other.x + self.y * other.y + self.z * other.z - self.w * other.w
```

common/src/math.rs
```diff
@@ -3,97 +3,390 @@
...
+    }
+    pub fn minkowski_outer_product(self, other: &Self) -> na::Matrix4<N> {
+        ((self).0) * na::RowVector4::new(other.0.x, other.0.y, other.0.z, -other.0.w)
```
patowen, line 198: Now that you've implemented Deref, it should be possible to simplify this to
```rust
self.0 * na::RowVector4::new(other.x, other.y, other.z, -other.w)
```

common/src/math.rs
```diff
@@ -3,97 +3,390 @@
...
-        let w = (sqr(self.0.x) + sqr(self.0.y) + sqr(self.0.z) + na::one()).sqrt();
-        na::Vector4::new(self.0.x, self.0.y, self.0.z, w)
+        let w = (sqr(gans.x) + sqr(gans.y) + sqr(gans.z) + na::one()).sqrt();
+        MVector(na::Vector4::new(gans.x, gans.y, gans.z, w))
+    }
+    #[inline]
+    pub fn zero() -> Self {
+        Self(na::zero())
+    }
+    pub fn origin() -> Self {
```
patowen, line 210: Let's `#[inline]` this function as well.

client/src/graphics/draw.rs
```diff
@@ -482,8 +482,8 @@ impl Draw {
                         .expect("positionless entity in graph");
                     if let Some(character_model) = self.loader.get(self.character_model) {
                         if let Ok(ch) = sim.world.get::<&Character>(entity) {
-                            let transform = transform
-                                * pos.local
+                            let transform = na::Matrix4::from(transform)
+                                * na::Matrix4::from(pos.local)
                                 * na::Matrix4::new_scaling(sim.cfg().meters_to_absolute)
                                 * ch.state.orientation.to_homogeneous();
```
patowen, line 488: You can simplify this slightly to
```rust
let transform = na::Matrix4::from(transform * pos.local)
    * na::Matrix4::new_scaling(sim.cfg().meters_to_absolute)
    * ch.state.orientation.to_homogeneous();
```

common/src/character_controller/collision.rs
```diff
@@ -58,7 +67,7 @@ pub fn check_collision(
             // This normal now represents a contact point at the origin, so we omit the w-coordinate
             // to ensure that it's orthogonal to the origin.
             normal: na::UnitVector3::new_normalize(
-                (math::mtranspose(&displacement_transform) * hit.normal).xyz(),
+                ((displacement_transform.mtranspose()) * hit.normal).xyz(),
```
patowen, line 70:
```suggestion
                (displacement_transform.mtranspose() * hit.normal).xyz(),
```

common/src/collision_math.rs
```diff
@@ -43,14 +43,14 @@ impl Ray {
...
         sinh_radius: f32,
     ) -> Option<f32> {
-        let mip_pos_a = math::mip(&self.position, line_normal0);
-        let mip_dir_a = math::mip(&self.direction, line_normal0);
-        let mip_pos_b = math::mip(&self.position, line_normal1);
-        let mip_dir_b = math::mip(&self.direction, line_normal1);
+        let mip_pos_a = &self.position.mip(line_normal0);
+        let mip_dir_a = &self.direction.mip(line_normal0);
+        let mip_pos_b = &self.position.mip(line_normal1);
+        let mip_dir_b = &self.direction.mip(line_normal1);
```
patowen, line 53:
```suggestion
        let mip_pos_a = self.position.mip(line_normal0);
        let mip_dir_a = self.direction.mip(line_normal0);
        let mip_pos_b = self.position.mip(line_normal1);
        let mip_dir_b = self.direction.mip(line_normal1);
```

common/src/collision_math.rs
```diff
@@ -63,17 +63,17 @@ impl Ray {
...
-        let mip_pos_b = math::mip(&self.position, point_normal1);
-        let mip_dir_b = math::mip(&self.direction, point_normal1);
-        let mip_pos_c = math::mip(&self.position, point_normal2);
-        let mip_dir_c = math::mip(&self.direction, point_normal2);
+        let mip_pos_a = &self.position.mip(point_normal0);
+        let mip_dir_a = &self.direction.mip(point_normal0);
+        let mip_pos_b = &self.position.mip(point_normal1);
+        let mip_dir_b = &self.direction.mip(point_normal1);
+        let mip_pos_c = &self.position.mip(point_normal2);
+        let mip_dir_c = &self.direction.mip(point_normal2);
```
patowen, line 76:
```suggestion
        let mip_pos_a = self.position.mip(point_normal0);
        let mip_dir_a = self.direction.mip(point_normal0);
        let mip_pos_b = self.position.mip(point_normal1);
        let mip_dir_b = self.direction.mip(point_normal1);
        let mip_pos_c = self.position.mip(point_normal2);
        let mip_dir_c = self.direction.mip(point_normal2);
```

common/src/collision_math.rs
```diff
@@ -84,9 +84,9 @@ impl Ray {
 
     /// Finds the tanh of the distance a point will have to travel along a ray before it
     /// intersects the given plane.
-    pub fn solve_point_plane_intersection(&self, plane_normal: &na::Vector4<f32>) -> Option<f32> {
-        let mip_pos_a = math::mip(&self.position, plane_normal);
-        let mip_dir_a = math::mip(&self.direction, plane_normal);
+    pub fn solve_point_plane_intersection(&self, plane_normal: &MVector<f32>) -> Option<f32> {
+        let mip_pos_a = &self.position.mip(plane_normal);
+        let mip_dir_a = &self.direction.mip(plane_normal);
```
patowen, line 89:
```suggestion
        let mip_pos_a = self.position.mip(plane_normal);
        let mip_dir_a = self.direction.mip(plane_normal);
```

common/src/plane.rs
```diff
@@ -46,24 +46,24 @@ impl Mul<Plane<f64>> for Side {
...
-    pub fn normal(&self) -> &na::Vector4<N> {
+    pub fn normal(&self) -> &MVector<N> {
         &self.normal
     }
 
     /// Shortest distance between the plane and a point
-    pub fn distance_to(&self, point: &na::Vector4<N>) -> N {
-        let mip_value = mip(&self.normal, point);
+    pub fn distance_to(&self, point: &MVector<N>) -> N {
+        let mip_value = &self.normal.mip(point);
```
patowen, line 66:
```suggestion
        let mip_value = self.normal.mip(point);
```

common/src/plane.rs
```diff
@@ -72,15 +72,16 @@ impl<N: na::RealField + Copy> Plane<N> {
 impl Plane<f64> {
     /// Like `distance_to`, but using chunk coordinates for a chunk in the same node space
     pub fn distance_to_chunk(&self, chunk: Vertex, coord: &na::Vector3<f64>) -> f64 {
-        let pos = lorentz_normalize(&(chunk.chunk_to_node_f64() * coord.push(1.0)));
+        let pos =
+            (MVector::<f64>::from(chunk.chunk_to_node_f64() * coord.push(1.0))).lorentz_normalize();
```
patowen, line 76:
```suggestion
        let pos = (MVector::from(chunk.chunk_to_node_f64() * coord.push(1.0))).lorentz_normalize();
```
The `::<f64>` is unneeded here.

common/src/traversal.rs
```diff
@@ -30,9 +30,9 @@ pub fn ensure_nearby(graph: &mut Graph, start: &Position, distance: f32) {
                 continue;
             }
             visited.insert(neighbor);
-            let neighbor_transform = current_transform * side.reflection();
-            let neighbor_p = neighbor_transform * math::origin();
-            if -math::mip(&start_p, &neighbor_p) > distance.cosh() {
+            let neighbor_transform = current_transform * *side.reflection();
+            let neighbor_p = neighbor_transform * MVector::origin();
+            if -(start_p.mip(&neighbor_p)) > distance.cosh() {
```
patowen, line 35:
```suggestion
            if -start_p.mip(&neighbor_p) > distance.cosh() {
```

common/src/traversal.rs
```diff
@@ -59,20 +59,20 @@ pub fn nearby_nodes(
...
-        transform: na::Matrix4::identity(),
+        transform: MIsometry::identity(),
     });
     visited.insert(start.node);
 
     while let Some(current) = pending.pop_front() {
-        let current_p = current.transform * math::origin();
-        if -math::mip(&start_p, &current_p) > distance.cosh() {
+        let current_p = current.transform * MVector::origin();
+        if -(start_p.mip(&current_p)) > distance.cosh() {
```
patowen, line 72:
```suggestion
        if -start_p.mip(&current_p) > distance.cosh() {
```

patowen (changes requested):
I've reviewed the code more carefully this time. I believe that once my suggestions are dealt with, this PR will be ready to merge.

patowen (changes requested):
Looks good to me! Once you run `cargo fmt` on this, it should be ready to merge.

patowen (approved):

patowen:
Thanks!

patowen:
The "package" job for Ubuntu failed for this PR, but that is entirely unrelated to this PR. I believe the root cause is https://github.com/actions/checkout/issues/1809. I hope that a good solution is found for this before it's time to release another version of Hypermine, but otherwise, I'm not sure what the best solution will be that requires the least boring maintenance work. Ideally, these GitHub workflows shouldn't need to be edited very often at all, but this is the second time this year that the package workflows randomly broke.

PR #422, opened 2024-07-28
Update winit

Ralith:
Builds, but still needs manual testing.

### DIFF
--- a/client/src/graphics/window.rs
+++ b/client/src/graphics/window.rs
@@ -7,13 +7,12 @@ use lahar::DedicatedImage;
 use raw_window_handle::{HasDisplayHandle, HasWindowHandle};
 use tracing::{error, info};
 use winit::event::KeyEvent;
-use winit::event_loop::EventLoopWindowTarget;
+use winit::event_loop::ActiveEventLoop;
 use winit::keyboard::{KeyCode, PhysicalKey};
 use winit::{
     dpi::PhysicalSize,
     event::{DeviceEvent, ElementState, MouseButton, WindowEvent},
-    event_loop::EventLoop,
-    window::{CursorGrabMode, Window as WinitWindow, WindowBuilder},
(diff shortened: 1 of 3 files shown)


patowen:
I was able to run this branch successfully on my Windows machine, and all the buttons still seem to work as expected.

client/src/graphics/window.rs
```diff
@@ -111,194 +104,167 @@ impl Window {
...
-                                0.0,
-                                0.0,
-                                2.0 * (anticlockwise as u8 as f32 - clockwise as u8 as f32)
-                                    * dt.as_secs_f32(),
-                            );
-
-                            sim.step(dt, &mut self.net);
-                            last_frame = this_frame;
-                        }
+    /// Run the event loop until process exit
```
patowen, line 134: This comment is now outdated.

client/src/graphics/window.rs
```diff
@@ -707,3 +673,32 @@ struct Frame {
...
+    forward: bool,
+    back: bool,
+    left: bool,
+    right: bool,
+    up: bool,
+    down: bool,
+    jump: bool,
+    clockwise: bool,
+    anticlockwise: bool,
+    last_frame: Option<Instant>,
```
patowen, line 688: Out of all fields, this one seems the most out of place in the `InputState` struct, but at the same time, it does provide context for the rest of the fields, and in the context of a simulation, "amount of time passed" is indeed one of the inputs, so I don't think moving this field somewhere else would necessarily be any better.

patowen (approved):
Looks good to me! I left a few comments for minor things. Thanks for working on this, as it would have been very tricky for me to do.

patowen (approved):

client/src/graphics/window.rs
Ralith: Eh, lifted it out.

Issue #405, opened 2024-06-22
Add single-player optimizations


patowen:
When playing Hypermine in single-player, there are a few things that it does that do not make sense from an outsider's perspective:
- In Windows, the first time you play it, a Windows Firewall message pops up. This could be prevented by simulating network features by doing everything in memory instead of sending TCP/UDP packets through localhost.
- Both the server and client generate chunks independently, which, in single player, means that many chunks are generated twice unnecessarily.
- The server and client's steps are timed independently, meaning that one can lag without the other, causing difficult-to-justify jerky movement.

Ideally, it should be possible to resolve the above issues without significantly increasing maintenance burden.

I don't think it makes sense to prioritize this issue until enough gameplay features are fleshed out that the codebase has stabilized somewhat, and render-distance related optimizations have been improved. This is especially true because doing this early risks single-player and multi-player code diverging, making testing netcode difficult (and easy to forget).

PR #425, opened 2024-07-29
In-process server

Ralith:
Towards https://github.com/Ralith/hypermine/issues/405.

Rough in some places ~and not actually working correctly yet~.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -8,6 +8,7 @@ use quinn::rustls::pki_types::{CertificateDer, PrivatePkcs8KeyDer};
 use save::Save;
 
 use ash::khr;
+use server::Server;
 use tracing::{error, error_span, info};
 use winit::{
     application::ApplicationHandler,
@@ -44,21 +45,31 @@ fn main() {
             }
         };
 
-        std::thread::spawn(move || {
-            let span = error_span!("server");
-            let _guard = span.enter();
-            if let Err(e) = server::run(
-                server::NetParams {
-                    certificate_chain: vec![CertificateDer::from(cert)],
-                    private_key: PrivatePkcs8KeyDer::from(key).into(),
-                    socket,
-                },
-                sim_cfg,
-                save,
-            ) {
+        let server = match Server::new(
+            server::NetParams {
+                certificate_chain: vec![CertificateDer::from(cert)],
+                private_key: PrivatePkcs8KeyDer::from(key).into(),
+                socket,
+            },
+            sim_cfg,
+            save,
+        ) {
+            Ok(server) => server,
+            Err(e) => {
                 eprintln!("{e:#}");
                 std::process::exit(1);
             }
+        };
+
+        std::thread::spawn(move || {
+            #[tokio::main(flavor = "current_thread")]
+            async fn run_server(server: Server) {
+                server.run().await;
+            }
+
+            let span = error_span!("server");
+            let _guard = span.enter();
+            run_server(server);
         });
     }
     let mut app = App {
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -1,15 +1,12 @@
-use std::{
-    net::{SocketAddr, UdpSocket},
-    sync::Arc,
-};
+use std::{sync::Arc, thread};
 
 use client::{graphics, metrics, net, Config};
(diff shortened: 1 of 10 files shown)


patowen:
I wanted to provide this comment mainly out of curiosity:

Weirdly, when I tried to compile this PR, I got
```
error[E0282]: type annotations needed for `Box<_>`
  --> C:\Users\Patrick\.cargo\registry\src\index.crates.io-6f17d22bba15001f\time-0.3.28\src\format_description\parse\mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++
```

I resolved this by deleting Cargo.lock, so if anyone is interested in investigating what semver issue lead to this error, I've attached my Cargo.lock file.

[Cargo.lock.zip](https://github.com/user-attachments/files/16555368/Cargo.lock.zip)

It's probably not worthwhile debugging, though. It seems to be entirely unrelated to this PR and likely instead related to changes to my setup from hopping around various PRs.

Ralith:
Probably a minor issue in an upstream crate that got fixed.

client/src/main.rs
```diff
@@ -20,52 +18,56 @@ fn main() {
...
+            std::fs::create_dir_all(save.parent().unwrap()).unwrap();
+            let save = match Save::open(&save, config.local_simulation.chunk_size) {
+                Ok(x) => x,
+                Err(e) => {
+                    error!("couldn't open save: {}", e);
+                    return;
+                }
+            };
+
+            let runtime = tokio::runtime::Builder::new_current_thread()
```
patowen, line 39: I think part of why this doesn't work yet is that `runtime` is dropped and the end of this block, which is when `net` gets set to `handle`.

I believe the simplest fix is to spawn a thread for the runtime to live in. If the server code is to share a thread with the rest of the app, we'd need to see how hard it would be to share the thread with the blocking winit `run_app` function

patowen:
This improved efficiency seems to have revealed a subtle bug I introduced in <https://github.com/Ralith/hypermine/pull/414/files#diff-e8d8336a10bad73657026c14d2e141ca92d15c95a92aa7ca93df5366de21754eR186>.

Normally, the server's graph is kept in a clean state. Whenever nodes are added to it, these nodes are processed fully before any other code gets a chance to run. See <https://github.com/Ralith/hypermine/blob/95219142bd400c93fb7b9db1e3c59d52ad2dc486/server/src/sim.rs#L444-L462>. This generally creates a guarantee that if `snapshot` is called (for transmitting info to new clients), and `step` is called afterwards, the `Spawns` returned by `step` will never contain any graph nodes that were already returned by `snapshot`.

Before a server's first step, its state is polluted. Calls to `ensure_neighbor` are invoked without an associated `populate_fresh_nodes`. This means that when `snapshot` is called, these nodes are included as if they're already initialized. However, `step` still views these nodes as new/uninitialized, so it also includes them in the `Spawns` it returns.

The effect is that if a client connects to a server before the server runs its first step, then as long as there are any players outside of the root node that need to be loaded, the server will send a client the same node twice, resulting in a crash.

The simplest workaround is to add `result.step();` right before `result` is returned in `Sim::new`, guaranteeing that the first step will complete before anything else.

However, a longer-term fix would be to ensure that `server::Sim` never reaches this invalid state to begin with. I'm open to ideas on how to do this. Long-term, I wonder if we want to stop `Graph` from storing its own fresh nodes, since that feels like a potential foot-gun.

client/src/main.rs
Ralith: Yep, that was it. Thakns!

Ralith:
> Long-term, I wonder if we want to stop Graph from storing its own fresh nodes, since that feels like a potential foot-gun.

Our plan has always been to do away with the "fresh nodes" paradigm entirely in favor of deterministic node IDs, right?

patowen:
> Our plan has always been to do away with the "fresh nodes" paradigm entirely in favor of deterministic node IDs, right?

I believe so, although it's unclear to me what that will look like. It probably makes sense to figure that out before deciding where to store them.

PR #427, opened 2024-08-11
Ensure the server state is consistent whenever snapshot is called

patowen:
Before this commit, if `snapshot` was called before the server ran its first step, there would be some unpopulated nodes in the graph, resulting in such nodes being transmitted to the client twice, once during the call to `snapshot`, and once as an accumulated change when calling `step`.

This PR ensures that such an inconsistent state no longer occurs.

### DIFF
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -73,6 +73,11 @@ impl Sim {
         result
             .load_all_entities(save)
             .expect("save file must be of a valid format");
+        // Loading entities can cause graph nodes to also be created, so we should populate them before returning.
+        result.populate_fresh_graph_nodes();
+        // As no players have logged in yet, and `snapshot` may be called before the first call of `step`,
+        // make sure that `accumulated_changes` is empty to avoid accidental double-spawns of anything.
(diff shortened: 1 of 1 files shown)


server/src/sim.rs
```diff
@@ -185,6 +190,7 @@ impl Sim {
                 {
                     current_node = self.graph.ensure_neighbor(current_node, side);
                 }
+                self.populate_fresh_graph_nodes();
```
Ralith, line 193: Would it be tidier to do this once at the end of `load_all_entities` rather than separately for every character?

Ralith (approved):

server/src/sim.rs
patowen: My main concern is that `load_component` puts `server::Sim` in a weird state, and I expect it to be easier to reason about the code-long term if the reason for calling `populate_fresh_graph_nodes` is next to where it's called.

However, I goofed up, as `populate_fresh_graph_nodes` is not designed to be called multiple times in a single `step`. I believe that is the part that I ought to fix.

Ralith (approved):

Issue #428, opened 2024-08-13
Finish up incomplete serialization work


patowen:
I'm filing a separate issue from #397 for this to better track the work I have left for later while working on inventory code.

- [ ] Use the term `CharacterId` instead of `Name` when identifying player-controlled characters (as suggested in https://github.com/Ralith/hypermine/pull/414#discussion_r1693999349)
- [ ] Move inactive players to a separate redb table so that they can remain unloaded
- [ ] Consider reorganizing `Character` and `InactiveCharacter` components to avoid needing duplicate logic to extract `CharacterId` from both (as suggested in https://github.com/Ralith/hypermine/pull/414#discussion_r1693999571)
- [ ] Serialize player inventories when saving and loading (inventories are currently skipped in the save/load logic)

I plan to try to keep the below sections up-to-date to serve as a living design document that I or someone else will be able to reference while implementing the above.

# Decision status
## Disconnection vs deactivation
The concept of "player disconnecting" and "player character becoming inactive", while generally simultaneous, should not be baked in as fundamentally the same in the engine. Keeping them separate would allow for more options, such as a "battle royale" game mode, where disconnected players stand there doing nothing (like a sitting duck) to prevent cheating while also preventing disqualification from accidental disconnection.

Whether a player is active or inactive is included when serializing. Although all active players will be immediately logged out if a server is rebooted, we don't necessarily want to enforce that their characters should be deactivated as well.

## Serialization of inactive characters
Each inactive player should have its own `EntityNode`-like data structure but keyed by their player ID instead of `NodeId` (in a different `redb` table from the `EntityNode`s in the graph). The inactive players can keep track of where they _would_ be in if they were active (probably with a new component than the way position is tracked for active entities), but since they're inactive, they're not part of the `Graph` at all.

This way, they would only need to be stored in disk and never loaded into memory until they log on. An `EntityNode` that is part of the `Graph` will only contain active players, so with this idea, it would be safe to save changes to it without taking inactive players into account.

Deactivating entities not associated with players is not handled by the above logic if the key is `CharacterId`, but there is no clear use case for this, and we should be able to support it with some adjustments if it comes up.

# Open questions
## How to better represent characters in ECS
The current split between `Character` and `InactiveCharacter` is a little clunky. For instance, some existing code has to extract the name field from both for serialization.

I had expressed uncertainty about this in https://github.com/Ralith/hypermine/pull/414#discussion_r1694277120.

[...]

PR #431, opened 2024-10-05
Serialize the player's inventory in save files

patowen:
This completes part of #428.

Allows users to exit and re-open Hypermine while keeping the contents of their inventory, as previously, the inventory was cleared every time the game was loaded.

### DIFF
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -147,6 +147,7 @@ impl Sim {
         let entity_id = EntityId::from_bits(u64::from_le_bytes(save_entity.entity));
         let mut entity_builder = EntityBuilder::new();
         entity_builder.add(entity_id);
+        entity_builder.add(node);
         for (component_type, component_bytes) in save_entity.components {
             self.load_component(
                 read,
@@ -210,7 +211,22 @@ impl Sim {
                         orientation: na::UnitQuaternion::identity(),
                     },
                 }));
-                entity_builder.add(Inventory { contents: vec![] });
+            }
+            ComponentType::Material => {
+                let material: u16 =
+                    u16::from_le_bytes(component_bytes.try_into().map_err(|_| {
+                        anyhow::anyhow!("Expected Material component in save file to be 2 bytes")
+                    })?);
+                entity_builder.add(Material::try_from(material)?);
+            }
+            ComponentType::Inventory => {
+                let mut contents = vec![];
+                for chunk in component_bytes.chunks(8) {
+                    contents.push(EntityId::from_bits(u64::from_le_bytes(
+                        chunk.try_into().unwrap(),
+                    )));
+                }
+                entity_builder.add(Inventory { contents });
             }
         }
         Ok(())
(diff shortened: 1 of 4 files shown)


server/src/sim.rs
```diff
@@ -242,42 +254,69 @@ impl Sim {
...
-            postcard_helpers::serialize(
-                &SaveEntity {
-                    entity: id.to_bits().to_le_bytes(),
-                    components,
-                },
-                &mut repr,
-            )
-            .unwrap();
-            entities.push(repr);
+            let reprs = self.snapshot_entity_and_children(entity);
```
Ralith, line 257: Are child entities not also in `graph_entities`?

server/src/sim.rs
```diff
@@ -242,42 +254,69 @@ impl Sim {
...
+            entities.extend_from_slice(&reprs);
         }
 
         save::EntityNode { entities }
     }
 
+    fn snapshot_entity_and_children(&self, entity: Entity) -> Vec<Vec<u8>> {
```
Ralith, line 264: Factor out the function in one commit, then modify its behavior in a separate commit. When those two changes are in the same commit it's much harder to identify behavior changes.

Ralith (changes requested):

server/src/sim.rs
patowen: Not currently, although given how tricky this implementation is, I would consider this to be a mistake on my part.

There's a distinction between entities that have an inherent position (which is managed with a `Position` component) and nodes that don't, such as inventory items. For the latter, they belong to an entity that does have a position.

I had decided that `graph_entities` would only contain entities that have a position, but I have a good reason to second-guess that, since that doesn't correspond with how save files are organized at all. I'll go ahead and fix this before my next push to this branch.

server/src/sim.rs
patowen: Good point. Depending on how https://github.com/Ralith/hypermine/pull/431#discussion_r1797768956 is resolved, I may end up undoing the refactor.

server/src/sim.rs
```diff
@@ -210,7 +210,19 @@ impl Sim {
                         orientation: na::UnitQuaternion::identity(),
                     },
                 }));
-                entity_builder.add(Inventory { contents: vec![] });
+            }
+            ComponentType::Material => {
+                let material: u16 = u16::from_le_bytes(component_bytes.try_into().unwrap());
```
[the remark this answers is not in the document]
patowen: I'm now returning an error.

server/src/sim.rs
patowen: I've updated the PR so that child entities are now in `graph_entities`. To accomplish this (what I believe to be) the most understandable way possible, I've made `NodeId` into its own component intrinsically linked with `graph_entities` so that inventory items can be associated with a `NodeId` without needing a `Position`.

This is helpful because `Position` has a bunch of other functionality, such as syncing with clients, so separating this functionality is important.

server/src/sim.rs
patowen: I have now undone this refactor.

Ralith (approved):

PR #442, opened 2025-02-15
Document and simplify math module

patowen:
This PR makes the following changes to the `math` module:
- Functions have been re-ordered for simplicity (separated to a different commit for diff readability)
- `simba` has been added as a direct dependency so that `to_f32` and `to_f64` can be combined into a single `cast` function (separated to a different commit because of the extra dependency)
- Rustdoc comments are added to `MVector` and `MIsometry` structs and functions that could benefit from them.
- Code comments are added inside functions with complicated implementations to better explain them.
- A few functions like `as_ref` and `row` are modified to better match their `nalgebra` counterparts (with fewer copies)
- Functions that create an `MIsometry` are moved into `MIsometry` (instead of being free functions, or, in the case of `reflect()`, functions in `MVector`. These functions are also renamed to nouns to better represent the fact that they produce a matrix that performs the operation rather than performing the operation themselves.
- Most functions with "Minkowski" or "Lorentz" in their name have that part removed for brevity, as it is redundant with the `MVector` or `MIsometry` type they are associated with. This includes `mtranspose`, which is renamed to `inverse`, as that is the purpose of that operation.
    - `mip` is excluded from this rename due to familiarity and how often it's used (and the lack of a good name that still has brevity)
    - `minkowsky_outer_product` is also excluded from this rename to parallel `mip`.
- The algorthm to renormalize an isometry has been simplified to use nalgebra's QR factorization for the orientation component.
- The `midpoint`, `distance`, `reflection`, and `translation` functions are modified to only accept normalized `MVector`s as input. This is documented in the method header (and I hope that we can enforce this with types in the future).
    - In doing so, the `reflection` function also now only accepts direction-like `MVectors`, as the formula has a sign error if used with point-like `MVectors`.
    - This change also allows the `distance` function to be more numerically stable.
-  A `debug_assert!` is added to `MVector::normalized` to detect possible logic errors in the future.
- The `minkowski_outer_product` function is made private, as it is only used for intermediate calculations.
- A numerical stability warning was added to `MIsometry::renormalized` and `MVector::normalized`
- The `From<na::Vector3<N>>` implementation was removed for `MVector` because it is only used once (in `plane.rs`)
- Fix: `MIsometry` can no longer be multiplied by a number, since this violates the isometry constraint.

All changes outside the `math` module are just for compatibility with the `math` module's changes.

### DIFF
--- a/common/src/math.rs
+++ b/common/src/math.rs
@@ -8,6 +8,7 @@
 
 use na::{RealField, Scalar};
 use serde::{Deserialize, Serialize};
+use simba::scalar::SupersetOf;
 use std::ops::*;
 
 /// A stack-allocated 4-dimensional column-vector in Minkowski space. Such
@@ -106,6 +107,12 @@ impl<N: RealField + Copy> MVector<N> {
         self.0 * na::RowVector4::new(other.x, other.y, other.z, -other.w)
     }
 
+    /// Cast the components of `self` to another type.
+    #[inline]
+    pub fn cast<N2: RealField + Copy + SupersetOf<N>>(self) -> MVector<N2> {
+        MVector(self.0.cast())
+    }
+
     /// The column vector with components `[0, 0, 0, 0]`.
     #[inline]
     pub fn zero() -> Self {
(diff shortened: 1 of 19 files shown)


common/src/math.rs
```diff
@@ -53,79 +89,193 @@ impl<N: Scalar> Deref for MIsometry<N> {
...
     #[inline]
     pub fn identity() -> Self {
         Self(na::Matrix4::identity())
     }
-    #[inline]
-    pub fn map<N2: Scalar, F: FnMut(N) -> N2>(&self, f: F) -> MIsometry<N2> {
-        MIsometry(self.0.map(f))
+
+    /// The reflection about the hyperbolic plane represented by the given
+    /// normalized vector. An incorrect matrix will be returned if `a` and `b`
```
patowen, line 173: Note to self: `a` and `b` here don't exist. I copied this part of the comment from `translation` and forgot to update the comment to be accurate.

Ralith (approved):

PR #443, opened 2025-02-15
Document and refactor dodeca.rs

### LINKED ISSUE #381: Modernize Rust idioms

patowen:
Fixes #381

This PR makes the following changes to the `dodeca` module:
- `LazyLock` is used instead of `OnceLock` (This change is also made in cursor.rs)
- `Side` and `Vertex` have a `VALUES` array allowing for easier conversion between integers and enum values
- `VERTEX_COUNT` and `SIDE_COUNT` have been replaced with `Vertex::COUNT` and `Side::COUNT`
- `Vertex::from_sides` now takes in an array instead of 3 separate parameters for consistency with `Vertex::canonical_sides`
- Arrays are now initialized to their final values when it makes sense to do so (instead of being initialized to a dummy value and then filled in). This utilizes `array::from_fn` and `array::map` in various places.
- Complicated math is better explained in code comments
- Some implementations are replaced with simpler implementations, such as `is_facing`, `ADJACENT`, and `SIDES_TO_VERTEX`

Changes in other modules are for compatibility with the changes to `dodeca`.

### DIFF
--- a/common/src/dodeca.rs
+++ b/common/src/dodeca.rs
@@ -1,12 +1,22 @@
 //! Tools for processing the geometry of a right dodecahedron
 
-use data::*;
 use serde::{Deserialize, Serialize};
 
 use crate::math::{MIsometry, MVector};
 use crate::voxel_math::ChunkAxisPermutation;
 
 /// Sides of a right dodecahedron
+///
+/// These sides are arranged based on the following adjacency graph, although it
+/// is recommended not to hardcode side names in other code:
+/// ```nocode
+///          A
+/// (D-) I-E-B-C-D (-I)
+///  (K-) L-G-F-H-K (-L)
+///           J
+/// ```
+/// The above adjacency graph can be read as a map of a globe, where side A is
+/// at the north pole, and side J is at the south pole.
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
 pub enum Side {
     A,
@@ -24,16 +34,23 @@ pub enum Side {
 }
 
 impl Side {
-    #[inline]
-    pub fn from_index(x: usize) -> Self {
-        use Side::*;
-        const VALUES: [Side; SIDE_COUNT] = [A, B, C, D, E, F, G, H, I, J, K, L];
-        VALUES[x]
-    }
+    pub const VALUES: [Self; 12] = [
+        Self::A,
+        Self::B,
+        Self::C,
+        Self::D,
+        Self::E,
+        Self::F,
+        Self::G,
+        Self::H,
+        Self::I,
+        Self::J,
+        Self::K,
+        Self::L,
+    ];
 
     pub fn iter() -> impl ExactSizeIterator<Item = Self> {
-        use Side::*;
-        [A, B, C, D, E, F, G, H, I, J, K, L].into_iter()
+        Self::VALUES.iter().copied()
     }
 
     /// Whether `self` and `other` share an edge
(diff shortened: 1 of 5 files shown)


common/src/dodeca.rs
```diff
@@ -24,55 +34,63 @@ pub enum Side {
...
-        &reflections_f64()[self as usize]
+        &data::REFLECTIONS_F64[self as usize]
     }
 
     /// Whether `p` is opposite the dodecahedron across the plane containing `self`
     #[inline]
     pub fn is_facing(self, p: &MVector<f32>) -> bool {
-        let r = na::convert::<_, na::RowVector4<f32>>(self.reflection().row(3).clone_owned());
-        (r * na::Vector4::from(*p)).x < p.w
+        self.normal().mip(p) > 0.0
```
patowen, line 93: I cross-checked this with `Graph::normalize_transform`, and while the old implementation by coincident never had an infinite loop with a bunch of fuzz testing (but did have some redundant transformations), with this change, an infinite loop is now possible with the following:
```rust
graph.normalize_transform(
    NodeId::ROOT,
    &math::translate(
        &MVector::origin(),
        &MVector::from(na::Vector4::new(
            0.59122586, 0.8285565, 0.59424436, 1.5187331,
        ))
        .lorentz_normalize(),
    ),
);
```
I will want to revert this change or modify `Graph::normalize_transform` to be more robust. Since `Graph::normalize_transform` is the main use case of this function, a third `buffer` argument may be required to declare how far in front of the face the dodecahedron's face the point must be for this function to return true.

common/src/dodeca.rs
```diff
@@ -400,121 +431,97 @@ mod data {
...
+
+            // Since adjacent normals are already orthogonal, we can use them
+            // as-is for the first three columns of this matrix. We just need to
+            // negate them so that they point towards the origin instead of away
+            // because the dodeca's origin has positive cube-centric
+            // coordinates.
+
+            // As for the last column of the change-of-basis matrix, that would
+            // be the cube-centric origin in dodeca-centric coordinates, or in
+            // other words, the vertex's location in dodeca-centric coordinates.
```
Ralith, line 455: Might also help to explain the significance of the vertex?

common/src/dodeca.rs
```diff
@@ -24,48 +34,57 @@ pub enum Side {
 }
 
 impl Side {
-    #[inline]
-    pub fn from_index(x: usize) -> Self {
-        use Side::*;
-        const VALUES: [Side; SIDE_COUNT] = [A, B, C, D, E, F, G, H, I, J, K, L];
-        VALUES[x]
-    }
+    pub const COUNT: usize = 12;
```
Ralith, line 37: Why not `VALUES.len()`? Similar for `VERTEX`.

patowen: Good question. I probably should have done the same for `common::world::Material` earlier, which I modeled this on.

Ralith: tbf `len` might not have been `const` back then!

common/src/dodeca.rs
patowen: The significance of a dodecahedron's vertex is that it's the origin in cube-centric coordinates. Is the suggestion to say why it's the origin? There are a lot of facts I could write, but I'm not sure which are relevant to this comment:
- It is a vertex of a dodecahedron (already mentioned)
- It is the origin in cube-centric coordinates (already mentioned)
- It is where 8 dodecahedra intersect in the dodecahedral tiling
- It is the center of a cube in the dual cubic tiling

However, I think it is hoped that someone reading this method already knows what dodeca-centric and cube-centric (or node-centric and dual-centric) coordinates are. Perhaps module-level documentation on `dodeca` itself is needed to fully explain this. Thoughts?

This also reminds me that the words "node" and "dual" here are not ideal, and it would probably be good to think about renaming them.

Ralith: Type- or module-level docs might be the way to go, yeah. Skimming this without diving too deep I just couldn't remember why this was a function of vertex at all.

Ralith (approved):

common/src/dodeca.rs
patowen: I ended up adding some documentation to `Vertex` to explain the different coordinate systems better. Because of your approval, I will go ahead and merge this PR after these changes.

PR #444, opened 2025-02-16
Update packages to their latest versions

patowen:
This PR updates the following packages:
- rand
- rand_pcg
- rand_distr
- yakui
- yakui-vulkan
- tokio
- whoami
- webpki

Note that yakui released 0.3.0 recently, allowing us to use it as a cargo dependency instead of a github dependency.

The `nalgebra` package has not yet updated rand, so to avoid conflicts, I have removed the `rand` feature from `nalgebra` for now, replacing a call of `na::Vector3::from_distribution` with a call to `na::Vector3::from_fn` instead.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -14,8 +14,8 @@ server = { path = "../server" }
 tracing = "0.1.10"
 ash = { version = "0.38.0", default-features = false, features = ["loaded", "debug", "std"] }
 lahar = { git = "https://github.com/Ralith/lahar", rev = "7963ae5750ea61fa0a894dbb73d3be0ac77255d2" }
-yakui = { git = "https://github.com/SecondHalfGames/yakui", rev = "136f88d54f08fcb21a5215f422c8581f89d46dd2" }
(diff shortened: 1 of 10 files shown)


client/Cargo.toml
```diff
@@ -14,26 +14,26 @@ server = { path = "../server" }
...
-whoami = "1.2.1"
+whoami = "1.5.2"
 serde = { version = "1.0.104", features = ["derive", "rc"] }
 toml = { workspace = true }
 fxhash = "0.2.1"
 downcast-rs = "2.0.0"
 quinn = { workspace = true }
 futures-util = "0.3.1"
-webpki = "0.22.0"
+webpki = "0.22.4"
```
Ralith, line 36: Updating between semver-compatible versions isn't usually necessary, and may be gratuitous churn. Are these motivated by something in particular?

Ralith (changes requested):

client/Cargo.toml
patowen: VSCode had warned me about https://rustsec.org/advisories/RUSTSEC-2023-0052 with the older version of the package, so I decided to play it safe. There were similar warnings for other packages.

PR #447, opened 2025-02-21
Upgrade to the Rust 2024 edition

patowen:
This is done in two separate commits for easier readability:
- Run `cargo edition --fix` and look at (but do nothing about) the warnings, and then make manual changes based on what was learned from `cargo edition --fix`
- Run `cargo fmt`

The main warnings when running `cargo edition --fix` came from [Tail expression temporary scope](https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html#tail-expression-temporary-scope). There have been some changes to semantics here, which `cargo fix --edition` is unable to correct for, so it warns about it. However, my understanding is that for our use case, the new semantics perfectly make sense and do not need to be corrected for.

The main manual change I made was undoing the automatic changes from [if let temporary scope](https://doc.rust-lang.org/edition-guide/rust-2024/temporary-if-let-scope.html#if-let-temporary-scope), since the new semantics are fine for our use case. I also deleted an outdated comment (`Intermediate variable to make borrowck happy`) that I discovered while looking into the warnings from `cargo edition --fix`.

I ran `cargo fmt` in a separate commit because it makes a large diff for two reasons:
- There were edition changes to `cargo fmt`, most notably the order of imports.
- The edition fixes for [unsafe_op_in_unsafe_fn warning](https://doc.rust-lang.org/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html#unsafe_op_in_unsafe_fn-warning) and new `unsafe` blocks, which increase the level of indentation of many `unsafe` functions we have.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -2,7 +2,7 @@
 name = "client"
 version = "0.1.0"
 authors = ["Benjamin Saunders <ben.e.saunders@gmail.com>"]
-edition = "2021"
+edition = "2024"
 publish = false
 license = "Apache-2.0 OR Zlib"
 
--- a/client/benches/surface_extraction.rs
+++ b/client/benches/surface_extraction.rs
@@ -1,11 +1,11 @@
 use std::sync::Arc;
 
 use ash::vk;
-use bencher::{benchmark_group, benchmark_main, Bencher};
+use bencher::{Bencher, benchmark_group, benchmark_main};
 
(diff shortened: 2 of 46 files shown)


Ralith (approved):
I'd squash the first two commits together. Intermediate states that we immediately revert to the immediately prior state aren't particularly useful. LGTM otherwise.

PR #450, opened 2025-03-26
Improve numerical stability of ground plane distance estimation

### LINKED ISSUE #102: Improve numerical stability of ground plane distance estimation

patowen:
Fixes #102

This PR repurposes plane.rs to serve primarily as the ground plane in worldgen, with the extra logic suggested in #102 (with some minor changes) to handle situations in which the ground plane is far from the origin. The main differences or possible surprises are as follows:
- The normal vector is represented as `scaled_normal * e^exponent` instead of `scaled_normal * 2^exponent`. (a base of `e` instead of 2)
- The exponent is a floating point number instead of an integer. This reduces the amount of casting needed.
- The magnitude of the normal vector is estimated by taking the absolute value of the w-coordinate, as for large enough vectors, the magnitude is proportional to that value.
- The exponent is only updated when the vector becomes very large in magnitude (instead of updating every time the exponent could increase by 1). There's no real downside to this, and it enables another optimization:
- The approximate formula for `sinh(x)` (`ln(abs(x)*2) * signum(x)`) is used whenever the exponent is nonzero, so the decision does not need to determine the magnitude of `x` first.
- It is assumed that repeated transformations to the plane will be applied in such a way that the plane in general gets farther away from the origin (and all the points it's compared to) over time. Going the other way may result in inaccurate calculations because the exponent is set to never decrease. Note that the calculations would have been inaccurate anyway because of catastrophic cancellation.

A few related changes have been bundled with this main change:
- `Frustum` no longer uses `Plane`, since it does not need its more complicated features.
- `Plane`'s generic type parameter has been removed, as it is designed with one floating point type in mind, and there are no real downsides to this simplification anymore.
- All worldgen code (including `Plane`'s implementation) has been switched over to `f32`, as the precision issues of `Plane` were the main blocker to doing this earlier.

### DIFF
--- a/client/src/graphics/frustum.rs
+++ b/client/src/graphics/frustum.rs
@@ -1,5 +1,4 @@
-use common::Plane;
-use common::math::MPoint;
+use common::math::{MDirection, MPoint};
 
 #[derive(Debug, Copy, Clone)]
 pub struct Frustum {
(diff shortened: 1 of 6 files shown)


Ralith (approved):
Nice!

Issue #75, opened 2020-05-03
Generate large-scale structures


Ralith:
One of the best ways to exhibit hyperbolic geometry is with structures that are impossible in euclidean space. We should extend worldgen to produce a variety of such structures, e.g. yendorian trees, whose branches don't decrease in size compared to their parent branch. These structures are characterized by being substantially larger than a graph node; think large buildings, cave systems, and megastructures.

The main challenge here is generating such features deterministically no matter how you approach them, and in defining complex, interesting hyperbolic geometry in a maintainable way.

Ralith:
> The main challenge here is generating such features deterministically no matter how you approach them

We can solve this the same way we generate the noise fields: place each structure 100% behind some inter-node plane with respect to the origin. The graph ensures that a specific node is always the first one populated behind any given inter-node plane, so that node can be responsible for deciding what (if any) large scale structure should be generated.

PR #451, opened 2025-04-11
Uncouple client worldgen from rendering

### LINKED ISSUE #50: More accurate view distance based graph expansion
After moving each character, `Graph::ensure_nearby` is used to expand the graph to contain all nodes whose centers lie within the view distance, and all neighbors of those nodes. Because chunks cannot be populated with terrain until the cell of the cubic honeycomb containing them has all incident nodes generated, this leads to characters being within view distance of chunks that lie within generated nodes but which cannot be populated as they have incident nodes which have not yet been generated.

To correct this without expanding the graph at an unsustainable rate, we need to use a chunk-aware traversal that populates nodes if they are incident to a chunk which is within view distance. To avoid requiring a hyperbolic cube vs. sphere intersection test, we can approximate chunk visibility by considering node vertex visibility, as in #49.

patowen:
Uncoupling world generation from world rendering will be useful to allow flexibility in multiple areas in the future:
* The client's handling of world generation can stick entirely to safe Rust, making it easier to work with for newer contributors.
* The view distance and chunk generation distance can be kept separate, making it easier to find a resolution for #50 without killing performance or creating thick fog (as the user can then opt in or opt out of chunk pop-in).
* Once the client no longer fully relies on the server to decide which nodes should be generated, `WorldgenDriver` can be responsible for deciding how to expand the graph in addition to how to fill out its chunk data.
* Regarding graph traversal, `WorldgenDriver`'s FPS footprint can be reduced by spreading out graph traversal across multiple frames.
* It could also be responsible for storing modified chunks not yet added to the graph, rather than having that data live directly in `Sim` (Done in this PR).
* It could also be responsible for reclaiming memory for chunks far away from the player.
* In short, there's a lot of functionality that makes sense to put together in the general bucket of "manage the voxels and nodes in `Graph`", and decoupling it from rendering allows all that functionality to exist without adding unmanageable complexity to the code.

I am open to alternative names to `WorldgenDriver`, as I'm not convinced it's a good name for this struct. I call it that because it has a `drive` function, which is named that way because it sends and receives data from a `WorkQueue<ChunkDesc>`, making it behave a bit like a `Loader`.

This is part of a series of pull requests I plan to create to hopefully lead up to #75, as I discovered that my WIP branch for that issue was getting somewhat difficult to manage because of this coupling (since I do decouple server and client graph expansion in that branch).

### DIFF
--- a/client/src/worldgen_driver.rs
+++ b/client/src/worldgen_driver.rs
@@ -1,88 +1,73 @@
-use std::{sync::Arc, time::Instant};
+use std::time::Instant;
 
 use common::{
-    dodeca::{self, Vertex},
-    graph::NodeId,
-    math::{MIsometry, MPoint},
+    dodeca::Vertex,
+    graph::{Graph, NodeId},
     node::{Chunk, ChunkId, VoxelData},
+    proto::{BlockUpdate, Position},
+    traversal,
 };
+use fxhash::FxHashMap;
 use metrics::histogram;
-
-use crate::{
-    Config, Sim,
-    graphics::{Base, Frustum},
-    loader::{Cleanup, LoadCtx, LoadFuture, Loadable, Loader, WorkQueue},
-};
+use tokio::sync::mpsc;
 
 pub struct WorldgenDriver {
-    work_queue: WorkQueue<ChunkDesc>,
+    work_queue: WorkQueue,
+    /// Voxel data that have been downloaded from the server for chunks not yet introduced to the graph
+    preloaded_block_updates: FxHashMap<ChunkId, Vec<BlockUpdate>>,
 }
 
 impl WorldgenDriver {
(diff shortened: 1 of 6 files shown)


client/src/worldgen_driver.rs
```diff
@@ -91,6 +76,15 @@ impl WorldgenDriver {
         }
         histogram!("frame.cpu.drive_worldgen").record(drive_worldgen_started.elapsed());
     }
+
+    pub fn apply_block_update(&mut self, graph: &mut Graph, block_update: BlockUpdate) {
+        if !graph.update_block(&block_update) {
```
Ralith, line 81: style nit: invert and early return to reduce rightward drift

Ralith (approved):

PR #452, opened 2025-04-13
Make the server and client expand the graph independently

patowen:
The main change this PR makes is to ensure that the server and client can expand their respective graphs mostly independently (with the main exception being that the server needs to tell the client where its character is at the beginning).

### DIFF
--- a/client/src/sim.rs
+++ b/client/src/sim.rs
@@ -377,7 +377,10 @@ impl Sim {
             metrics::declare_ready_for_profiling();
         }
         for node in &msg.nodes {
-            self.graph.insert_neighbor(node.parent, node.side);
+            // We need to get a list of nodes from the server, especially on first log-in,
+            // since otherwise, we won't be able to know where the local character is with
+            // just the NodeId alone.
+            self.graph.ensure_neighbor(node.parent, node.side);
         }
         populate_fresh_nodes(&mut self.graph);
         for block_update in msg.block_updates.into_iter() {
--- a/client/src/sim.rs
+++ b/client/src/sim.rs
@@ -389,7 +389,8 @@ impl Sim {
                 tracing::error!("Voxel data received from server is of incorrect dimension");
                 continue;
             };
-            self.graph.populate_chunk(chunk_id, voxel_data);
+            self.worldgen_driver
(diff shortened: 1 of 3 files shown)


server/src/sim.rs
```diff
@@ -478,6 +479,9 @@ impl Sim {
                 .voxel_data
                 .push((chunk_id, voxels.serialize(self.cfg.chunk_size)));
         }
+        for (&chunk_id, voxels) in self.preloaded_voxel_data.iter() {
+            (spawns.voxel_data).push((chunk_id, voxels.serialize(self.cfg.chunk_size)));
```
Ralith, line 483: Extraneous parens?

client/src/sim.rs
```diff
@@ -377,7 +377,10 @@ impl Sim {
             metrics::declare_ready_for_profiling();
         }
         for node in &msg.nodes {
-            self.graph.insert_neighbor(node.parent, node.side);
+            // We need to get a list of nodes from the server, especially on first log-in,
+            // since otherwise, we won't be able to know where the local character is with
+            // just the NodeId alone.
+            self.graph.ensure_neighbor(node.parent, node.side);
```
Ralith, line 383: What's the significance of the change from `insert_neighbor`?

server/src/sim.rs
patowen: Maybe. I mainly added them to get rustfmt not to unnecessarily splay things out onto multiple lines.

client/src/sim.rs
patowen: Since the client can choose to generate nodes, we can no longer guarantee that the server won't send a node the client already knows about.

server/src/sim.rs
Ralith: I think having idiosyncratic syntax is worse than having a few extra linebreaks. If you want things more compact, better to tune the rustfmt config for the whole workspace, I think.

patowen: Fair point. I'll go ahead and revert this. I wonder why there's no clippy lint for it.

Ralith:
Merged with #453.

patowen:
> Merged with #453.

I went ahead and merged the PR title/description as well just in case it's useful for anyone looking at history.

PR #453, opened 2025-04-13
Refactor graph expansion to remove the concept of "fresh" nodes

patowen:
Note: The first few (2 as of the time I'm writing this) commits match https://github.com/Ralith/hypermine/pull/452, so this PR effectively branches off of https://github.com/Ralith/hypermine/pull/452, and only the later commits should be reviewed.

Since keeping the server and client graphs entirely in sync is one of the main reasons `Graph` needed to keep track of its freshly created nodes, and now they no longer need to be in sync this way, this PR removes this functionality to simplify `Graph`'s usage. Specifically, each `Node` is initialized immediately when it's added to the graph instead of being initialized to `None`. The `NodeState` field, which is used in fewer places, is still kept optional, and instead of being populated for every fresh node, it is populated for a given `NodeId` with an `ensure_node_state` function.

Part of why `NodeState` is kept optional is that its initialization order will need to be more complicated for megastructures, although there may be a way in the future to make things simpler on that front as well.

Since this PR makes heavy changes to graph expansion, it also addresses most of #50 by making the "distance" computations correct. However, with the current level of optimization, it is impossible to avoid the following three problems: unacceptably thick fog, pop-in, and unacceptable performance issues. Therefore, we still allow pop-in by making fog distance, view distance, and chunk generation distance separate settings with different defaults in the config.

### DIFF
--- a/client/src/worldgen_driver.rs
+++ b/client/src/worldgen_driver.rs
@@ -3,6 +3,7 @@ use std::time::Instant;
 use common::{
     dodeca::{self, Vertex},
     graph::{Graph, NodeId},
+    math::MPoint,
     node::{Chunk, ChunkId, VoxelData},
     proto::{BlockUpdate, Position},
     traversal,
@@ -41,24 +42,13 @@ impl WorldgenDriver {
             // there's no point trying to generate chunks.
             return;
         }
-        // An extra dodeca::BOUNDING_SPHERE_RADIUS is needed here because we need to generate
-        // enough chunks to ensure that the all chunks within chunk_generation_distance can be filled
-        // with world generation, which requires all nodes surrounding its vertex to be in the graph.
-        traversal::ensure_nearby(
-            graph,
-            &view,
-            chunk_generation_distance + dodeca::BOUNDING_SPHERE_RADIUS,
-        );
-        for (node_id, _) in traversal::nearby_nodes(
-            graph,
-            &view,
-            chunk_generation_distance + dodeca::BOUNDING_SPHERE_RADIUS,
-        ) {
-            graph.ensure_node_state(node_id);
-        }
+        let local_to_view = view.local.inverse();
+
+        traversal::ensure_nearby(graph, &view, chunk_generation_distance);
         let nearby_nodes = traversal::nearby_nodes(graph, &view, chunk_generation_distance);
 
-        'nearby_nodes: for &(node, _) in &nearby_nodes {
+        'nearby_nodes: for &(node, ref node_transform) in &nearby_nodes {
+            let node_to_view = local_to_view * node_transform;
             for vertex in Vertex::iter() {
                 let chunk_id = ChunkId::new(node, vertex);
 
@@ -66,12 +56,16 @@ impl WorldgenDriver {
                     continue;
                 }
 
-                let Some(params) =
-                    common::worldgen::ChunkParams::new(graph.layout().dimension(), graph, chunk_id)
-                else {
+                // Skip chunks beyond the chunk generation distance
+                if (node_to_view * vertex.chunk_bounding_sphere_center())
+                    .distance(&MPoint::origin())
+                    > chunk_generation_distance + dodeca::CHUNK_BOUNDING_SPHERE_RADIUS
+                {
                     continue;
-                };
+                }
 
+                // Generate voxel data
+                let params = common::worldgen::ChunkParams::new(graph, chunk_id);
                 if let Some(voxel_data) = self.preloaded_voxel_data.remove(&chunk_id) {
                     self.add_chunk_to_graph(graph, chunk_id, voxel_data);
                 } else if self.work_queue.load(ChunkDesc { node, params }) {
--- a/client/src/worldgen_driver.rs
+++ b/client/src/worldgen_driver.rs
@@ -3,7 +3,7 @@ use std::time::Instant;
 use common::{
     dodeca::{self, Vertex},
     graph::{Graph, NodeId},
-    node::{self, Chunk, ChunkId, VoxelData},
+    node::{Chunk, ChunkId, VoxelData},
     proto::{BlockUpdate, Position},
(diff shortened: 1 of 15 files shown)


Ralith:
> unacceptably thick fog

Tangential, but I wonder if we can find a better balance by using nonphysical fog that thickens rapidly but has a distant start point...

common/src/traversal.rs
```diff
@@ -11,8 +11,10 @@ use crate::{
     proto::Position,
 };
 
-/// Ensure all nodes within `distance` of `start` exist
+/// Ensure all nodes exist whose bounding spheres are within `distance` of `start`
```
Ralith, line 14: We lost the word "exist" here.

common/src/node.rs
```diff
@@ -28,6 +28,33 @@ impl ChunkId {
...
 impl Graph {
+    /// Returns the NodeState for the given node, panicking if it isn't initialized.
+    #[inline]
+    pub fn node_state(&self, node_id: NodeId) -> &NodeState {
+        self[node_id].state.as_ref().unwrap()
+    }
+
+    /// Initializes the NodeState for the given node and all its ancestors if not
+    /// already initialized.
+    pub fn ensure_node_state(&mut self, node_id: NodeId) {
```
Ralith, line 39: We often seem to need to access the state immediately after ensuring it. Would that code be more concise if this function returned the `&mut NodeState` directly?

client/src/worldgen_driver.rs
```diff
@@ -41,37 +42,31 @@ impl WorldgenDriver {
...
-            chunk_generation_distance + dodeca::BOUNDING_SPHERE_RADIUS,
-        );
-        for (node_id, _) in traversal::nearby_nodes(
-            graph,
-            &view,
-            chunk_generation_distance + dodeca::BOUNDING_SPHERE_RADIUS,
-        ) {
-            graph.ensure_node_state(node_id);
-        }
+        let local_to_view = view.local.inverse();
```
Ralith, line 45: nit: let's try to adopt a convention of `x_from_y` when naming transforms. This makes chains much easier to read and sanity-check because it matches the multiplication order: `x_from_y * y_from_z` is a lot more obviously correct than `y_to_x * z_to_y`.

client/src/worldgen_driver.rs
```diff
@@ -41,37 +42,31 @@ impl WorldgenDriver {
...
                 let chunk_id = ChunkId::new(node, vertex);
 
                 if !matches!(graph[chunk_id], Chunk::Fresh) {
                     continue;
                 }
 
-                let Some(params) =
-                    common::worldgen::ChunkParams::new(graph.layout().dimension(), graph, chunk_id)
-                else {
+                // Skip chunks beyond the chunk generation distance
```
Ralith, line 59: Does `nearby_nodes(..., chunk_generation_distance)` already accomplish this? If it's not intended to, explain in a comment.

common/src/worldgen.rs
```diff
@@ -181,22 +181,22 @@ pub struct ChunkParams {
...
 
 impl ChunkParams {
-    /// Extract data necessary to generate a chunk
-    ///
-    /// Returns `None` if an unpopulated node is needed.
-    pub fn new(dimension: u8, graph: &Graph, chunk: ChunkId) -> Option<Self> {
-        let state = graph[chunk.node].state.as_ref()?;
-        Some(Self {
+    /// Extract data necessary to generate a chunk, generating new graph nodes if necessary
+    pub fn new(dimension: u8, graph: &mut Graph, chunk: ChunkId) -> Self {
```
Ralith, line 185: Can we remove the `dimension` argument in favor of fetching it from the graph directly?

Ralith (changes requested):

common/src/traversal.rs
Ralith: Wait, no we didn't.

common/src/node.rs
patowen: It should only need to return a `&NodeState`, but otherwise, that seems like a good suggestion to me.

client/src/worldgen_driver.rs
patowen: Oh, that's an interesting approach. Maybe in a separate PR? The `dodeca` module has a lot of functions we would also want to change.

client/src/worldgen_driver.rs
patowen: `nearby_nodes` only returns nodes. We would need a `nearby_chunks` function to be as precise as this.

common/benches/bench.rs
```diff
@@ -39,7 +39,7 @@ fn build_graph(c: &mut Criterion) {
     c.bench_function("worldgen", |b| {
         b.iter(|| {
             let mut graph = Graph::new(12);
-            ensure_nearby(&mut graph, &Position::origin(), 3.0);
+            ensure_nearby(&mut graph, &Position::origin(), 2.25);
```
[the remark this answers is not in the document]
patowen: `ensure_nearby` was modified to return more nodes for a given distance, so this benchmark needed to be adjusted accordingly.

common/src/traversal.rs
patowen: Oh yeah, the grammar feels a bit weird, but I thought it would be easier to read, since a sentence ending in "`start` exist" might end up being a garden path sentence.

patowen:
> > unacceptably thick fog
> 
> Tangential, but I wonder if we can find a better balance by using nonphysical fog that thickens rapidly but has a distant start point...

Possibly. I'd have to try it out. I do plan on trying to better optimize worldgen in the future, and I'll have to double-check that we're not including too many chunks the user wouldn't see past the fog.

client/src/worldgen_driver.rs
Ralith: Sure.

PR #455, opened 2025-04-17
Fix package-linux workflow by removing unsupported Ubuntu 20.04

patowen:
The Package/Linux workflow is currently failing on master, being auto-cancelled with the following error:
> This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -43,14 +43,15 @@ jobs:
   package-linux:
     name: Linux
     # Oldest supported runner, for wide glibc compat
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
         with:
           lfs: true
       - name: Install dependencies
         run: sudo apt update && sudo apt-get -y install libasound2-dev libvulkan-dev libfontconfig-dev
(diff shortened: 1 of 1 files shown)


.github/workflows/package.yml
```diff
@@ -43,14 +43,15 @@ jobs:
...
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v4
         with:
           lfs: true
       - name: Install dependencies
         run: sudo apt update && sudo apt-get -y install libasound2-dev libvulkan-dev libfontconfig-dev
-      # No prebuilt shaderc, since the official binaries don't to be compatible with Ubuntu 20.04.
+      # No prebuilt shaderc, since the official binaries don't seem to be compatible with Ubuntu 20.04,
```
patowen, line 53: Is this something we should address in this PR? I'm wondering if it's possible to use a prebuilt shaderc with Ubuntu 22.04.

patowen:
Currently testing this on https://github.com/patowen/hypermine/pull/21

.github/workflows/package.yml
Ralith: It'd speed up the builds a lot, but no need for it to happen in this PR.

PR #456, opened 2025-05-24
Clippy: Fix result_large_err by boxing redb errors

patowen:
The `ReadTransactionStillInUse` error variant of the redb error type is 160 bytes long, with most of its size coming from the `redb::tree_store::TableTree` type.

An alternative implementation would be to ignore the error or add a "clippy.toml" to increase the error threshold to 161, but 160 bytes is quite long for a `Result` type, so boxing the error would likely be the better design.

Note (for anyone reading the comment thread): This PR is a revised version of a PR that initially just ignored the clippy lint in the places it triggered from.

### DIFF
--- a/save/src/lib.rs
+++ b/save/src/lib.rs
@@ -37,7 +37,7 @@ impl Save {
                     init_meta_table(&db, &defaults)?;
                     defaults
                 }
-                Err(e) => return Err(OpenError::Db(DbError(e.into()))),
+                Err(e) => return Err(OpenError::Db(DbError(Box::new(e.into())))),
             }
         };
         Ok(Self { meta, db })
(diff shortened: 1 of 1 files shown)


patowen:
No particular reason from me. I don't have a good sense of how much we care about the error sizes, and I didn't expect that it would be a bottleneck, so I went with the easiest approach (of not changing the code). I also checked the redb examples, and they also took this shortcut (https://github.com/cberner/redb/pull/974/files).

Boxing the error does seem better though. I need to figure out how to do this in an idiomatic way that interacts with thiserror. 160 bytes is indeed very big.

Ralith (approved):

PR #457, opened 2025-06-01
Rename "length" to "depth" and make "parent" more useful

patowen:
The term "depth" is arguably more intuitive when expressing how far away from the root a graph's node is, so this should hopefully reduce the barrier to entry for learning the Hypermine codebase.

As for "parent", previously, it was often interpreted as though every node (except the root) has one parent, even though most logic took all "shorter neighbors" into account, so "parent" now refers to any neighbor node closer to the root. The old meaning of parent is renamed to "primary parent", which is the parent with the first side in enum order.

Given that I hope to add the new terminology of "peer" nodes into the code base, I think it makes sense to rethink existing terms to make sure they're consistent with each other and with what one might expect. I would like to rename some of the terminology in `dodeca` as well, but I'm less sure about the right terminology, so I'll need to think on it some more and/or open an issue.

### DIFF
--- a/client/src/graphics/voxels/mod.rs
+++ b/client/src/graphics/voxels/mod.rs
@@ -181,7 +181,7 @@ impl Voxels {
                             }
                         }
                     }
-                    let node_is_odd = sim.graph.length(node) & 1 != 0;
+                    let node_is_odd = sim.graph.depth(node) & 1 != 0;
                     extractions.push(ExtractTask {
                         index: scratch_slot,
                         indirect_offset: self.surfaces.indirect_offset(slot.0),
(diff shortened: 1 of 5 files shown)


Ralith (approved):
I wonder if "shallower neighbor" would be a better term than "parent"? There's not really a hierarchy per se, after all.

patowen:
> I wonder if "shallower neighbor" would be a better term than "parent"? There's not really a hierarchy per se, after all.

Definitely worth considering. My current preference is "parent" for two reasons: (1) It's shorter, making it less likely to result in long difficult-to-parse phrases when explaining certain logic, and (2) there is a hierarchy worth considering, as a node is not allowed to be added to the graph until all of its parents are added first.

PR #459, opened 2025-11-22
Add outline for detailed documentation

patowen:
The Hypermine code base has become large enough that it is not always clear where a new potential contributor should start. It's also easy for existing contributors (like me) to have to relearn parts of the code that we are less familiar with.

I believe this project would benefit from some kind of documentation outside the main codebase (but still managed by pull requests). Such documentation would have the following benefits:
* A simple "How to play" whose history is synced with the code would be useful.
* We get a logical order of what documentation to read, helping to provide an initial understanding to allow the reader to find the relevant code and its documentation for the question they are trying to ask.
* When we are not limited to code comments, enhanced documentation options, such as diagrams, animations, or potentially even interactive elements could be a possibility.

I have not yet worked out the logistics of including diagrams etc. in the documentation yet. For documentation maintainability and for keeping the repo size down (to avoid running into limits), none of these diagrams should be checked into source control. Instead, code that can generate these diagrams consistently should be included instead. I am interested in suggestions on which graphics APIs such code should use (or if there are other ideas on how to handle this problem in general).

Note that this PR just adds the outline for the documentation, along with ideas and suggestions on how it should be written. While all this information could live in a GitHub issue, I believe this PR format makes it easier to discuss, and having the outline in the repo itself makes it easier to track changes to these plans in general. One downside is that CI runs unnecessarily. It might be possible to configure it not to do this.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,91 @@
+# Current outline
+This is subject to change.
+* Introduction
+* How to play Hypermine
+    * Controls
+    * Config file
+    * The save file (with a warning that compatibility between versions of Hypermine are not guaranteed and another warning to back up save files)
+    * Setting up multiplayer
+* Background math
+    * Linear algebra
+        * (This can link to external resources, but readers should be guided on what parts of linear algebra are worth learning, and making these docs self-contained would be a good long-term goal. If we do use external links, we should include a date so that readers know when health of each link was last checked.)
+        * Vectors
+        * Matrices, matrix-vector multiplication, and its meaning
+        * Matrix-matrix multiplication and its meaning
+            * (We should likely explain both the "transformation" and "change of basis" interpretations)
+        * 3D examples (assuming previous sections have used 2D examples)
+        * Dot products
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -0,0 +1,88 @@
...
+        * (We want such a section because it's a good segue to problem solving techniques for hyperbolic geometry problems)
+        * Representing points as unit vectors
+        * Projections, reflections, rotations, translations (which are rotations expressed differently)
+    * Hyperbolic geometry
+        * Minkowski space (with the "inner product" as the main difference)
+        * Representing points as "normalized" vectors
+        * Projections, reflections, rotations, translations, horo-rotations
+        * A note about floating point precision
+        * (Do we put advanced shape-casting math here? Probably not.)
+* Tiling the world
```
Ralith: I propose leaning heavily on the analogous 2D tiling for ease of visualization wlog, once the analogy has been established.

Ralith (approved):

PR #461, opened 2025-12-14
Add documentation summarizing the world generation algorithm

patowen:
Rather than writing documentation in order, my plan is to prioritize what I believe to be the least-well-understood parts of Hypermine that I am familiar with. I also plan to write it in Markdown with TODO placeholders for images, as I believe that this would be the most efficient way for me to write the most essential parts of the documentation with the fewest distractions.

If/when it's time to add visualizations, I (or whoever adds them) should hopefully be able to focus entirely on the "how" rather than the "what".

As this documentation is meant to be understandable, please let me know if any parts of it are confusing, and I can try to reword it (or if it is fundamentally difficult to understand, I can try to make sure that there are enough placeholder images).

### DIFF
--- a/docs/world_generation.md
+++ b/docs/world_generation.md
@@ -1,0 +1,73 @@
+# World generation
+World generation in Hypermine is constrained the following principles:
+* Consistency: The contents of a chunk when it is first visited should be determined only by the chunk's location and the world generation parameters, which never change within a world.
+* Isotropy: World generation should be as isotropic as reasonably possible, so that it is not obvious which direction leads to the origin.
+
+These constraints help inspire the details of the world generation algorithm, which are explained further in the sections below.
+
+## Noise
+Hypermine relies on a random noise function for multiple purposes, such as the deciding the shape of the terrain and what material the ground should be. Because of this, to understand Hypermine's world generation, it's important to understand the noise function it uses.
+
+Hyperbolic space has a unique challenge compared to Euclidean space: Common approaches to generating good-looking terrain, such as fractal noise, rely on using grids of multiple different scales, while in hyperbolic space, grids cannot be arbitrarily scaled. This would cause terrain features to have a limited size if generated this way. To avoid this limitation, Hypermine uses a different approach specifically designed for a hyperbolic tiling, inspired by Hyperrogue, which is described in the following section.
+
+### Coarse noise function output
+The first step is to form a coarse approximation of the noise function, deciding on one value for each node. To do this, we take advantage of the tiling itself. For a 2D analogy, the pentagonal tiling of the hyperbolic plane can be thought of as a set of lines dividing the hyperbolic plane instead of individual pentagons.
+
+TODO: Picture of pentagonal tiling with lines colored to distinguish them from each other
+
+Similarly, in 3D, the dodecahedral tiling can be thought of as a set of planes dividing hyperbolic space. This interpretation of the dodecahedral tiling is important for understanding how the noise function works between nodes.
+
+To decide on a noise value for each node, we break the dodecahedral tiling up into these planes. We associate each plane with a randomly chosen noise delta, such that crossing a specific plane in one direction increases or decreases the noise value by a specific amount, and crossing the same plane from the other side has the opposite effect. Once we decide on a noise value for the root node, this definition fully determines the noise value of every other node.
+
+The following diagram shows an example of the 2D equivalent of this algorithm.
+
+TODO: Picture of pentagonal tiling with each line labeled with the noise delta, using arrows or something similar to show how this offset applies. The center of each pentagon is also labeled with a number with its current noise value. Integers are used everywhere to allow the reader to verify the math easily in their head.
+
+In this diagram, the randomly-chosen noise delta of each line, along with the derived noise value of each node, is shown. Note how the difference in noise values between any two adjacent nodes matches the noise delta of the line dividing them.
+
+This algorithm allows for random variation while keeping nearby nodes similar to each other, which is what we need from a noise function. One notable quirk worth mentioning is that the noise value is unbounded, which currently means that hills and valleys in Hypermine can become arbitrarily high and deep, respectively.
+
+### Fine noise function output
+The next step is to use this coarse approximation of the noise function to produce the actual noise function. To begin, set the noise value at the center of each node to the coarse output we computed earlier.
+
+TODO: Picture of pentagonal tiling with a color at the center of each node to represent the noise value.
+
+Then, trilinearly interpolate these values based on voxel coordinates to create a continuous function across the world. Note that in 2D, this would be bilinear interpolation.
+
+TODO: Picture of the same pentagonal tiling with gradients added. Decorate the center of each node with a dot to highlight the control points of the interpolation.
+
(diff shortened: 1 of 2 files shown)


docs/world_generation.md
```diff
@@ -0,0 +1,75 @@
+# World generation
+World generation in Hypermine is constrained the following principles:
+* The generated contents of each chunk must depend only on the parameters of the world generation algorithm, not on gameplay.
```
Ralith: It's not obvious to me what point this is trying to make. How could initial chunk contents depend on gameplay?

docs/world_generation.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Coarse noise function output
+The first step is to form a coarse approximation of the noise function, deciding on one value for each node. To do this, we take advantage of the tiling itself. For a 2D analogy, the pentagonal tiling of the hyperbolic plane can be thought of as a set of lines dividing the hyperbolic plane instead of individual pentagons.
+
+TODO: Picture of pentagonal tiling with lines colored to distinguish them from each other
+
+Similarly, in 3D, the dodecahedral tiling can be thought of as a set of planes dividing hyperbolic space. This interpretation of the dodecahedral tiling is important for understanding how the noise function works between nodes.
+
+To decide on a noise value for each node, we break the dodecahedral tiling up into this planes. We associate each plane with a randomly chosen noise value offset, such that crossing a specific plane in one direction increases or decreases the noise value by a specific amount, and crossing the same plane from the other side has the opposite effect. Once we decide on a noise value for the root node, this definition fully determines the noise value of every other node.
```
Ralith: editorial: 
- "into this planes" is probably not what you intended.
- "value offset" feels like reinventing the term "delta" more obscurely, but maybe that's just me

docs/world_generation.md
```diff
@@ -0,0 +1,75 @@
...
+### Fine noise function output
+The next step is to use this coarse approximation of the noise function to produce the actual noise function. To begin, set the noise value at the center of each node to the coarse output we computed earlier.
+
+TODO: Picture of pentagonal tiling with a color at the center of each node to represent the noise value.
+
+Then, trilinearly interpolate these values based on voxel coordinates to create a continuous function across the world. Note that in 2D, this would be bilinear interpolation.
+
+TODO: Picture of the same pentagonal tiling with gradients added. Decorate the center of each node with a dot to highlight the control points of the interpolation.
+
+Finally, for each voxel, add a random offset to its noise value, drawn independently from some distribution.
```
Ralith: Might bear mention how we seed that randomness.

docs/world_generation.md
```diff
@@ -0,0 +1,75 @@
...
+
+Note that generating larger trees requires a more complicated algorithm that has not yet been planned out or implemented.
+
+## Implementation details
+Hypermine generates worlds in three main stages:
+* Generate a `NodeState` for each node
+* Generate a `ChunkParams` for each chunk, based on the `NodeState` of all 8 nodes adjacent to the chunk's origin
+* Using the information in `ChunkParams`, asynchronously generate the voxel data for each chunk
+
+The reasoning behind these three phases can be seen most clearly in the noise generation algorithm. The coarse noise function created at the beginning of the algorithm is stored in the `NodeState`. Since the noise needs to be interpolated between adjacent `NodeState`s, the `ChunkParams` needs all 8 adjancent nodes to be constructed. The actual interpolation and the inclusion of additional noise is done in the final, asynchronous phase of the algorithm, which depends entirely on `ChunkParams`.
```
Ralith: Maybe mention why we bother making this async?

Ralith:
The required checks should be in the repository settings, which you should have access to. Disabling that would be a fine expediency, IMO.

patowen:
> The required checks should be in the repository settings, which you should have access to. Disabling that would be a fine expediency, IMO.

For some reason, when I go to https://github.com/Ralith/hypermine/settings, I just get a 404. I believe it might be because I'm a "contributor" and not an "owner".

docs/world_generation.md
patowen: First bullet point: Whoops. Good catch. This is why peer review is important (I proofread it, but I think I just read it as "into these planes", which is what I should have written)

Second bullet point: I like the idea of "delta". In an earlier draft, I had thought of using "temperature" as a concrete example of the noise function, but after I wrote out more of the docs, I found that it wasn't as helpful as I expected, so I switched from the word "temperature" to "noise value". However, I do like the brevity of "noise delta" instead of "noise value offset".

docs/world_generation.md
```diff
@@ -0,0 +1,75 @@
+# World generation
+World generation in Hypermine is constrained the following principles:
+* The generated contents of each chunk must depend only on the parameters of the world generation algorithm, not on gameplay.
+* Everything must be generated in its settled state to ensure that unmodified terrain does not need to be stored on disk, and to help with immersion.
+* World generation should be as isotropic as reasonably possible, so that it is not obvious which direction leads to the origin.
+
+These constraints help inspire the details of the world generation algorithm, which are explained further in the sections below.
+
+## Noise
+Every procedural world generation algorithm needs to start with some way of generating noise. This noise is useful for determining properties like temperature and rainfall (which affects what material the ground is made of), and it is also used in determining the shape of the terrain. While Perlin noise is often used as a standard approach, it is not clear how it would be adapted to a large hyperbolic world. Instead, Hypermine uses a different approach specifically designed for a hyperbolic tiling, inspired by Hyperrogue.
```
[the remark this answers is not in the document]
patowen: I had originally learned about Perlin noise from <https://web.archive.org/web/20080724063449/http://freespace.virgin.net/hugo.elias/models/m_perlin.htm>, which is linked to from <https://en.wikipedia.org/wiki/Value_noise> as an example of a misinterpretation of Perlin noise. My whole world is turned upside-down (EDIT: Although, based on <https://en.wikipedia.org/wiki/Talk:Perlin_noise#Common_mistakes>, it seems like this mistake was hard to avoid, and that perhaps Perlin noise is ill-defined. Either way, I have never gotten familiar with gradient noise).

Given this revelation, I might not be qualified to make general statements like this, and I might need to simplify this part of the docs to just say what Hypermine does. However, it would be nice to have some kind of motivation to make these docs more memorable. Perhaps this is worth a TODO that someone more knowledgeable than me can fill out (or I can fill out after doing more research).

Thanks for the call-out!

That being said, I'm still not sure whether Perlin (or value) noise could be readily adapted. We would need a regular grid of points, which is not so easy in hyperbolic space, especially if we want mountains larger than a handful of nodes in size. The high frequency components are doable, but the low frequency components would need something else.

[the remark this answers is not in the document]
Ralith: Gradient noise is, by design, band-limited. You're thinking of wideband noise that's usually achieved by composing many layers of differently scaled gradient noise (often termed "fractal brownian motion" for some reason), which indeed would be difficult to adapt.

docs/world_generation.md
Ralith: Alternatively, we could avoid talking about how/when the computation takes place entirely in this section, and focus on the theoretical model.

patowen:
It looks like I was able to merge it without an additional push. Thanks for the review!

PR #463, opened 2026-01-18
Upgrade crates for January 2026

patowen:
This PR updates the following crates:
criterion: 0.5 -> 0.8
hecs: 0.10 -> 0.11
nalgebra: 0.33 -> 0.34
png: 0.17 > 0.18
prost: 0.13 -> 0.14
prost-build: 0.13 -> 0.14
rcgen: 0.13 -> 0.14
redb: 2 -> 3
toml: 0.8 -> 0.9
whoami: 1 -> 2

It also replaces bincode 1.2.1 with postcard 1.0.4, as bincode is no longer being maintained, and postcard was already in use in this codebase.

### DIFF
--- a/client/src/net.rs
+++ b/client/src/net.rs
@@ -43,6 +43,12 @@ async fn run(
     endpoint.set_default_client_config(client_cfg);
 
     let result = inner(cfg, incoming, outgoing, endpoint.clone()).await;
+    // Close the connection with a generic message, as graceful disconnections are not yet implemented.
+    // See https://github.com/Ralith/hypermine/issues/26
+    endpoint.close(
+        connection_error_codes::CLIENT_CLOSED_CONNECTION,
+        b"client closed connection",
+    );
     endpoint.wait_idle().await;
     result
 }
--- a/server/src/sim.rs
+++ b/server/src/sim.rs
@@ -390,7 +390,20 @@ impl Sim {
         self.world
             .insert_one(entity, InactiveCharacter(character))
             .unwrap();
-        self.accumulated_changes.despawns.push(entity_id);
+        if let Some(index) = self
+            .accumulated_changes
+            .spawns
+            .iter()
+            .position(|e| *e == entity)
+        {
+            // Ensure that the same entity does not show up in the spawns
+            // and despawns list if the character entity is spawned and deactivated
+            // in the same frame. This can happen if a client connects and
+            // immediately disconnects due to an error.
+            self.accumulated_changes.spawns.remove(index);
+        } else {
+            self.accumulated_changes.despawns.push(entity_id);
+        }
     }
 
     fn spawn(&mut self, bundle: impl DynamicBundle) -> (EntityId, Entity) {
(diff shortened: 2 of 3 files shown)


patowen:
@Ralith, it looks like bincode is just used in [codec.rs](https://github.com/Ralith/hypermine/blob/ffc1ad2b81e0ad790e7f346100a9d6a13bab4ba7/common/src/codec.rs) for quinn. The replacement suggestions are wincode, postcard, and rkyv.

As you have worked extensively on quinn, I expect that you have thoughts on what libraries should be used for networking, so I wanted your thoughts before looking into making any kind of switch.

patowen:
postcard seems to work well. Thank you for the suggestion. I had gotten tripped up by assuming that the tag with the length of each message was 4 bytes instead of 3, and the resulting parsing error ended up not leaving any logs, tripping me up further.

For the latter issue, I bundled additional logging in this PR in case something like this happens again.

client/src/net.rs
```diff
@@ -75,15 +75,21 @@ async fn inner(
 
     // Receive the server's hello message
     let hello = codec::recv::<proto::ServerHello>(&mut ordered)
-        .await?
+        .await
+        .inspect_err(|e| {
+            tracing::error!("Error when receving proto::ServerHello from server: {}", e)
```
Ralith: Why log here? The error should be getting returned regardless, and logged somewhere up the stack as a result, which means this is almost certainly a duplicate log message. If it's missing context, we should attach context to the error with e.g. `.context(...)` rather than duplicating the logging.

client/src/graphics/gltf_mesh.rs
```diff
@@ -411,7 +411,7 @@ async fn load_material(
         }
     };
     let mut color_data = &color_data[..];
-    let mut color_reader = png::Decoder::new(&mut color_data)
+    let mut color_reader = png::Decoder::new(Cursor::new(&mut color_data))
```
Ralith: Why this change? `&[u8]` already implements `BufRead`.

server/src/sim.rs
```diff
@@ -390,7 +390,20 @@ impl Sim {
...
+            .accumulated_changes
+            .spawns
+            .iter()
+            .position(|e| *e == entity)
+        {
+            // Ensure that the same entity does not show up in the spawns
+            // and despawns list if the character entity is spawned and deactivated
+            // in the same frame. This can happen if a client connects and
+            // immediately disconnects due to an error.
+            self.accumulated_changes.spawns.remove(index);
```
Ralith: nit: fix to unrelated bug.

server/src/sim.rs
```diff
@@ -390,8 +390,8 @@ impl Sim {
         self.world
             .insert_one(entity, InactiveCharacter(character))
             .unwrap();
-        self.accumulated_changes.despawns.push(entity_id);
-    }
+            self.accumulated_changes.despawns.push(entity_id);
+        }
```
Ralith: Stray change?

PR #479, opened 2026-02-05
Gradient propgation

DimpyRed:
This is a demo introducing environmental gradients that are locally similar between nodes, and applies elevation to them. As a feature, it's complete enough to add to the game.

Code may need some cleaning, and there may need to be some optimization (worldgen is noticeably slower), but I'd like some other people's eyes and opinions on this for those activities.

### DIFF
--- a/common/src/worldgen/mod.rs
+++ b/common/src/worldgen/mod.rs
@@ -658,7 +658,7 @@ impl EnviroFactors {
                 MDirection::<f32>::new_unchecked(v[0], v[1], v[2], 0.0)
             },
             10.0,
-            0.0, //rng.sample(unif),
+            rng.sample(unif),
         );
         let average_gradient = parent
             .elevation_gradient


common/src/worldgen/mod.rs
```diff
@@ -112,10 +113,26 @@ impl NodeState {
                 temperature: 0.0,
                 rainfall: 0.0,
                 blockiness: 0.0,
+                elevation_gradient: EnviroGradient::new_from_direction(1.5_f32, *Side::B.normal()),
             },
             (Some(parent), None) => {
                 let spice = graph.hash_of(node) as u64;
-                EnviroFactors::varied_from(parent.node_state.enviro, spice)
+                let (traversal_direction, up_direction) = {
```
Ralith: It looks like we're expanding a small amount of information into a much larger amount before passing it all through a function call. Could we push all this down into `EnviroFactors::varied_from`? Maybe pass `parent` wholesale?

common/src/worldgen/mod.rs
```diff
@@ -530,32 +549,173 @@ struct NeighborData {
...
+const ELEVATION_GRADIENT_MAGNITUDE_CEILING: f32 = 3.0; // maximum 27.0
+
+#[derive(Copy, Clone)]
+/// A gradient representing evironmental factors.
+/// Consists of a magnitude and direction, which are for the most part operated on indpendently.
+/// The magnitude can be scaled arbitrarily, for example the elevation gradient is on a cubic scale.
+struct EnviroGradient {
+    magnitude: f32,
+    direction: MDirection<f32>,
+}
```
Ralith: style nit: include an empty line between top-level blocks

common/src/worldgen/mod.rs
```diff
@@ -530,32 +549,173 @@ struct NeighborData {
...
+            .as_ref()
+            + (b.elevation_gradient
+                .reflect_from_side(side_b)
+                .direction
+                .as_ref()
+                - ab.elevation_gradient
+                    .reflect_from_side(side_a)
+                    .reflect_from_side(side_b)
+                    .direction
+                    .as_ref()))
```
Ralith: style: this is kind of hard to visually parse and has serious rightward drift. Lift out some intermediate variables?

common/src/worldgen/mod.rs
```diff
@@ -530,32 +549,173 @@ struct NeighborData {
...
+        Self {
+            magnitude: {
+                if m < ELEVATION_GRADIENT_MAGNITUDE_FLOOR {
+                    ELEVATION_GRADIENT_MAGNITUDE_FLOOR - (m - ELEVATION_GRADIENT_MAGNITUDE_FLOOR)
+                } else if m > ELEVATION_GRADIENT_MAGNITUDE_CEILING {
+                    ELEVATION_GRADIENT_MAGNITUDE_CEILING
+                        - (m - ELEVATION_GRADIENT_MAGNITUDE_CEILING)
+                } else {
+                    m
+                }
```
Ralith: Could this be simplified using `checked_sub` to combine the subtraction with the branching?

patowen:
> @patowen, please review for isotropy. Seems fine AFAICT but you have a better understanding of the problem than I.

I don't believe this is isotropic as it is currently written, mainly because the principles that make the current worldgen isotropic don't carry over to this PR. A lot of this depends on the `continue_from` logic.

In the current (outside of this PR) worldgen logic, the elevation increases or decreases when you cross any plane in the tiling. The `continue_from` logic helps ensure that whenever you cross a specific plane, the elevation change is always the same. However, this logic is highly sensitive, and it doesn't work as intended for `gradient_magnitude` because of the way it's capped. I believe `continue_from` correctly ensures that the change in `gradient_magnitude` is fixed for any given plane you cross, but the right setup can cause this logic to force `gradient_magnitude` to rise above `ELEVATION_GRADIENT_MAGNITUDE_CEILING` for child nodes with two parents, but not child nodes with one parent, making these two types of nodes distinguishable from each other.

I expect that `gradient_direction` has similar issues, as the root cause is related to the value not being a simple symmetrical addition or subtraction. If we just look at `perturb` alone, it seems fine, where going backwards or forwards is equivalent, but I don't see a good way to get it to play well with `continue_from`.

The effect of the gradient on the elevation is also a potential issue. For instance, it plays a role in `varied_from`, but it does not play a role in `continued_from`. While I haven't ruled out that the gradients are set correctly in `continued_from` for elevation and gradient to be consistent with each other, I find it unlikely.

If we believe that these issues are okay to ignore, I'll likely need to try to see if there's a way to systematically find a spot where the lack of isotropy becomes obvious. This may take a while for me to find and/or rule out as a possibility.

However, since I don't like just being a downer, I'll also try to include some ideas as well:

I think that a conceptually easier way to guarantee isotropy for any value you want to vary across hyperbolic space is to rely on unbounded scalar environmental factors and use them as input to a function (possibly a noise function in Euclidean space, or something else).

This doesn't immediately solve the "gradient" problem, though, since even if we assign a gradient to each node in an isotropic way, actually applying that gradient to the elevation is nontrivial, since the naive approach will cause the gradient to fail to apply when `continue_from` is called. However, as I write this, I do wonder if there's a good way to design another noise function with similar principles but whose domain is the dividing planes of the tiling instead of the nodes. This could achieve a similar purpose of the gradients by instead adding some consistency to the plane-by-plane elevation changes.

I also think a fair amount of terrain variety can be set up with some simpler approaches, such as having an elevation and elevation multiplier environmental factor, where a low elevation multiplier means that everything stays close to the ground plane.

DimpyRed:
Hi all, thank you for the feedback. Some flaws in the isotropy of the current method were highlighted.

I want to offer different of gradient propagation that should be a little easier to conceptualize and guarantee isotropy on.

varied_from steps:
1. The parent's elevation derivative in direction of traversal is calculated from the gradient and face normal.
2. A scalar-valued normally-distributed "gradient delta" is chosen.
3. The child node's elevation is set to parent elevation + (parent elevation derivative + (gradient delta) /2) * distance
4. The child node's gradient is set to parent gradient + gradient delta * displacement

continue_from steps:
1. Set the elevation of child node c to that of a + (b - c) as usual.
2. Calculate magnitude of the gradient of parent a in direction ab->a, and magnitude of the gradient of parent b in direction ab->b.
3. The gradient of c is constructed as follows:
 - The component in direction b->c is equal to the gradient at node a in direction ab->a.
 - The component in direction a->c is equal to the gradient at node b in direction ab->b.
 - The component perpendicular to both the two components above is inherited from any of the parents (all three parents and the child will have the same perpendicular component)

The idea is all nodes touching a plane have the same "towards plane" gradient component, and same elevation change with it's neighbor on the other side of the plane.

Issue #411, opened 2024-06-28
Pick block


VexianEmpire:
a button to select the block you are looking at?
button could be the middle click mouse button or the P key or something else

could be a potential solution to be able to access all of the blocks that naturally generate before the inventory system is finished and without having to give every single block its own key

(if this already exists i am very sorry i didnt find it anywhere)

Ralith:
In my experience, this pattern is extremely convenient even when constrained by inventory. We should implement this.

Issue #403, opened 2024-06-21
Add an input-management layer


patowen:
Currently, keyboard and mouse input is all handled in an ad-hoc manner in `window.rs`. Adding an extra layer of abstraction could help simplify the process of adding a new control. We could solve the following problems in one central place (although we don't need to solve them all now to resolve this issue):

- Allowing rebinding of controls
- Adding contextual controls
- Having input presses that trigger an event on the next tick
- Having an axis control for movement
- Detecting conflicts (and allowing the same key to do multiple things if unambiguous enough)
- etc. (This list is not exhaustive.)

Ideally, after this issue is done, if somebody wants to add a new key that does something, they'll just need to add the logic with minimal boilerplate code.

Ralith:
This is very much needed. I've toyed with the idea of building something reusable inspired by OpenXR/Steam Input, but it's a hard design problem.

PR #505, opened 2026-03-08
material cycle and picker keys

shadow-kestrel:
this pr adds keys to cycle the selected material through all materials, and select the material the player is looking at. implements #411

there's a case to be made for scrollwheel/middle-click bindings, but mine doesnt work reliably so i cant do so myself. #403 would make this a non-issue though

### DIFF
--- a/client/src/sim.rs
+++ b/client/src/sim.rs
@@ -190,10 +190,9 @@ impl Sim {
             Material::VALUES[(self.selected_material as usize + 1) % Material::COUNT];
     }
 
-    /// Cycles the selected material through all materials, in reverse.
     pub fn prev_material(&mut self) {
         self.selected_material = Material::VALUES
-            [(self.selected_material as usize - 1 + Material::COUNT) % Material::COUNT];
+            [(self.selected_material as usize + Material::COUNT - 1) % Material::COUNT];
     }
 
     /// selects the material of the block the player is looking at. Will never select void.


client/src/sim.rs
```diff
@@ -166,6 +166,16 @@ impl Sim {
         self.selected_material = *MATERIAL_PALETTE.get(idx).unwrap_or(&MATERIAL_PALETTE[0]);
     }
 
+    pub fn next_material(&mut self) {
+        self.selected_material =
+            Material::VALUES[(self.selected_material as usize + 1) % Material::COUNT];
```
Ralith: Use `wrapping_add` to avoid overflow.

Ralith (commented):
Thanks!

client/src/sim.rs
patowen: I don't think overflow is possible here in the first place. Even if `Material::COUNT` was `std::usize::MAX`, and `self.selected_material as usize` was `std::usize::MAX - 1` (its maximum possible given the material count), it wouldn't overflow.

Also, if the logic was changed so that it could overflow (such as using `+10` for large increments), having it panic in debug mode seems like the better option, since otherwise, it would result in a math error, not wrapping in the way we'd expect.

client/src/sim.rs
```diff
@@ -555,20 +598,7 @@ impl Sim {
...
-            self.cfg.character.block_reach,
-        );
-
-        let Ok(ray_casting_result) = ray_casing_result else {
-            tracing::warn!("Tried to run a raycast beyond generated terrain.");
-            return None;
-        };
-
-        let hit = ray_casting_result?;
+        let hit = self.looking_at()?;
```
patowen: Nice refactor!

Ralith (approved):

PR #515, opened 2026-08-17
Introduce skid_steer as a new way to load assets

patowen:
This is a relatively significant refactor of Hypermine, switching up asset loading to use a new library called "skid_steer".

It introduces the following changes:
* A custom `AssetLoader` type is added to drive asset loading, acting as the glue between skid_steer and Hypermine. This completely replaces the old `loader` module.
* All uses of the staging buffer and transfer module from the temporary `lahar_deprecated` module have been replaced with a vendored `growable_ring` implementation from @Ralith, and a `ParallelQueue` from the current version of lahar. This allows us to finally remove the `lahar_deprecated` module, a task that I had hoped to do years ago.
* PNG loading and GLTF loading, the only two processes that depended on the old `Loader`, have been migrated to the new `AssetLoader`.
* File-processing (reading PNGs and GLBs) and Vulkan-wrangling (making Vulkan calls) have been fully separated, with all data being read into memory in a convenient form before being passed into Vulkan data structures.
* To support some of the above changes, some data has been moved around, including the addition of a new `ShaderData` struct to pass around global data used for rendering.


client/src/graphics/gltf_mesh.rs
```diff
@@ -419,165 +269,13 @@ async fn load_material(
...
-                    &[vk::ImageMemoryBarrier::default()
-                        .dst_access_mask(vk::AccessFlags::TRANSFER_WRITE)
-                        .src_queue_family_index(vk::QUEUE_FAMILY_IGNORED)
-                        .dst_queue_family_index(vk::QUEUE_FAMILY_IGNORED)
-                        .old_layout(vk::ImageLayout::UNDEFINED)
-                        .new_layout(vk::ImageLayout::TRANSFER_DST_OPTIMAL)
-                        .image(handle)
-                        .subresource_range(range)],
-                );
-                xf.device.cmd_clear_color_image(
```
patowen, line 560: When separating out gtlf-reading logic with Vulkan wrangling, I had decided to remove the functionality that uses `cmd_clear_color_image` to fill the 1x1 image with a specified color, in favor of just specifying the color directly.

However, this requires a conversion from f32 color to SRGB, which you might have opinions on how to do. One option could be to import something like <https://crates.io/crates/palette> (although you might have a preferred crate for this). Given that we might want to do more in the future, I don't think we want to hardcode the gamma correction formula.

client/src/graphics/gltf_mesh.rs
```diff
@@ -249,153 +192,60 @@ async fn load_geom(
...
-            )
-            .await;
+            return Ok(MeshMaterialDefinition {
+                width: 1,
+                height: 1,
+                srgb_rgba_color_data: prim
+                    .material()
+                    .pbr_metallic_roughness()
+                    .base_color_factor()
+                    .map(|c| 255) // TODO: Will likely want a crate for color conversion
```
patowen: See https://github.com/Ralith/hypermine/pull/515/changes/BASE..0908719a3be8e59af41e30e075cbe97f0001cf17#r3834892686.

I deliberately kept the lint failing due to this line so that I wouldn't forget to address this.

EDIT: Actually, now that there's an unresolved PR comment, this shouldn't be forgotten anyway. I'll fix the lint.

client/src/graphics/asset_loader.rs
```diff
@@ -0,0 +1,403 @@
...
+                num_checked_events: 0,
+            }
+        }
+
+        fn push_event(&self, event: Event) {
+            self.events.lock().unwrap().push(event);
+        }
+
+        fn get_all_events(&mut self) -> Vec<Event> {
+            let events = self.events.lock().unwrap().clone();
```
Ralith, line 404: nit: `&mut self` means we have exclusive access, so there's no need to mess around with locking. Prefer `get_mut`.

client/src/graphics/asset_loader.rs
```diff
@@ -0,0 +1,403 @@
...
+            self.num_checked_events = events.len();
+            events
+        }
+
+        /// Drains all events that were returned by the most recent call to get_all_events
+        fn drain_queried_events(&mut self) {
+            self.events
+                .lock()
```
Ralith, line 425: As above.

client/src/graphics/asset_loader.rs
```diff
@@ -0,0 +1,403 @@
...
+            asset: asset_loader.load(DummyAssetSource {
+                name: name.to_owned(),
+                events: events.clone(),
+                progress_receiver,
+            }),
+            progress_sender,
+        }
+    }
+
+    fn test_eventual_success(mut f: impl FnMut() -> Result<(), anyhow::Error>) {
```
Ralith: This is complex and may still be fragile; there's no hard guarantee on a desktop OS that any particular thread will be scheduled in any particular time interval. Prefer to actually wait for the event we're interested in. For example, you could use a `std::sync::Condvar` to implement a helper that blocks until an event is delivered, and call that in a loop.

client/src/graphics/asset_loader.rs
```diff
@@ -6,20 +6,48 @@ use std::{
...
+    staging: Arc<GrowableRing>,
 }
 
 // Rather than exposing its fields, we expose helper functions for the kinds of tasks
 // one would need the context for. This helps add some level of separation between how global
 // data is organized and what asset loading code sees
 impl AssetLoadContext {
+    /// # Safety
+    /// - [`Work::cmd`] must not be used outside the lifetime of the returned [`Work`]
+    /// - Any Vulkan resources this work uses must not be destroyed before the [`Work`] is dropped
```
Ralith: Surely they must outlive the *execution* of the work?

client/src/graphics/asset_loader.rs
```diff
@@ -116,6 +229,93 @@ fn run_task_executor_thread(asset_load_context: AssetLoadContext, loader: skid_s
...
+        let _ = queue_watch_sender.send(semaphore_value);
+        unsafe { staging.tick(&gfx.device, semaphore_value) };
+    }
+    unsafe { queue.drain(&gfx.device) };
+    unsafe { queue.destroy(&gfx.device) };
+}
+
+/// Timeline sempahores must be signaled in a strictly increasing sequence, so having multiple threads manage
+/// the semaphore that unparks the timeline queue is error-prone. The purpose of this thread is to centralize
+/// management of this semaphore.
```
Ralith: Do we actually need to trigger this from multiple threads? If so, would it be simpler to use a `Mutex<u64>` that we hold across the signal operation?

client/src/graphics/asset_loader.rs
```diff
@@ -258,20 +472,29 @@ mod tests {
...
-        async fn load(mut self, _context: &Context<'_>) -> Option<Self::Output> {
+        async fn load(mut self, context: &Context<'_>) -> Option<Self::Output> {
             let mut status = DummyAssetStatus {
                 name: self.name.clone(),
                 progress: 0,
                 can_cancel: true,
                 events: self.events.clone(),
             };
 
+            // Use the parallel queue and set up an allocation to exercise this functionality
```
Ralith, line 523: That should probably be a separate test?

client/src/graphics/asset_loader.rs
```diff
@@ -351,6 +574,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(not(feature = "run_vulkan_tests"), ignore)]
```
Ralith: Let's just install lavapipe and maintain full test coverage in CI.

client/src/graphics/asset_loader.rs
```diff
@@ -116,6 +229,93 @@ fn run_task_executor_thread(asset_load_context: AssetLoadContext, loader: skid_s
                 thread::current().name().unwrap_or("<unnamed>")
             );
         });
+    let asset_load_context = Rc::get_mut(&mut asset_load_context)
+        .expect("runtime using this context should already be dropped");
+
+    // Fail-safe to ensure that the parallel queue is driven at least once after all work has been submitted before
+    // we drain the handle
```
Ralith: Why would that be needed? If we don't intend for it to be needed, then this will prevent us from detecting bugs with tests.

client/src/graphics/asset_loader.rs
```diff
@@ -116,6 +229,93 @@ fn run_task_executor_thread(asset_load_context: AssetLoadContext, loader: skid_s
...
+    gfx: &Base,
+    mut queue: ParallelQueue,
+    queue_unpark_semaphore: vk::Semaphore,
+    shutdown_token: CancellationToken,
+    queue_watch_sender: &tokio::sync::watch::Sender<u64>,
+    staging: &GrowableRing,
+) {
+    loop {
+        // Systems increment the `queue_unpark_semaphore` value when they want to guarantee
+        // that we don't park unless certain things are done. `ParallelQueueWaiter`
```
Ralith: What is `ParallelQueueWaiter`?

client/src/graphics/asset_loader.rs
```diff
@@ -28,8 +56,24 @@ impl AssetLoadContext {
         &self.gfx.memory_properties
     }
 
-    pub fn queue_family(&self) -> u32 {
-        self.gfx.queue_family
+    pub async fn wait_for_completion(&self, semaphore_value: u64) {
+        // To actually get the work to start, we need to unpark the queue. We do it here to avoid getting stuck awaiting something we never kicked off.
```
Ralith, line 88: Why is this needed? Submitting work should unpark the queue automatically.

client/src/graphics/asset_loader.rs
```diff
@@ -116,6 +229,93 @@ fn run_task_executor_thread(asset_load_context: AssetLoadContext, loader: skid_s
...
+    };
+}
+
+fn run_queue_driver_thread(
+    gfx: &Base,
+    mut queue: ParallelQueue,
+    queue_unpark_semaphore: vk::Semaphore,
+    shutdown_token: CancellationToken,
+    queue_watch_sender: &tokio::sync::watch::Sender<u64>,
+    staging: &GrowableRing,
```
Ralith: Consider aggregating this stuff into a struct with a `fn run(self)`.

Ralith (changes requested):
Review is just getting started, but it may be a few days before I make more progress.

client/src/graphics/gltf_mesh.rs
Ralith: I haven't gotten to the context of this question yet, but the `color` crate is an alternative to consider for color space conversions; it's maintained by some extremely reputable 2D graphics folks. I don't think we *need* an external crate for gamma conversion but it's fine to use one.

client/src/graphics/asset_loader.rs
patowen: I messed up here, removing this cfg in the wrong commit. I think what happened is that I knew this was the second commit, but I accidentally edited the second-to-last commit. Sorry about that.

client/src/graphics/asset_loader.rs
patowen: It is a struct that no longer exists. I must have missed this when proofreading.

client/src/graphics/asset_loader.rs
patowen: The `events` field is an `Arc<Mutex<Vec<Event>>>`, so although we have exclusive access to the `EventList`, we don't have exclusive access to the events it holds.

client/src/graphics/asset_loader.rs
patowen: I don't see how that would make it more robust. Either the condition variable is triggered, and the test passes, or the condition variable is not, and the test hangs. In the latter, I would prefer the test to fail with a timeout error. The three-second timeout I added should be orders of magnitude longer than it would actually take.

I feel like if the asset loader hangs for 3 seconds for no discernible reason, that would be a bug in the asset loader, not a bug in the test.

While the function feels complex, at it's core, it's quite simple: Keep trying until it succeeds, or until it's time to give up.

If we want to be more robust than that, we'd need to dig into the implementation details of skid_steer and detect when it's no longer going to make progress.

client/src/graphics/asset_loader.rs
patowen: Good catch. We do want to prevent someone from naively having a `load` function submit work but fail to wait on it.

client/src/graphics/asset_loader.rs
patowen: I don't see the benefit of having a test that doesn't use the parallel queue. Is there a benefit to having multiple different implementations of `skid_steer::Source` in the unit tests?

client/src/graphics/asset_loader.rs
patowen: So, thinking about it, I'm not sure whether we need this fail-safe.

On one hand, we need it because if a `skid_steer::Source` submits work without waiting on it, trying to drain the handle without driving the queue would result in a deadlock. After all, in the standard flow, waiting on the queue is what generally triggers the queue to be unparked (due to the implementation of `wait_for_completion`).

On the other hand, no asset sources should do that, since it would be unsound in almost all situations. I generally prefer crashes or validation errors over deadlocks though, since they tend to be easier to pin down.

Overall, I'm still in favor of keeping it. I think that it's more likely to ease debugging than make it more difficult (due to passing tests).

client/src/graphics/asset_loader.rs
patowen: Submitting work does not unpark the queue automatically. The only thing calling `Work::end` does is send it to the mpsc. The receiver of this mpsc only activates when `ParallelQueue::drive` is called, and that is what triggers the `queue_submit` function.

I should probably double-check the comments to make sure I say "sent for submission" instead of "submitted" in the appropriate places, since they are subtly different.

client/src/graphics/asset_loader.rs
patowen: Since sending work for submission does not on its own unpark the parallel queue, we need to trigger this every time we send work for submission. This probably would be simpler with a mutex. It can be tricky at times to weigh options between different synchronization primitives.

I do like this version because it only signals the semaphore even if everyone tries to signal it at the same time, although it's unclear whether that's worth running an extra thread.

client/src/graphics/asset_loader.rs
Ralith: > In the latter, I would prefer the test to fail with a timeout

Condvar with `wait_timeout`, then?

client/src/graphics/asset_loader.rs
Ralith: Signaling should not be an expensive operation.

client/src/graphics/asset_loader.rs
Ralith: I assumed that's why you had all the different event types. Seems like a lot of logic independent of the parallel queue if the use of the parallel queue is all you want to test.

client/src/graphics/asset_loader.rs
patowen: That might work. I guess the idea is to specialize this function to wait for events to show up instead of waiting an arbitrary amount of time with exponential backoff.

client/src/graphics/asset_loader.rs
Ralith: How does this behavior help trigger a crash or validation error?

client/src/graphics/asset_loader.rs
Ralith: Whoops, right you are. We could close that gap, but I guess it's nice to be able to batch stuff up.

client/src/graphics/asset_loader.rs
Ralith: Would it be more useful to detect that class of bug by panicking if a load task finishes without waiting for work it submitted?

patowen: I'm not convinced it is useful enough to be worth the extra scaffolding. The solution currently in place seems simplest, as it's essentially a way to ensure that the call to `Handle::drain` (which comes right after it) is preceded by a guarantee that the queue will be driven.

To answer your question on how it would help trigger a crash or validation error, it would ensure that the work sent for submission is actually submitted. If said work depends on resources that were already destroyed, that should trigger the validation error.

An alternative possibility is to let the deadlock happen, but in that case, I'd likely want to wrap it in a timer that at least logs a helpful warning if it runs out.

client/src/graphics/asset_loader.rs
patowen: That's fair. It might be easier to read the tests if we just have an asset type that exercises the parallel queue without doing anything else.

patowen: I'm currently second-guessing this. When trying to maximize coverage of unit tests, anything that reduces the number of different kinds of tests you need to run seems like a win for simplicity. With this in mind, the question I would ask is whether there is logic we want to test that specifically depends on the asset _not_ interacting with the staging buffer and parallel queue. I would argue that the answer is no, in which case, we should only have one `DummyAssetSource` type, which would interact with Vulkan as well as having it's load progress able to be manipulated from the outside.

I feel like any attempt to separate this into two separate asset sources will double the code required to get the same level of test coverage without any tangible benefit.

The main tangible benefit I can think of to splitting this up would be that if there is a test failure, it would be easy to find the root cause. However, I already think it's easy enough to find the root cause, just based on my experience debugging the test failures that did happen.

client/src/graphics/asset_loader.rs
patowen: My next force-push will include an attempt to resolve this.

EDIT: Done

client/src/graphics/asset_loader.rs
patowen: My next force-push will include an attempt to resolve this.

EDIT: Done

client/src/graphics/asset_loader.rs
patowen: Thinking about this further, I feel like the current implementation is the one I prefer, as it follows the principle of least surprise, as we shouldn't need to expect draining the handle to deadlock. I consider unparking the semaphore to be a hard prerequisite to calling `Handle::drain`. After all, if the gap mentioned in https://github.com/Ralith/hypermine/pull/515#discussion_r3868034186 were closed, this would have happened automatically, and we wouldn't even be having this discussion (even though the soundness risk would still be present).

I don't think there's a clean, reliable, non-over-engineered way to check whether the load task waited for the work it submitted, since `load` functions are not currently required to report the necessary information to trigger this panic.

client/src/graphics/asset_loader.rs
patowen: My next force-push will include an attempt to resolve this.

EDIT: Done

client/src/graphics/gltf_mesh.rs
patowen: I might consider the `color` crate. Although implementing it myself (or copying `color`'s implementation would be pretty simple, it's complicated by the fact that the gamma conversion function is piecewise, starting linear and ending with a power function, and I also have to convert from a float to a `u8`, which has its own rounding standards for color (that I don't remember off the top of my head). My understanding is that when something is messy due to existing standards, it's a good sign that it would be best to reuse an existing implementation.

client/src/graphics/asset_loader.rs
```diff
@@ -28,8 +57,24 @@ impl AssetLoadContext {
...
+            .wait_for(|&value| value >= semaphore_value)
+            .await
+            .is_err()
+        {
+            // Don't assume things have finished loading just because `queue_watch_sender` was dropped.
+            // Instead, print an error and wait forever, since the work will never complete.
+            tracing::error!(
+                "Work was submitted to the parallel queue but never completed. This should never happen."
+            );
+            std::future::pending::<()>().await;
```
Ralith: Would an abort be more convenient here?

client/src/graphics/mod.rs
```diff
@@ -1,5 +1,6 @@
 #![allow(clippy::missing_safety_doc)] // Vulkan wrangling is categorically unsafe
 
+pub mod asset_loader; // TODO: `pub` is here to suppress usage warnings. Remove in a later commit.
```
Ralith: nit: `#[expect(whatever)]` is the idiomatic pattern to suppress such lints..

client/src/graphics/asset_loader.rs
```diff
@@ -116,6 +249,85 @@ fn run_task_executor_thread(asset_load_context: AssetLoadContext, loader: skid_s
...
+    gfx: Arc<Base>,
+    queue: ParallelQueue,
+    queue_unpark_semaphore: vk::Semaphore,
+    queue_shutdown_token: CancellationToken,
+    queue_watch_sender: tokio::sync::watch::Sender<u64>,
+    staging: Arc<GrowableRing>,
+}
+
+impl QueueDriver {
+    pub fn new(
```
Ralith: style nit: is this function carrying its weight?

client/src/graphics/gltf_mesh.rs
```diff
@@ -249,153 +192,60 @@ async fn load_geom(
...
-            )
-            .await;
+            return Ok(MeshMaterialDefinition {
+                width: 1,
+                height: 1,
+                srgb_rgba_color_data: prim
+                    .material()
+                    .pbr_metallic_roughness()
+                    .base_color_factor()
+                    .map(|_c| 255) // TODO: Will likely want a crate for color conversion
```
Ralith: Technically we're supposed to multiply any base color texture by this factor, so maybe the definition should take both pieces of data; then the previous GPU-driven initialization strategy can be retained easily.

In practice I don't really care about properly loading GLTF files that actually use this for anything other than solid colors without a texture. GLTF is one of those formats where you have to pick your battles.

Ralith (approved):

client/src/graphics/asset_loader.rs
Ralith: I don't care that much, and this already seems like an awful lot of code for a test, so keeping it unified is fine with me.

client/src/graphics/mod.rs
patowen: The main issue I had was that I still wanted to get warnings about non-pub things within the module that were unused. the `[expect(unused)]` blast radius was too large for my preferences.

That being said, it's not a big deal either way, so I'm happy to be more idiomatic here.

client/src/graphics/asset_loader.rs
patowen: Thanks! I would like to agree with you wherever I can to ease the reviewing process, especially when the decision is opinion-based, but until I fully understand the reasoning behind a particular request in a PR, it can be very difficult for me to fulfill it without lowering my standards to allow for more guesswork.

I needed to push back here because I was trouble picturing how to set up well-designed unit tests under the constraint that the tests exercising the parallel queue should be kept separate from the existing tests, so I felt like keeping the code as-is was the safest option for now.

client/src/graphics/gltf_mesh.rs
patowen: True. I think a suitable compromise there would be to log a warning if there's a texture, and the `base_color_factor` is not 1, stating that we don't currently support configuration like that.

EDIT: Updated

client/src/graphics/mod.rs
Ralith: It's just an intermediate state, so w/e.

client/src/graphics/asset_loader.rs
patowen: I would probably prefer a `panic`, since that would allow a panic handler to run. If we're worried about soundness due to Vulkan, it seems like it would make sense to go ahead and turn on panic abort for this crate.

Ralith: Vulkan should be happy so long as we don't destroy the device or instance before exiting. We could achieve that by catching panics before the relevant `Drop` impl runs and aborting there.

patowen: I could run something like
```rust
// Don't assume things have finished loading just because `queue_watch_sender` was dropped.
// Instead, we should panic, since the work will never complete.
let Err(_e) = panic::catch_unwind(|| {
    panic!(
        "Work was submitted to the parallel queue but never completed. This should never happen."
    )
});
// Aborting to prevent potential issues with code that is not exception-safe. We avoid aborting immediately
// so that panic handlers can run.
std::process::abort();
```

However, I'm thinking that the right move is to just panic. I believe that there are already several places where panicking is unsound, and I'm not 100% sure that it's unsound to panic here, as `AssetLoader`'s `Drop` implementation does wait for the queue to drain. I think that if we're worried about Vulkan unsoundness, the right move is to just apply panic-abort throughout the entire `client` trait.

With this in mind, for this PR, I think the right move is to just call `panic!` without overthinking things.

Ralith: > I think that if we're worried about Vulkan unsoundness, the right move is to just apply panic-abort throughout the entire client trait.

We'd want to introduce a `catch_unwind` specifically before a `Drop` impl that cleans up the Vulkan device/instance runs. That's not here.

Because it's a pre-existing issue, no need to resolve it in this PR. A plain panic here is fine.

patowen: Sounds good. I went ahead and used `.expect` to turn this into a plain panic.

client/src/graphics/gltf_mesh.rs
patowen: I have now introduced the color crate. Since https://github.com/Ralith/hypermine/pull/515#discussion_r3887598837 is open, I'll go ahead and resolve this thread.

client/src/graphics/gltf_mesh.rs
patowen: Resolving in favor of https://github.com/Ralith/hypermine/pull/515#discussion_r3887598837

patowen:
I'll request one more quick review to double-check my latest changes, although based on your previous approval, I'm pretty confident this is safe to merge as-is.

Ralith (approved):

Ralith (approved):

PR #519, opened 2026-08-22
Use install-vulkan-sdk-action GitHub action, including lavapipe software rasterizer

patowen:
This PR also un-ignores the tests that were ignored due to the inability to run them on GitHub actions.

https://github.com/marketplace/actions/install-vulkan-sdk-and-runtime

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,20 +18,12 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-      - name: Set Windows environment variables
-        if: matrix.os == 'windows-latest'
-        shell: bash
-        run: |
-          echo "VULKAN_VERSION=1.3.290.0" >> "$GITHUB_ENV"
-          echo "VULKAN_SDK=C:/VulkanSDK/1.3.290.0" >> "$GITHUB_ENV"
-      - name: Install dependencies
-        if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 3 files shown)


.github/workflows/rust.yml
```diff
@@ -18,20 +18,12 @@ jobs:
...
     steps:
-      - name: Set Windows environment variables
-        if: matrix.os == 'windows-latest'
-        shell: bash
-        run: |
-          echo "VULKAN_VERSION=1.3.290.0" >> "$GITHUB_ENV"
-          echo "VULKAN_SDK=C:/VulkanSDK/1.3.290.0" >> "$GITHUB_ENV"
-      - name: Install dependencies
-        if: matrix.os == 'ubuntu-latest'
-        run: sudo apt update && sudo apt-get -y install libasound2-dev libvulkan-dev libfontconfig-dev libshaderc-dev
```
Ralith, line 29: We still need the non-Vulkan dependencies here.

Ralith (changes requested):

.github/workflows/rust.yml
Ralith: Or apparently not, I guess? Must've been copy-pasted...

Ralith (approved):

.github/workflows/rust.yml
patowen: Oh yeah, I took the success of CI to mean that we didn't need them, although I wonder if we want these extra dependencies anyway.

Ralith: Eh, we'll add 'em when we need 'em.

